Issue #7, opened 2025-03-27
[TASK] Set up ruff linter locally


jakewheeler:
## 🚧 What needs to be done

We want to install and configure [ruff](https://docs.astral.sh/ruff/) to be run against the Refiner's codebase locally. Meaning, developers should be able to run ruff to lint and format code while we work.

Additionally we should configure ruff to run via `pre-commit` ([similar to eCR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/.pre-commit-config.yaml#L2)) so that we can expect the code to be regularly, and automatically, linted & formatted.

## 🤔 Why it needs to be done

Setting up and making use of ruff will enable the dev team to be able to focus on more important tasks, like feature development, and not need to put much thought into things like code style and formatting.

## 🏗️ Acceptance Criteria

- [ ] ruff has been locally installed as a dev dependency and can be run on-demand
- [ ] ruff has been set up with a basic configuration
- [ ] ruff will automatically run when a dev makes a commit
- [ ] docs have been updated

## ⚙️ Technical Notes

We can likely borrow much of this from the way it works in the [eCR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer).

PR #9, opened 2025-03-27
chore: Set up ruff linter locally

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds the `ruff` linter to the project and provides some basic configuration.

## 🔗 Related Issue

Fixes #
- #7 

## ✅ Acceptance Criteria

- [x] ruff has been locally installed as a dev dependency and can be run on-demand
- [x] ruff has been set up with a basic configuration
- [x] ruff will automatically run when a dev makes a commit* (as long as `pre-commit` was installed and the `install` command was run in the project)
- [x] docs have been updated

## 🧪 How to test

Please refer to the [`Running the linter` section](https://github.com/CDCgov/dibbs-ecr-refiner/pull/9/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5) in the updated README for setup & testing instructions. Feel free to let me know if these can be improved!

## ℹ️ Additional Information

Maybe not the biggest issue since I imagine we'll often be using `ruff` via the `pre-commit` git hook, it would be a lot nicer if we didn't have to install `ruff` in both the `refiner` and the `trigger-code-reference`. However, these are currently two independent projects 😕

Is there a better way to go about this with how things are currently set up?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.2
+    hooks:
+      # Run the linter.
+      - id: ruff
+        args: [--fix]
+      # Run the formatter.
+      - id: ruff-format
--- a/README.md
+++ b/README.md
@@ -2,6 +2,24 @@
 
 🚧 The DIBBs eCR Refiner is under construction 🚧
 
+## Running the linter
+
+The linter can be run two different ways: either manually via the `ruff` command or automatically when you go to create a new commit, which is powered by `pre-commit`.
+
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
```
jakewheeler, line 1: This config seems to work correctly for both `refiner` and `trigger-code-reference`.

robertmitchellv (approved):
this looks great and works for me locally. i think `ruff` is great and would ideally love to maybe add a ticket to move away from `pip` to `uv` at some point as well

PR #11, opened 2025-03-31
chore: docker compose hot reloads on file changes

### LINKED ISSUE #10: [TASK] Configure `docker-compose.yaml` to hot-reload the application on file changes
## 🚧 What needs to be done

Currently, the `docker compose` environment will not automatically update when changes are made to files. We should modify the `docker-compose.yaml` (and other files, if required) so that we can make local development using Docker much simpler.

## 🤔 Why it needs to be done

Developers will be able to immediately see the effects of their code changes during local development without needing to restart their containers and/or delete images.

## 🏗️ Acceptance Criteria

- [ ] When an application file is modified locally, the container will update its files and the FastAPI server will reload
- [ ] Restarting containers and/or rebuilding images is not needed to see code changes reflected

## ⚙️ Technical Notes

There are likely various different ways to make this happen. It looks like [uvicorn](https://www.uvicorn.org/settings/#development) itself offers a reload feature and could be one path to take.

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR:

1. Makes it so that editing files in `/api/refiner` and `/api/trigger-code-reference` will cause the respective service to reload
2. Makes it so that new files can be added or removed locally and the changes will be reflected in the running container(s)

## 🔗 Related Issue

Fixes #10 

## ✅ Acceptance Criteria

- [x] When an application file is modified locally, the container will update its files and the FastAPI server will reload
- [x] Restarting containers and/or rebuilding images is not needed to see code changes reflected

## 🧪 How to test

A quick way to understand how this works:

1. From the project root, run `docker compose up`
4. In `api/refiner/app/main.py` try changing the `health_check()`'s return to `{"status": "TEST"}`
5. Save the file and you should see logs that let you know the server has reloaded
6. Hit the API's health check endpoint (`/`) and you should see your modified response

## ℹ️ Additional Information

What has been done so far seems to do everything I'd want it to with the way the repo is currently structured. Is there anything else that'd be helpful to change right now?

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,8 +1,12 @@
 services:
   message-refiner-service:
     platform: linux/amd64
+    command: uvicorn app.main:app --host 0.0.0.0 --port 8080 --reload
     build:
       context: ./api/refiner
+    working_dir: /app
+    volumes:
+      - ./api/refiner:/app
     depends_on:
       - trigger-code-reference-service
     ports:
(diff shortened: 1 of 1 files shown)


robertmitchellv (approved):
really great QOL improvement to the dev experience. thanks for this!

tested it locally and works great

image

PR #13, opened 2025-04-01
chore: fix CODEOWNERS file syntax

jakewheeler:
This should solve the issue with folks not getting auto-assigned for PR reviews.


robertmitchellv (approved):
Thanks for figuring this out!

PR #15, opened 2025-04-01
fix: add the icd-9 url back to the list of expected values

### LINKED ISSUE #14: [BUG] Fix error when SNOMED codes are used in `conditions_to_include`
## 📜 Description

When SNOMED codes are used as a query param in `conditions_to_include` there is a `500` error. This needs to be figured out so that we can demo current state.

## 👣 Steps to reproduce

1. build clean images using `docker compose build --no-cache`
2. start the services with `docker compose up -d`
3. using the API or rest tool of your choice and a test eICR for the body, `POST` this url:

`http://localhost:8080/ecr?conditions_to_include=240589008`

## 🌃 Environment:

- OS: ubuntu from the python slim 3.13 image

## ??? Expected Result

We should receive a `200` and have the dictionary of codes returned from the TCR and used to search through the eICR sections for any matches.

## 🧪 Actual result

a `500` is returned and there are errors in the container logs.

#### Priority

- [x] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

The icd-9 url at some point was removed from the list of expected systems, which was causing an error since those codes were still in the sqlite db. For now, this will keep things working as expected as we think about what the core needs are of a TCR explicitly as it relates to refiner and begin the work of folding the two services into one.

## 🔗 Related Issue

Fixes #14 
- #14 

## 🧪 How to test

1. build clean images using `docker compose build --no-cache` at the project root
2. start the services with `docker compose up -d`
3. using the API or rest tool of your choice and a test eICR for the body, `POST` this url: `http://localhost:8080/ecr?conditions_to_include=240589008`
4. if there are any observations connected to this it should be preserved.

### DIFF
--- a/api/refiner/app/utils.py
+++ b/api/refiner/app/utils.py
@@ -43,10 +43,11 @@ def create_clinical_services_dict(
     """
     Transform the original Trigger Code Reference API response to have keys as systems
     and values as lists of codes, while ensuring the systems are recognized and using their
-    shorthand names so that we can both dynamicall construct XPaths and post-filter matches
+    shorthand names so that we can both dynamically construct XPaths and post-filter matches
     to system name varients.
     """
     system_dict = {
+        "http://hl7.org/fhir/sid/icd-9-cm": "icd9",
         "http://hl7.org/fhir/sid/icd-10-cm": "icd10",
         "http://snomed.info/sct": "snomed",
         "http://loinc.org": "loinc",


jakewheeler (approved):
Tested with the sample doc you sent me yesterday and this is working for me. Thanks for fixing this!

Issue #4, opened 2025-03-26
[TASK] Create GitHub Action to run tests on PRs


jakewheeler:
## 🚧 What needs to be done

When working on a PR, we want to be able to run the project's tests every time we push new commits.

## 🤔 Why it needs to be done

Running the tests automatically while working on changes will enable us to better ensure that the changes we're making are not causing regressions.

## 🏗️ Acceptance Criteria

- [ ] `.github/workflows` directory with an Action that runs the project's tests have been added to the project
- [ ] Green checkmark or red X displays next to the PR title, depending on check status
- [ ] PR branch re-runs tests every time new commits are pushed to it

## 📜 Additional context

- Can we adapt a lot of this from the [eCR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/.github/workflows/container-message-refiner.yaml)?

PR #19, opened 2025-04-02
refactor: merge TES code into Refiner application

### LINKED ISSUE #18: [TASK] Package TES related code into the Refiner
## 🚧 What needs to be done

Currently, this repo contains a copy of the `trigger-code-reference` service which spins up alongside the refiner when the application runs. The refiner then sends http requests to the TCR in order to get the data it needs from the local TES SQLite database.

In this ticket, we should roll the `trigger-code-reference` into the refiner applications code base so that it all runs as a single application. This will greatly reduce complexity of the application while also allowing for much more straightforward packaging and deployment of the app.

NOTE: The reason behind this decision is due to the TCR deviating to more closely meet the needs of the eCR Viewer. The refiner will likely have its own specific needs when it comes to the TES data that is seeded and used. We want to ensure we're able to customize this data and code to meet the project's needs.

## 🤔 Why it needs to be done

- Simplifies the application code base
- Eliminates HTTP calls between services
- Single image building and deployment

## 🏗️ Acceptance Criteria

- [ ] Application works as it did before
- [ ] TCR related code lives in the `refiner` code base
- [ ] TCR code has been cleaned up to remove anything unnecessary

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

- `trigger-code-reference` service was condensed into a single `refiner/app/db.py` file and includes only what the refiner was previously making use of
  - Docker compose file was updated to only start the refiner container
- Made a couple of small updates to `main` to now call the `get_value_sets_for_condition` function rather than make an http request to the TCR to get clinical info
- Updated `dev-requirements.txt` to include dependencies required for running scripts to seed the TES database (`refiner/scripts/`)
- Updated documentation for the seed script
-  Fixed refiner integration test
- Updated `pytest.ini` to include an `integration` marker (eliminates a warning we were seeing previously)
- Created a new `test_db.py` file
- Added GitHub Action to run all tests in the pipeline

## 🔗 Related Issue

Fixes #18, #4

## ✅ Acceptance Criteria

- [ ] Application works as it did before
- [ ] TCR related code lives in the refiner code base
- [ ] TCR code has been cleaned up to remove anything unnecessary

## 🧪 How to test

The main things to check are that the refiner works as it did previously and that the tests are all passing now.

To test the app manually:

1. Ensure your docker compose environment was shut down and start it up using this branch, `docker compose up`
2. When started, you should see only the `message-refiner-service` start up
3. Try sending some test requests and checking the results

I used a diff checker to test the outputs between my changes here and what's in `main` and was able to get the exact same result between branches. Please double check me!

If you'd like to also check that the tests are working:

```sh
cd refiner
pytest tests/
```

You should see all tests passing.

## Additional Info

There is some code clean up work I think can happen but I didn't feel it was necessary to make those changes now. I mostly wanted to focus on getting things working as they were previously without adding too many new things.


.gitignore
```diff
@@ -21,5 +21,8 @@ node_modules/
 .coverage
 coverage.xml
 
-#Environment
-.env
+# Environment
+**/.env
+
+# Tools
+.requests
```
robertmitchellv, line 28: i've also been doing this with `.requests` 👍🏻

robertmitchellv (approved):
thanks for this Jake!

- tests are passing locally ✅ 
- rebuilt the container and tested with the `test_eICR.xml` and the `.http` i've been using for demos and everything is working as expected 🎉

PR #20, opened 2025-04-03
docs: add information for ways to use the TES api and demoable refiner requests

### LINKED ISSUE #17: [SPIKE] explore and document the TES API
## 🚧 What needs to be done

Given the TES' FHIR server endpoint and large CapabilityStatement there is a lot we can do with the TES that we're currently not using in the TCR. Additionally, since most of what we have been doing with the TCR is based on the eRSD, it is important to understand in what ways we can begin to tailor the TCR specifically for the refiner rather than a tool that is shared with the viewer. Since there isn't any documentation we'll need a thorough run down of what we can do so we can begin planning a unified refiner/TCR service.

## 🤔 Why it needs to be done

The TCR currently has a lot of viewer specific functionality and is solely based on the eRSD. The refiner will likely need to have a bit more awareness of condition related groupers that can be short cuts to grabbing clinical codes related to conditions, which can help with refining. Since there isn't documentation we can use as a shortcut we need to explore the endpoint.

## ⏰ Timebox

This should be limited to 2 days.

## 📝 To-do list

  - [ ] ...
- [ ] Research + Analysis
  - [ ] ...
  - [ ] ...
  - [ ] ...
- [ ] Review and Finalize
  - [ ] ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Adding a `docs/` directory with two subdirectories:

1. `refiner`, which will contain an `.http` file and eICR input for testing the refiner and demoing what it is capable today
2. `tes`, which will have a README with information on how to use the TES and some different ways to get codes by coded value and, description, and human readable name as well as the FHIR API CapabilityStatement and `.http` file with example requests
 
## 🔗 Related Issue

Fixes #17
- #17 

## 🧪 How to test

For `refiner`, you can run:

```bash
docker compose up -d
```

to start the services and then try to run the examples from the `.http` file.

For `tes` you can obtain an API key, use the python examples in the README or the `.http` file following the development advice for adding the API key in the header and testing out the examples.

There is also a python script and `requirements.txt` if you want to use an `.env` file and just run all of the examples in one go to see that they all work as expected.

## ℹ️ Additional Information

There are potentially more ways to leverage this API that will be added in the future.


docs/tes/README.md
```diff
@@ -0,0 +1,303 @@
...
+  },
+  "dev": {
+    "API_URL": "https://tes.tools.aimsplatform.org/api/fhir",
+    "API_KEY": "your-api-key"
+  }
+}
+```
+
+> [!NOTE]
+> the `"$schema"` value is for using the Kulala NeoVim package. There are likely other `.env` examples if you are using a different tool for `.http` files.
```
jakewheeler, line 280: Just wanted to note I'm using [Rest Client](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) for VS Code and that allows for [environments](https://marketplace.visualstudio.com/items?itemName=humao.rest-client#environment-variables) as well.

The `explore-tes.http` file works perfectly after updating my VS Code `settings.json` file to include the following:
```json
  "rest-client.environmentVariables": {
    "$shared": {
      "API_URL": "https://tes.tools.aimsplatform.org/api/fhir",
      "API_KEY": "key-goes-here"
    }
  }
```

docs/tes/README.md
```diff
@@ -0,0 +1,303 @@
...
+
+To use the provided python examples with the TES API, create a `.env` file with the following structure:
+
+```
+TES_API_URL="https://tes.tools.aimsplatform.org/api/fhir"
+TES_API_KEY="your-api-key"
+```
+
+> [!NOTE]
+> the `requirements.txt` is associated with the [`explore-tes.py`](explore-tes.py) file, so please make sure you run `pip install -r requirements.txt` prior to running the script.
```
jakewheeler, line 292: I created a new `venv` & activated it in this directory, installed the deps, ran the script and it worked as expected 👍

docs/tes/README.md
robertmitchellv: i think i'll add this info then because it will be more likely that others will have the same package on vscode than neovim--thanks for testing that!

docs/tes/README.md
```diff
@@ -0,0 +1,303 @@
+# Terminology Exchange Service (TES) API Integration Guide
```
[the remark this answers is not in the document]
robertmitchellv: that's a good point; i'll add a note about that. i think there's a lot we can do with searching specific conditions using the text name or the snomed concept to help in creating the filters. the groupers are harder to figure out since they have names like `rs-grouper-integer` but they could also be useful as well for similar reasons.

there may be a pattern to of using the code in the RR to get a short hand name (similar to what eCR Viewer does with stamping) and then using that name to get all clinical codes for the disease rather than just searching the code itself and where else it shows up. i'll try to add this today.

jakewheeler (approved):
Great work on this!

PR #28, opened 2025-04-08
docs: add documentation on interplay between eICR and RR and how this impacts direction of the refiner

### LINKED ISSUE #25: [SPIKE] How will the RR's data be leveraged for the refiner?
## 🚧 What needs to be done

Given the way that the refiner _has_ been working and knowing that a critical piece of this work moving forward is predicated on exactly what way the RR's data will be leveraged by the refiner it is important to think through these connections. There may be some changes that are required to the refiner to accommodate the path forward and we need clearly sketched out points of connection between the two files to best plan for this undertaking.

## 🤔 Why it needs to be done

We are currently lacking the most critical piece of information when it comes to refining since we're only working with the eICR. This needs to get done in order for our product to, in the bare minimum sense, be able to meet users' needs.

## ⏰ Timebox

1-2 days

## 📝 To-do list

  - [ ] ...
- [ ] Research + Analysis
  - [ ] ...
  - [ ] ...
  - [ ] ...
- [ ] Review and Finalize
  - [ ] ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

There are two files being added to the repo:

- `eCR-CDA-Notes.md`
- `refiner-config.json`

### `eCR-CDA-Notes.md`

I combined the older `eICR-Notes.md` doc with updates that fold in details about the RR that will be important for future state refiner functionality. This is focused on eICR 1.1 and RR 1.1 for now; support for 3.1 is not folded into the document yet since there are many new sections and trigger code containers--it can get confusing and this is meant to help onboard technical folks working on refiner to understand what is most important for our work.

### `refiner-config.json`

This is a reworking of the `refiner_details.json`. In order to get some of the core blocks of info we need for business logic in refiner, I spend a lot of time thinking of how to model things that are important in the eICR and the RR. This also maps out eICR 1.1 and 3.1 as well as RR 1.1.

Here is an example of data related to a header:

```json
{
  "2.16.840.1.113883.10.20.15.2:2016-12-01": {
    "type": "header",
    "display_name": "Public Health Case Report, Release 2, STU Release 1.1",
    "spec": "eICR",
    "version": ["1.1"],
    "required": true,
    "code_type": "atomic",
    "code": "55751-2",
    "contains": {
      "1.1": [
        "2.16.840.1.113883.10.20.22.2.22.1:2015-08-01",
        "1.3.6.1.4.1.19376.1.5.3.1.3.4",
        "2.16.840.1.113883.10.20.22.2.2.1:2015-08-01",
        "2.16.840.1.113883.10.20.22.2.38:2014-06-09",
        "2.16.840.1.113883.10.20.22.2.10:2014-06-09",
        "2.16.840.1.113883.10.20.22.2.5.1:2015-08-01",
        "2.16.840.1.113883.10.20.22.2.12",
        "2.16.840.1.113883.10.20.22.2.3.1:2015-08-01",
        "2.16.840.1.113883.10.20.22.2.17:2015-08-01"
      ]
    }
  }
}
```

The keys in this json doc are a combination of the `templateId`'s `root` with its `extension` (if available). This ends up looking like the Object ID (OID) and a date separated by a colon. This is consistent with how the specs present unique IDs for specific blocks of information.

`"type"` indicates what part of the document this information is for; rather than just call this related to `ClinicalDocument`, i'm calling this `"header"` (we can change this in the future if necessary. Other examples will use the XML tag where the information is contained, e.g., `<section>` `==` "section", `<observation>` `==` "observation", etc.

`display_name` is the string name for what this information is about. This might not be leveraged in the refiner itself but it helps us understand what each of these pieces of information represent because the OID:date key is not intuitive.

`"spec"` refers to whether this is for an eICR or an RR.

`"version"` is a list since some templates for the eICR can be for both the 1.1 or 3.1.

`"required"` is not used in the same way that it is in the spec; this is more related to sections so that we know if a section **MUST** be included in the output or not. For more downstream things like observations they are not required at a document level. There are validation rules like, "some trigger code data **must** be in an eICR" and this field is **not** related to that level of validation.

`"code_type"` is how we can understand in what way we might need to unpack the `"code"` value. These are the types that I've thought are most important:

- `"atomic"` means that it's just a single string value, this is typically going to be a single LOINC code.
- `"list"` means that the values are a list of possible values
- `"map"` means that there are key/value pairs
- `"flexible"` means that the `<code>` can come from a ValueSet that is pretty long, like the RCTC list for trigger code related information

`"code"` is the atomic, list, map that was described above. If `"flexible"` then it is `null`

`"contains"` is a list of OID:date values the represent relationships. This is how we can connect different pieces of information in a more shallow way rather than relying on a lot of nesting.

This was done by hand using the vol 2 of the specs for eICR 1.1 and 3.1 as well as vol 2 of the spec for RR 1.1. 

The next set of tasks will involve coming up with a strategy for leverage the information about the RR specified in `refiner-config.json` to return a Python list of only "Reportable" SNOMED codes.

## 🔗 Related Issue

Fixes #25 
- #25


refiner/assets/refiner-config.json
robertmitchellv, line 1: this will exist side-by-side the `refiner_details.json` file since that is still be used by the app currently. eventually i'd like to move away from this kind of important application related data being in a single file but that will be worked on in later tickets. i wanted to have it in a similar place for _after_ we read in the RR in order to return a dictionary where the key is a SNOMED and the value is "Reportable" (or a simple list of SNOMED codes, whatever makes the most sense).

refiner/eCR-CDA-Notes.md
robertmitchellv, line 1: i didn't get any other feedback from other folks on the team so we can probably just make this available and change it later if there's a need. it's still focused on eICR 1.1, which will be the focus until we are officially as a team working on supporting 3.1. if it would be preferred to add the 3.1 info in now i can also do that

jakewheeler (approved):
Great work on this!

PR #47, opened 2025-04-15
feat: add endpoint that accepts a zip file containing an rr and eicr

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary

- Add an endpoint that accepts a zip file to the refiner api
- Unzips the files and assigns it to a local variable 
- Add error handling if there is issue with either file 
- Add test for zip endpoint

## 🔗 Related Issue #16 

Fixes # 
- #16 

## 🧪 How to test

Using postman, hit `http://localhost:8080/zip-upload`  with body containing file that is set to a zip file containing and RR and eICR. I used the MonMothma files for this. Ensure a refined ecr is returned from the refiner


refiner/app/main.py
```diff
@@ -64,6 +67,138 @@ async def health_check():
     return {"status": "OK"}
 
 
+@app.post(
+    "/zip-upload",
+    status_code=200,
+    summary="Refine eCR from ZIP",
+)
+async def refine_ecr_from_zip(
+    file: UploadFile = File(...),
```
jakewheeler, line 76: What does the `File(...)` part do?

refiner/app/main.py
```diff
@@ -64,6 +67,138 @@ async def health_check():
...
+                    return Response(
+                        content=f"Failed to decode {filename}: {str(e)}",
+                        status_code=status.HTTP_400_BAD_REQUEST,
+                    )
+
+            if not eicr_xml:
+                return Response(
+                    content="CDA_eICR.xml not found in ZIP.",
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                )
```
jakewheeler, line 149: It might be nice to have this file processing in its own function which could return the `eicr_xml` and `rr_xml`. We'd need to untangle this code from the error handling responses, but that might not be a bad idea anyway. What do you think?

lina-roth: We could, it will be more complicated to bubble up specific error responses for xml processing

lina-roth: if you feel strongly about it I can update it!

jakewheeler: That's true. It's probably alright to leave it for now. It'd be great to clean all of the handlers up but I'm not familiar with the best way to approach that in FastAPI yet. 

I'd say go ahead and leave it as is for now unless you have some ideas on a different way to go about it!

lina-roth: Do you mean the description part on the params? I jsut copied that from the first one so I didnt have to scroll back and forth to see the values :) probably could put it at the top of the file or something

jakewheeler: No, that's fine as is! When I say "clean up handlers" I was more so meaning "remove logic from handlers" but your comment about error handling made me think there's probably a higher-level way to go about doing this. I'd rather do a bit of research before making any changes. Let's keep what you have now as it is 👍

jakewheeler (approved):
Thanks for getting this in place!

robertmitchellv (approved):
great foundation for being able to work with both files; thanks for getting this over the finish line Lina 🎉

PR #58, opened 2025-04-23
docs: Update `refiner-config.json` with `"value_code"` for RR information

### LINKED ISSUE #57: [TASK] make corrections to `refiner-config.json`
## 🚧 What needs to be done

There are a couple of small issues with the RR portion of the `refiner-config.json` file. The header needs to have `"contains"` filled in and we need to extend the data model to have a `"value"` field that is either a map or `null`.

## 🤔 Why it needs to be done

In places in the RR where we would look for the `RR1` code, it will also have a `<value>` tag that will describe the reportability status with codes like:

- `RRVS2`: May be reporable
- `RRVS3`: Not reportable
- `RRVS1`: Reportable
- `RRVS4`: No rule met

given this slightly different set up for the RR, having a `value` field in the json will better encode all of this information.

## 🏗️ Acceptance Criteria

- [ ] ...

## ⚙️ Technical Notes

- ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

There are patterns in the RR that differ slightly from the eICR in connection to coded elements that have associated values that needed to be represented by the `refiner_config.json` file.

For example, in the RR where we would look for the `RR1` code, it will also have a `<value>` tag that will describe the reportability status with codes like:

- `RRVS2`: May be reporable
- `RRVS3`: Not reportable
- `RRVS1`: Reportable
- `RRVS4`: No rule met

These have been filled out. Additionally the `"contains"` list for some of the RR entries have better linking for representing how these templates are connected.

## 🔗 Related Issue

Fixes #57
- #57

### DIFF
--- a/refiner/assets/refiner_config.json
+++ b/refiner/assets/refiner_config.json
@@ -60,7 +60,7 @@
   },
   "2.16.840.1.113883.10.20.22.2.22.1:2015-08-01": {
     "type": "section",
-    "display_name": "Encounters",
+    "display_name": "Encounters Section",
     "spec": "eICR",
     "version": ["1.1", "3.1"],
     "required": true,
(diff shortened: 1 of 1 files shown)


jakewheeler (approved):
Thanks for catching and correcting this! ✅

PR #51, opened 2025-04-18
feat: Install and configure client dependencies

jakewheeler:
# 🔀 PULL REQUEST

There's a good bit of client-side work to set up initially, so this is primarily a continuation of the work completed in #32.

## 💡 Summary

My goal with this PR is to:

- Get client-side dev tooling in place
  - Styling
  - Data fetching
  - Routing
  - Linting & formatting
  - Testing
  - Workspace setting defaults (for VS Code users)
- Provide a small demo that interacts with the Refiner API
- Create GitHub Actions to run against client code
- Provide a sample client test
- Put `/api/demo` routes in place for [this view](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=119-302&m=dev)
  - `/api/demo/upload` - Uses a sample zip file stored on the server and passes it to the upload function
  - `/api/demo/download` - Returns the sample zip file stored on the server 
- Get started on a few of the designs in [Figma](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=119-80&p=f&t=OJ8plU1gl2JTRys6-0)
  - Note: these need more work! I was primarily testing that we could get things looking somewhat correct with what's all been included in this PR 

## 🔗 Related Issue

Fixes #50 , #37

## ✅ Acceptance Criteria

See #50.

## 🧪 How to test

There are a few new client-side dependencies so you'll want to make sure they are installed. Running `docker compose up` should do that for you.

### Sample API interactions & Designs

1. Start up containers and navigate to [http://localhost:8081](http://localhost:8081)
2. Use the `demo` link to navigate to [http://localhost:8081/demo](http://localhost:8081/demo)
3. You'll see some sample designs here. Scroll to the bottom one and click the "Run test" button
4. You'll see the `Refined eICR` text area below populate with the refined Mon Mothma data - this is running that sample data through the zip upload process (`/api/demo/upload`)
5. Click on the `Download test file` link, which should download `monmothma.zip` (`/api/demo/download`)

### Tooling

In `/client`, you can now run a few new commands:
- `npm run lint`
- `npm run lint:check`
- `npm run fmt`
- `npm run fmt:check`

The `check` commands are primarily used in CI to ensure our automated code quality checks are passing. Running `lint` or `fmt` will automatically lint and format, respectively.

## ℹ️ Additional Information

[...]

PR #59, opened 2025-04-24
refactor: Separate API routes out by functionality

### LINKED ISSUE #56: [TASK] Separate API routes out by functionality
## 🚧 What needs to be done

All API routes currently exist in `main.py`. As the application grows, more and more API routes will need to be added and we want to keep the project as organized as possible. One way we can do this is by breaking out API routes by functionality and grouping them together.

For example, we will now have `/demo` related routes that are there to support demo site functionality. These routes can be placed into `/app/routes/demo.py`. The `demo.py` file can export its own router, which can then be used by the app router in `main.py`. It might look something like this:

```
from .routes import demo
...
router = APIRouter(prefix="/api")
router.include_router(demo.router)
...
app.include_router(router)
```

## 🤔 Why it needs to be done

- Begin reducing `main.py` to being primarily used for FastAPI setup/config
- Group routes will help us better organize as the application grows
- Simplifies route management

## 🏗️ Acceptance Criteria

- [ ] All `/demo` related routes are moved to their own file (`/app/routes/demo.py`)
- [ ] Demo-related router is configured and used by the FastAPI app
  - [ ] Demo-related router uses a `/demo` prefix
- [ ] Demo routes are accessible and continue to function as they did before
- [ ] Tests are updated as needed and continue to pass

## ⚙️ Technical Notes

N/A

## 📜 Additional context

- Non-demo related routes can continue to live in `main.py` for now. We should likely decide at a different point on where we want these to go

jakewheeler:
# 🔀 PULL REQUEST

**Please note the base branch is** #51 

## 💡 Summary

This PR is mainly focused on beginning to organize the new `/demo` routes that were added to the application.

- demo functionality was moved into `/routes/demo.py`
- the demo file exports a router used by the FastAPI app
- demo related tests were moved into `/tests/routes/test_demo.py`

## 🔗 Related Issue

Fixes #56 

## ✅ Acceptance Criteria

See #56 

## 🧪 How to test

1. Ensure `/api/demo/download` functions as it did previously
2. Ensure `/api/demo/upload` functions as it did previously

## ℹ️ Additional Information

Any other considerations to make at this point?


refiner/app/utils.py
```diff
@@ -69,3 +74,47 @@ def create_clinical_services_dict(
                     transformed_dict[shorthand_system] = []
                 transformed_dict[shorthand_system].extend(entry.get("codes", []))
     return transformed_dict
+
+
+async def read_zip(file: UploadFile) -> tuple[str, str]:
```
jakewheeler, line 79: Is there a better place to put this? I was thinking we could start a `services` directory that is a sibling to `routes`  which could have files with reusable functions but wasn't sure if I should do that in this PR, or if we even want to go that direction. 

Let me know if you have any thoughts!

refiner/app/utils.py
```diff
@@ -69,3 +74,47 @@ def create_clinical_services_dict(
...
+
+            if not eicr_xml:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="CDA_eICR.xml not found in ZIP.",
+                )
+
+            return eicr_xml, rr_xml
+    except BadZipFile:
+        raise HTTPException(
```
jakewheeler, line 117: I think it'd be preferable if this function wasn't throwing exceptions that come from FastAPI directly. However, this function currently relies on FastAPI's `UploadFile` too. I think it'd be best if reading an eCR from a zip didn't rely on FastAPI at all so that there's no issues when we go to (potentially) deploy to Lambda.

Maybe there's nothing we need to do right this second, but this is something I think will need to change at some point.

refiner/app/utils.py
robertmitchellv: would you want things like `refine.py` and others to live in this directory as well? something like

`from services.refine import function`?

or were you thinking of something else?

refiner/app/utils.py
robertmitchellv: yeah, i think that's a great idea; FastAPI should just be handling the ingress of the zip and the processing should be something that `lxml` is doing on the python side of things. i think a backlog ticket for this is a great idea 💡

refiner/app/utils.py
jakewheeler: yeah, something along those lines! I'm thinking anything in `services` could be used easily by API routes or elsewhere (such as Lambda).

refiner/app/utils.py
jakewheeler: Awesome, thanks for your input! I will write up a ticket for this.

robertmitchellv (approved):
great work Jake 🙌🏻 everything worked locally; file download, refining, branding looks great, pytest passing 🎯

refiner/app/utils.py
robertmitchellv: any structure that helps us decouple from fastapi more and get ready for lambda i would be in favor of--happy to figure out the right way to get that started

PR #72, opened 2025-05-01
[feat] add method to parse RR and extract reportable condition snomed code

### LINKED ISSUE #29: [TASK] Update Refiner to read and parse the Reportability Response document
## 🚧 What needs to be done

One key piece missing from the Refiner's functionality today is the ability to read and parse the Reportability Response (RR) XML document. Currently, the Refiner is only able to read and parse the eICR document.

We will need to wait on the completion of #16 to begin this work. Once that ticket is finished, the Refiner will have access to the RR via the zip file that is uploaded to the application.

## 🤔 Why it needs to be done

We plan to make use of the RR's content in order to determine which condition(s) are deemed reportable. With this information available to the Refiner, we should be able to begin splitting the eICR based on this info.

## 🏗️ Acceptance Criteria

TBD for full acceptance criteria

- [ ] Refiner code has been modified to read the RR XML
- [ ] Code has been written to find reportable conditions
- [ ] All reportable conditions and any other required information is stored (in an in-memory data structure like a dictionary, not a database) for usage

A new ticket should be created to begin splitting the eICR based on the reportable conditions.

## ⚙️ Technical Notes

N/A

## 📜 Additional context

Should anything else be included in this work?

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary
- Update /zip-upload endpoint to parse RR
- Once RR is parsed, it will extract the reportable snomed codes 
- If conditions_to_include is not passed in from caller, it will set it as the reportable snomed code found in the RR 

## 🔗 Related Issue

Fixes #29 
- #29 

## ✅ Acceptance Criteria

-  Refiner code has been modified to read the RR XML
-  Code has been written to find reportable conditions
 - All reportable conditions and any other required information is stored (in an in-memory data structure like a dictionary, not a database) for usage

## 🧪 How to test

Using postman, hit http://localhost:8080/zip-upload with body containing file that is set to a zip file containing and RR and eICR. I used the MonMothma files for this. Ensure a refined ecr is returned from the refiner

## ℹ️ Additional Information

- Started to play around with checking for multiple snomed codes in the RR and checking the RRSV1 to see if displayName="Reportable" but was having some issues. Mainly having no RR to refer to that has multiple reportable conditions and RRSV1 values of Reportable and Not Reportable. So this will require some further digging although I assume Robert has some ideas

### DIFF
--- a/refiner/app/rr_parser.py
+++ b/refiner/app/rr_parser.py
@@ -1,0 +1,77 @@
+import xml.etree.ElementTree as ET
+from fastapi import Response, status
+from typing import Union
+
+NAMESPACES = {
+    'cda': 'urn:hl7-org:v3',
+    'sdtc': 'urn:hl7-org:sdtc',
+    'voc': 'http://www.lantanagroup.com/voc'
+}
+
+def parse_xml(rr_xml: str) -> Union[ET.Element, Response]:
+    """
+    Parses a raw RR XML string and returns the root Element.
+
+    Args:
+        rr_xml (str): The raw XML string.
+
+    Returns:
+        ElementTree.Element if successful, or FastAPI Response on error.
+    """
+    try:
+        rr_root = ET.fromstring(rr_xml)
+        return rr_root
+    except ET.ParseError as e:
+        return Response(
+            content=f"Failed to parse RR XML: {str(e)}",
+            status_code=status.HTTP_400_BAD_REQUEST,
+        )
+
+# def get_reportable_conditions(root):
+#     """
+#     Returns a comma-separated list of all SNOMED CT codes found
+#     in the Report Summary section (code 55112-7), or None if none.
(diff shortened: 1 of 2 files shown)


refiner/app/rr_parser.py
```diff
@@ -0,0 +1,49 @@
+import xml.etree.ElementTree as ET
```
robertmitchellv, line 1: i think for consistency we should stick to `lxml` for our xml processing instead of the built in `xml` element tree but this can wait for later.

robertmitchellv (approved):
🙏🏻 thanks for this; we now have a foundation that we can build on for RR processing

i tested all of the zip uploads with and without query params and everything worked as expected (i'll push updates to `docs/refine/` later). all tests passed locally ✨

PR #76, opened 2025-05-01
docs: add new multipart form zip demo requests to http file

### LINKED ISSUE #75: [TASK] update `refiner-demo-requests.http` in `docs/refiner`
## 🚧 What needs to be done

As soon as #72 is merged we will need to add some examples of using the multipart form in the `http` file. We will also need a `zip` with both a :

- `CDA_eICR.xml` file and
- `CDA_RR.xml` file

## 🤔 Why it needs to be done

The demo files should be kept up to date as we extend and add features/functionality to the codebase

## 🏗️ Acceptance Criteria

- [ ] at least two examples of a multipart form request with a zip file that has no query params and another that has at least 1 query param used.

## ⚙️ Technical Notes

- ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Once #72 is merged, this will add files to test with the refiner endpoints via `.http` file including the zip that extracts the SNOMED condition and passes the value to `conditions_to_include`.

## 🔗 Related Issue

Fixes #75 
- #75 

## ✅ Acceptance Criteria

- [x] at least two examples of a multipart form request with a zip file that has no query params and another that has at least 1 query param used.

## 🧪 How to test

Using your `http` tool of choice (in my case, neovim's `kulala-nvim`), open the file and run through the commands.


jakewheeler (approved):
Tested all endpoints without issue ✅  Thanks for updating these!

PR #78, opened 2025-05-06
refactor: Decouple application logic from FastAPI

### LINKED ISSUE #67: [TASK] Decouple application logic from FastAPI
## 🚧 What needs to be done

During the review of #59 we decided it would be ideal to begin the process of decoupling the application logic from the FastAPI routes. What this means in practice is that FastAPI routes not contain any core business logic besides taking input, sending the input to functions for processing, and returning a response.

## 🤔 Why it needs to be done

The goal of this work is to ensure that the application is modular enough to be used out of a FastAPI environment. Ideally, we should be able to re-use the vast majority of the code we're writing in AWS Lambda, for example.

The work completed in this ticket will help us to establish a pattern for creating future services as well.

## 🏗️ Acceptance Criteria

- [ ] Create a `services` directory within `/refiner/app`
- [ ] Move any existing `*.py` files containing core app logic into `services` (as of now, this appears to encompass `db.py` and `refine.py`)
- [ ] Create new `*.py` files within `services` as needed and move existing related functionality into these files. For example, it may make sense to create `zipfile.py` (or something similar) to handle all functionality related to working with zip files
- [ ] Remove any FastAPI specific error handling from functions within files in `services` and ensure API errors are handled by the routes instead
- [ ] Clean up the API route code (if needed)
- [ ] Tests are modified as needed

## ⚙️ Technical Notes

Using service functionality in a route file will most likely end up looking something like this:

```python
from ..services.example import test_function
...
...
```

## 📜 Additional context

This ticket is fairly prescriptive but doesn't necessarily need to be! If you have other opinions on how to go about this work, please feel free to run them by the team and decide on a different path forward.

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

The goal of this work is to ensure that the application is modular enough to be used out of a FastAPI environment. Ideally, we should be able to re-use the vast majority of the code we're writing in AWS Lambda, for example.

The proposal is the following:

```
.
├── app
│   ├── api
│   │   └── v1
│   │       ├── demo.py
│   │       ├── ecr.py
│   │       ├── file_io.py
│   │       └── v1_router.py
│   ├── core
│   │   ├── app
│   │   │   ├── base.py
│   │   │   └── openapi.py
│   │   ├── config.py
│   │   ├── exceptions.py
│   │   └── models
│   │       ├── api.py
│   │       └── types.py
│   ├── main.py
│   ├── services
│   │   ├── db.py
│   │   ├── file_io.py
│   │   ├── refine.py
│   │   └── terminology.py
│   └── tes.db
```

- the routes will go in `/api`
  - the routes are versioned and bundled in `v1_router`
- things that are shared in both the services and routes are in `/core`
  - `/models` contains the `api` and `types` modules
  - `/app` contains the base service `base.py` and the openapi config
  - there is also a core `config` and an `exceptions` module that is the interface between routes and services so that FastAPI can be decoupled from the functions in `/services`
- services are in `/services`
  - the main functionality of the back end code
- `main.py` is in the `app` directory (complexity reduced)
- the DB is in the `app` directory

## 🔗 Related Issue

Fixes #67 
- #67 

## ✅ Acceptance Criteria

- [x] Create a `services` directory within `/refiner/app`
- [x] Move any existing `*.py` files containing core app logic into `services` (as of now, this appears to encompass `db.py` and `refine.py`)
- [x] Create new `*.py` files within `services` as needed and move existing related functionality into these files. For example, it may make sense to create `zipfile.py` (or something similar) to handle all functionality related to working with zip files
- [x] Remove any FastAPI specific error handling from functions within files in `services` and ensure API errors are handled by the routes instead
- [x] Clean up the API route code (if needed)
- [x] Tests are modified as needed

## 🧪 How to test

The unit tests have been significantly over-hauled an `pytest-cov` has been added to the `dev-requirements.txt`. You can run:

```bash
pytest --cov=app tests
```

These are the results (we probably want to eventually add an action for this):

```
Name                          Stmts   Miss  Cover
-------------------------------------------------
app/__init__.py                   0      0   100%
app/api/__init__.py               0      0   100%
app/api/v1/__init__.py            0      0   100%
app/api/v1/demo.py               40      3    92%
app/api/v1/ecr.py                47      4    91%
app/api/v1/file_io.py            24     17    29%
app/api/v1/v1_router.py           8      0   100%
app/core/__init__.py              0      0   100%
app/core/app/base.py             29      5    83%
app/core/app/openapi.py          16      5    69%
app/core/config.py               10      1    90%
app/core/exceptions.py           39      0   100%
app/core/models/api.py           13      0   100%
app/core/models/types.py         15      1    93%
app/main.py                      21      1    95%
app/services/__init__.py          0      0   100%
app/services/db.py               60      8    87%
app/services/file_io.py          51      0   100%
app/services/refine.py          174      5    97%
app/services/terminology.py      26      4    85%
-------------------------------------------------
TOTAL                           573     54    91%
=============================================================================== 57 passed in 1.98s ===============================================================================

```
## ℹ️ Additional Information

This was a pretty large overhaul so if anything slipped through please let me know!


refiner/app/main.py
```diff
@@ -1,19 +1,13 @@
...
 
-from app.base_service import BaseService
-from app.db import get_value_sets_for_condition
-from app.models import RefineECRResponse
-from app.refine import refine, validate_message, validate_sections_to_include
-from app.rr_parser import get_reportable_conditions, parse_xml
-from app.utils import create_clinical_services_dict, read_json_from_assets, read_zip
-
-from .routes import demo
+from .api.v1 import demo, ecr
```
jakewheeler, line 8: What do you think about exporting a single `router` from the `v1` package that includes all `/v1` routes? Basically something like this is what I'm thinking:
```python
from .api.v1 import v1_router
...
...
router = APIRouter(prefix="/api")
router.include_router(v1_router) # /api/v1/demo/*, /api/v1/ecr/*, etc.
```
This feels like it'd be slightly easier to manage. Any downsides I'm not considering?

refiner/app/main.py
```diff
@@ -23,15 +17,12 @@
     service_path="/message-refiner",
     description_path=Path(__file__).parent.parent / "README.md",
     include_health_check_endpoint=False,
-    openapi_url="/message-refiner/openapi.json",
+    openapi_url="/openapi.json",
 ).start()
 
-router = APIRouter(prefix="/api")
+router = APIRouter(prefix="/api/v1")
```
jakewheeler, line 23: We'll need to make sure the client-side routes are updated so that the demo doesn't break. We'll need to update the URLs in these two files:
- `RunTest.tsx`
- `demo.ts`

This is a good reminder to me that we should consider some basic e2e tests at some point to catch little things like this. Some centralized URL organization in the client probably wouldn't hurt either so we don't eventually need to try hunting them all down to make changes 😄

I'll make a note to myself to write up tickets for these things.

refiner/app/services/terminology.py
```diff
@@ -0,0 +1,17 @@
+from .db import get_value_sets_for_condition
+
+
+def _get_clinical_services(condition_codes: str) -> list[dict]:
```
jakewheeler, line 4: Since this is a "private" function and currently only used in `ecr.py`, I'm thinking we move this function into that file and delete this one entirely. What do you think?

refiner/app/services/utils.py
```diff
@@ -4,17 +4,23 @@
...
 def read_json_from_assets(filename: str) -> dict:
     """
     Reads a JSON file from the assets directory.
 
-    :param filename: The name of the file to read.
-    :return: A dictionary containing the contents of the file.
+    Args:
+        filename: The name of the file to read.
+    Returns:
+        dict: Contents of the JSON file.
```
jakewheeler, line 19: Does this functionally change anything? Like the generated `openapi` docs? I don't really have a personal preference when it comes to the new vs. old docstrings, mostly just curious.

refiner/app/main.py
robertmitchellv: i think that's a great idea; i'll add that to this PR 🙏🏻 thanks for the suggestion

refiner/app/services/utils.py
robertmitchellv: i think there was one way were were doing this originally for the `/redoc` endpoint that had a specific style that i realized i'm not using here. i was defaulting to google style for docstrings but after looking at the [fastapi docs](https://fastapi.tiangolo.com/tutorial/path-operation-configuration/?h=docstring#description-from-docstring) it seems like what we should really be using is markdown, which i actually quite like. it'd look like this:

```python
from fastapi import FastAPI
from pydantic import BaseModel

app = FastAPI()


class Item(BaseModel):
    name: str
    description: str | None = None
    price: float
    tax: float | None = None
    tags: set[str] = set()


@app.post("/items/", response_model=Item, summary="Create an item")
async def create_item(item: Item):
    """
    Create an item with all the information:

    - **name**: each item must have a name
    - **description**: a long description
    - **price**: required
    - **tax**: if the item doesn't have tax, you can omit this
    - **tags**: a set of unique tag strings for this item
    """
    return item
```

maybe i can write up a ticket for adding the formatting to the ruff to enforce and then i can merge it back in here? how does that sound?

refiner/app/services/terminology.py
robertmitchellv: this may be a bit of prework on my end; i will end up adding a lot more to this piece in the next ticket where i migrate our database to the new model. i can remove or just pick it back up in the next ticket

refiner/app/main.py
robertmitchellv: ah right. i wasn't thinking about the UI at all. i'll make sure to update in those places. and e2e tests sound like a great plan. i'll make sure to test it after i make the changes

refiner/app/services/utils.py
jakewheeler: yeah, good idea! I like the markdown format as well.

refiner/app/services/utils.py
```diff
@@ -4,17 +4,23 @@
...
     """
-    return json.load(open(pathlib.Path(__file__).parent.parent / "assets" / filename))
+    return json.load(
+        open(pathlib.Path(__file__).parent.parent.parent / "assets" / filename)
```
jakewheeler, line 22: We'll have to add an additional `.parent` for `_get_demo_zip_path()` in `demo.py` as well 🥲 

I wonder if we should make a more generic "get full file path from assets directory" function to avoid having to change this in multiple places in the future?

There might be a better way to handle this but this is roughly what I'm thinking:

```python
# utils.py
def get_asset_path(subpath: Path) -> Path:
    """
    Returns the full path of a file starting from the assets directory
    """
    return Path(__file__).parent.parent.parent / "assets" / subpath


def read_json_from_assets(filename: str) -> dict:
    """
    Reads a JSON file from the assets directory.

    Args:
        filename: The name of the file to read.
    Returns:
        dict: Contents of the JSON file.
    """
    return json.load(open(get_asset_path(filename)))

# demo.py
def _get_demo_zip_path() -> Path:
    return get_asset_path(Path("demo") / "monmothma.zip")
```

refiner/app/services/terminology.py
robertmitchellv: that sounds good to me

refiner/app/services/utils.py
```diff
@@ -78,22 +94,27 @@ def create_clinical_services_dict(
 
 async def read_zip(file: UploadFile) -> tuple[str, str]:
```
jakewheeler, line 95: I don't know that it necessary needs to be handled in this PR but we should likely update this function to accept a standard Python file/file-like type instead of FastAPI's `UploadFile` so that we can easily make use of it for zip files that come from elsewhere (S3?).

It looks like there is an `UploadFile.file` which will give us the underlying `BytesIO`. It may make sense to pass that into `read_zip()` from the handler instead of the `UploadFile` 🤔

refiner/app/services/utils.py
robertmitchellv: yeah that's a good point; since we have `/assets` at the root next to `/app` and `/tests` as well as `/tests/assets` i might want to use something like this for tests too. it seems like there will be an overlap of sample data that the demo will need and that tests will also need but i'm not sure about centralizing one `/assets` directory with sub-directories for "demo" or "sample-data" or leaving them the way that they are with a little overlap.

i do want to add something like this though!

refiner/app/services/utils.py
robertmitchellv: that's a really good point. we really shouldn't have a fastapi class in `/services` anyway. i think i'll add a file to `/core` that can provide a kind of bridge between services and the routes. i'll see if i can add that too

refiner/app/main.py
robertmitchellv: it'd be great monday to go over this; i don't think the download/upload is working correctly

jakewheeler: It looks like the path returned by `_get_demo_zip_path()` is wrong. This works for me:
```python
def _get_demo_zip_path() -> Path:
    return (
        Path(__file__).parent.parent.parent.parent / "assets" / "demo" / "monmothma.zip"
    )
```

refiner/tests/routes/test_demo.py
```diff
@@ -1,118 +1,115 @@
...
+from fastapi import Response
 from fastapi.testclient import TestClient
 
 from app.main import app
-from app.routes.demo import _get_demo_zip_path, _get_file_size_difference_percentage
 
 client = TestClient(app)
 
-api_route_base = "/api/demo"
+api_route_base = "message-refiner/api/v1/demo"
```
jakewheeler, line 12: Why is `message-refiner/` being included?

refiner/tests/routes/test_demo.py
```diff
@@ -1,118 +1,115 @@
...
-    assert result == expected
+def test_file_size_difference_percentage(
+    unrefined: str, refined: str, expected: int
+) -> None:
+    """
+    Test the file size difference calculation function
+    """
 
+    # Test the function logic directly without importing
+    def calculate_size_diff(unrefined_doc: str, refined_doc: str) -> int:
```
jakewheeler, line 31: Why not import the function? If we break the real function this test will still continue to pass.

refiner/tests/test_api_integration.py
```diff
@@ -0,0 +1,169 @@
...
+        response = client.get("/message-refiner/healthcheck")
+        assert response.status_code == 200
+        assert response.json() == {"status": "OK"}
+
+    def test_openapi_docs(self):
+        response = client.get("/message-refiner/api/openapi.json")
+        assert response.status_code == 200
+
+
+class TestECREndpoint:
```
jakewheeler, line 25: Cool, didn't realize we could group tests inside classes like this.

refiner/tests/test_service_file_io.py
```diff
@@ -0,0 +1,182 @@
+from zipfile import ZipFile
```
jakewheeler, line 1: Would it make sense to group this file and the other `test_service_*` files within `refiner/tests/services`? It seems like previously we were mimicking the `app` structure within `tests` in order to do our grouping.

refiner/app/services/refine.py
```diff
@@ -22,62 +24,98 @@
...
-    validated_message: etree.Element,
+    Returns:
+        dict: Extracted information from the RR document
+    """
+
+    rr_root = xml_files.parse_rr()
+
+    return {
+        "reportable_conditions": get_reportable_conditions(rr_root),
+    }
```
jakewheeler, line 114: Why return a `dict` and not `get_reportable_conditions(rr_root)` directly?

refiner/app/api/v1/demo.py
```diff
@@ -52,31 +54,41 @@ async def demo_upload(file_path: Path = Depends(_get_demo_zip_path)) -> JSONResp
...
+                {
+                    "unrefined_eicr": xml_files.eicr,
+                    "refined_eicr": refined_eicr,
+                    "reportable_conditions": rr_results["reportable_conditions"],
+                    "stats": [
+                        f"eCR file size reduced by {
+                            _get_file_size_difference_percentage(
+                                xml_files.eicr, refined_eicr
+                            )
+                        }%",
```
jakewheeler, line 73: In `main` this output shows 38% and in this PR it shows 49%. It seems like the files are exactly the same. What am I missing here?

refiner/app/main.py
```diff
@@ -1,75 +1,26 @@
...
-# Instantiate FastAPI via DIBBs' BaseService class
-app = BaseService(
-    service_name="Message Refiner",
-    service_path="/message-refiner",
-    description_path=Path(__file__).parent.parent / "README.md",
-    include_health_check_endpoint=False,
-    openapi_url="/message-refiner/openapi.json",
-).start()
+# service router
+service_router = APIRouter()
```
jakewheeler, line 15: I would be in favor of deleting this router entirely and sticking `/healthcheck` directly on `router`. I don't know that there's much value (at least currently) in having this additional layer.
```python
@router.get("/healthcheck")
async def health_check() -> dict[str, str]:
    """
    Check service health status.

    Returns:
        dict[str, str]: Service status response:
            - {"status": "OK"} with HTTP 200 if service is healthy
    """

    return {"status": "OK"}


app.include_router(router)
```

The other thing is that we are going to want every API route behind `/api` so that routes meant for the server don't interfere with routes that should be handled by React Router on the client. Right now we can successfully query `http://localhost:8080/healthcheck` which is not ideal.

refiner/app/main.py
robertmitchellv: ah gotcha; i'll take a look at the `file_io` function again. i was trying to get it to work anywhere so we don't have to keep editing paths.

refiner/tests/test_service_file_io.py
```diff
@@ -0,0 +1,182 @@
...
+    assert root.tag.endswith("ClinicalDocument")
+
+
+def test_parse_xml_invalid():
+    """
+    Test parsing invalid XML raises appropriate error.
+    """
+
+    with pytest.raises(XMLValidationError):
+        file_io.parse_xml("<not>valid</xml>")
```
jakewheeler, line 44: Should we be asserting something here?

refiner/tests/routes/test_demo.py
robertmitchellv: i ended up adding `message-refiner` to the base class; so the other urls have that added in now as well. i thought it made sense to do that but if that's not a common thing to do i can remove it

refiner/tests/routes/test_demo.py
robertmitchellv: oh i was supposed to remove that; i was trying to fix something that wasn't working with conftest. i'll take care of that now!

refiner/tests/test_service_file_io.py
robertmitchellv: i like the idea of having a test for a specific file and i also like the idea of having files that aren't too long. i also thought `test_refine` and `test_refiner` were too similar. maybe we can chat about that?

refiner/app/core/models/types.py
```diff
@@ -0,0 +1,76 @@
...
+
+    eicr: str
+    rr: str
+
+    def parse_eicr(self) -> etree._Element:
+        """
+        Parse eICR content into XML element tree.
+        """
+
+        from ...services.file_io import parse_xml
```
jakewheeler, line 65: Why import here instead of at the top of the file?

refiner/app/api/v1/demo.py
```diff
@@ -52,31 +54,41 @@ async def demo_upload(file_path: Path = Depends(_get_demo_zip_path)) -> JSONResp
...
+        xml_files = await file_io.read_xml_zip(upload_file)
+        rr_results = refine.process_rr(xml_files)
+        refined_eicr = refine.refine_eicr(xml_files)
+
+        return JSONResponse(
+            content=jsonable_encoder(
+                {
+                    "unrefined_eicr": xml_files.eicr,
+                    "refined_eicr": refined_eicr,
+                    "reportable_conditions": rr_results["reportable_conditions"],
```
jakewheeler, line 67: Nice 👀

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+}
+```
+
+## Development
+
+### Required Environment Variables
+- `APP_VERSION`: Application version (default: "1.0.0")
+- Additional configuration as needed
```
jakewheeler, line 149:
```suggestion
### Required Environment Variables

- `APP_VERSION`: Application version (default: "1.0.0")
- Additional configuration as needed
```

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+            "additional": "context"
+        }
+    }
+}
+```
+
+## Development
+
+### Required Environment Variables
+- `APP_VERSION`: Application version (default: "1.0.0")
```
jakewheeler, line 148: I realize this is not new, but I am wondering how this should work. Is the intention to set `APP_VERSION` during a build which will then be reflected in the Docker image?

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+│   │   ├── api.py       # API models and schemas
+│   │   └── types.py     # Core type definitions
+│   ├── config.py        # Core configuration
+│   └── exceptions.py    # Exception hierarchy
+└── services/
+    ├── db.py            # Database logic
+    ├── file_io.py       # File processing logic
+    ├── refine.py        # XML refinement logic
+    └── terminology.py   # Clinical terminology services
+```
```
jakewheeler, line 45: IMO a more general README per directory (`api`, `core`, `services`) that describes its usage feels simpler to maintain. I can see this being neglected over time 😓

jakewheeler: Additionally, aside from the comments, folks are able to see this already by browsing the project itself. Giving more detail about each package feels like the main piece of value we want to keep around.

refiner/app/main.py
robertmitchellv: that sounds good to me; i can make that change

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+│   ├── InputValidationError
+│   ├── XMLValidationError
+│   └── SectionValidationError
+├── ProcessingError
+│   ├── FileProcessingError
+│   └── XMLProcessingError
+└── ResourceError
+    ├── ResourceNotFoundError
+    └── ResourceAccessError
+```
```
jakewheeler, line 100: Similar to my other comment, I'd probably remove this due to the likelihood it'll be neglected as changes happen. If we wanted to make a point of this, I'd recommend linking to the `exceptions.py` file directly.

However, I'd argue error handling is more of an expectation rather than a Refiner feature, so we could potentially just axe this section 😅

refiner/app/README.md
robertmitchellv: i honestly don't remember (it might've been added after we stopped working on the refiner) but i think it had something to do with the old `phdi` and getting everything built and available via `orchestration`. i'm not sure what impact it would have on our deployment needs

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+   - Section filtering based on LOINC codes
+   - Clinical service code matching
+   - Template ID validation
+   - XML tree transformation
+
+3. **Output Generation**
+
+   - Updated text element creation
+   - Minimal section generation for excluded content
+   - Final XML document assembly
```
jakewheeler, line 82: This is good!

refiner/app/README.md
robertmitchellv: yeah that's a fair point. i think i mostly just wanted to jot this since there are so many big changes. it doesn't necessarily need to be a part of the codebase and i'm fine with either removing it or having specific READMEs in `/api`, `/core`, and `/services`

refiner/app/README.md
robertmitchellv: that's fine--there are also a lot of exceptions i put in there that i'm not explicitly using yet but i wanted to try and think of the patterns we'd need as we build out. and i expect they'll change quite a bit too

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+The application uses Python's Protocol system for flexible interfaces:
+
+```python
+@runtime_checkable
+class FileUpload(Protocol):
+    async def read(self) -> bytes: ...
+```
+
+This allows for multiple implementations (FastAPI's UploadFile, S3 files, etc.)
```
jakewheeler, line 59: I would actually probably axe this as well 🤔 feels like an implementation detail rather than a "key feature". 

Maybe we rework `Key Features` to include only the content of `2. XML Processing Pipeline` since that's really where the core value is. What do you think?

I could definitely see us add something later about supporting multiple deployment types, but maybe it's too soon to make a point of that.

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+2. **Direct XML** (`POST /api/v1/ecr`)
+
+   - Accepts raw XML eICR content
+   - Supports section and condition filtering
+   - Returns refined XML document
+
+### Query Parameters
+
+- `sections_to_include`: LOINC codes for section filtering
+- `conditions_to_include`: SNOMED codes for clinical services
```
jakewheeler, line 121: Should we keep this or would it be better to allow the OpenAPI docs to handle this and we mention how to access that?

refiner/app/README.md
```diff
@@ -0,0 +1,156 @@
...
+}
+```
+
+## Development
+
+### Required Environment Variables
+- `APP_VERSION`: Application version (default: "1.0.0")
+- Additional configuration as needed
+
+### Type Safety
```
jakewheeler, line 152: **(No action needed)** I'd be curious to check out [mypy](https://mypy-lang.org/) at some point.

refiner/app/core/models/types.py
robertmitchellv: i read that it solves circular imports; it's because of the way that the function from `file_io` and the `XMLFiles` class are both being loaded in. i don't have a ton of experience but i like the idea of lazy loading for things. i can add something like this maybe:

```python
class XMLFiles(NamedTuple):
    """
    Container for eICR and RR XML documents.
    
    Note:
        parse_xml is imported inside methods to avoid circular imports:
        - types.py defines XMLFiles
        - file_io.py uses XMLFiles
        - XMLFiles methods use parse_xml from file_io.py
    """
    eicr: str
    rr: str
    
    def parse_eicr(self) -> etree._Element:
        # Import inside method to avoid circular dependency
        from ...services.file_io import parse_xml
        return parse_xml(self.eicr)
```

refiner/app/core/app/base.py
```diff
@@ -120,7 +95,7 @@ async def health_check() -> StatusResponse:
             Returns:
                 StatusResponse: {"status": "OK"} with HTTP 200 if service is healthy.
             """
-            return STATUS_OK
+            return {"status": "OK"}
```
jakewheeler, line 98: We don't need to address this in this PR but we may want to remove the `include_health_check_endpoint` option and associated code since it doesn't seem like we need it.

refiner/tests/test_service_file_io.py
robertmitchellv: `pytest.raisees` acts like an assertion. we're testing that the specific exception we want gets raised. we could add something like this:

```python
assert "specific error message" in str(excinfo.value)
```

after but this would be counted as a second test. we just need to make sure the error messages match

refiner/app/README.md
robertmitchellv: probably instructions for how to access via openapi

refiner/app/README.md
robertmitchellv: yeah it's also been on my todo list as well

refiner/tests/routes/test_demo.py
jakewheeler: I don't think we need it. I think as long as our "base" is always `/api` and every other route starts there, we should be all set. That feels like the main thing that matters.

refiner/tests/routes/test_demo.py
jakewheeler: I figured it was something like that 😆 Thanks!

refiner/app/README.md
robertmitchellv: folded this in; i think it's a much better idea so thanks for that!

refiner/tests/test_service_file_io.py
jakewheeler: Oh yeah, I've had that exact same thought about `test_refine` and `test_refiner`. I wouldn't say I have too strong of an opinion on _how_ we organize our tests as long as we keep it consistent as they change over time.

refiner/tests/test_service_file_io.py
jakewheeler: Got it! I wasn't sure if this was truly doing a check or if `with pytest.raises(XMLValidationError):` does what it's supposed to do and we simply always make it to the end of the test function so it passes without really doing anything.

refiner/app/services/refine.py
robertmitchellv: i like the idea of wrapping the RR processing because we're going to make changes to what gets called inside this function and i think we should return an object that is the input for other functions, like refine, that might need more structure. i can see this eventually looking like this:

```python
return {
        "reportable_conditions": get_reportable_conditions(rr_root),
        "trigger_codes": get_trigger_codes(rr_root),
        "reporting_timeframe": get_reporting_timeframe(rr_root),
        "jurisdiction_data": get_jurisdiction_info(rr_root),
    }
```

refiner/tests/test_service_file_io.py
robertmitchellv: i'm open to rethinking this as well; in one of my info science courses we read a paper where the author said that people tend to either be _lumpers or splitters_. i'm on the _splitter_ side of things. i like things split up with more granularity so that's really what is important to me in the organizing

refiner/tests/routes/test_demo.py
jakewheeler: Outcome from our discussion: We'll remove `message-refiner/` from the paths.

refiner/tests/test_service_file_io.py
jakewheeler: Outcome from our discussion: we'll leave these files as they are for now but will follow up as a team to see how we want to structure them longer term.

refiner/app/services/refine.py
jakewheeler: Outcome from our discussion: we'll leave this as is due to what Robert mentioned above.

refiner/app/api/v1/demo.py
jakewheeler: Outcome from our discussion: Code was accidentally modified and will be fixed.

refiner/app/README.md
jakewheeler: Outcome from our discussion: Not something we need to worry about right now.

client/package-lock.json
```diff
@@ -48,15 +48,12 @@
       "version": "4.4.2",
```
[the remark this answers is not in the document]
robertmitchellv: this is fixed now!

refiner/app/main.py
robertmitchellv: this is fixed now as well; was able to download the file and move through the demo flow

refiner/tests/routes/test_demo.py
robertmitchellv: this is resolved now!

refiner/app/api/v1/demo.py
robertmitchellv: ok, i wasn't 100% sure why this was happening but i figured it out; in the `app/services/file_io.py` module we're parsing like this:

```python
parser = etree.XMLParser(remove_blank_text=True)
```

this `remove_blank_text` is apparently pretty aggressive so it's doing more of a reduction. i removed it for now so we're back to 38% but if we want the number to go up we can put it back in

refiner/tests/routes/test_demo.py
robertmitchellv: this is fixed now!

refiner/app/main.py
robertmitchellv: this is also fixed; healthcheck is on `/api` and there is only one router in `main.py`

refiner/app/core/models/types.py
robertmitchellv: the other solutions seem to be more complicated, like dependency inversion, service locator, or package-level initialization. i think this is actual a fine way to handle this problem and i even found an example here in the django codebase: https://github.com/django/django/blob/825ddda26a14847c30522f4d1112fb506123420d/django/db/models/base.py#L2209

refiner/app/api/v1/demo.py
```diff
@@ -5,14 +5,19 @@
...
 router = APIRouter(prefix="/demo")
 
 
 def _get_demo_zip_path() -> Path:
-    return Path(__file__).parent.parent.parent / "assets" / "demo" / "monmothma.zip"
+    """
+    Get the path to the demo ZIP file.
+    """
+
+    return file_io.get_asset_path("demo", "monmothma.zip")
```
jakewheeler, line 20: nice!

jakewheeler (approved):
Great work and thanks for addressing all of my feedback! 🎉

PR #79, opened 2025-05-06
feat: file size reduction percentage calculation

### LINKED ISSUE #42: [USER STORY] Size Reduction Metric
## 👤💭 User Story

As an Informatician or eCR Ingestion Lead, I want to how refinement affects the size of my files, so that I can understand the impact of it.

## 🏗️ Acceptance Criteria

- [ ] File size reduction percentage - file size in bytes

## 🎬 Scenarios

- ...

## 🎨 Design Notes

- https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=119-167&t=0qC4EAKTb63ayLB2-1

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR enhances the response from the `/api/demo/upload` route to include a `stats` array containing information about the document that was refined. These `stats` can be rendered by the client, which is also part of this PR.

Right now, `stats` includes the refined file size reduction as well as a placeholder related to how many observations were found. It should be straightforward to add more refinement stats to the demo over time if we choose to do so.

## 🔗 Related Issue

Fixes #42 

## ✅ Acceptance Criteria

- [x] File size reduction percentage (calculated using bytes)

### Client

1. Navigate to [http://localhost:8081/demo](http://localhost:8081/demo)
2. Run through the demo until you get to the success page
3. You should see `eCR file size reduced by 38%`, which is the reduction in size coming from the API's calculating

### API

1. Make a request to `http://localhost:8080/api/demo/upload`
2. You'll receive a response back that now includes a `stats` array and looks like this

```
"stats": [
    "eCR file size reduced by 38%",
    "Found X observations relevant to the condition(s)"
]
```

## ℹ️ Additional Information

N/A


refiner/app/routes/demo.py
```diff
@@ -15,6 +15,19 @@ def _get_demo_zip_path() -> Path:
     return Path(__file__).parent.parent.parent / "assets" / "demo" / "monmothma.zip"
 
 
+def _get_file_size_difference_percentage(
```
jakewheeler, line 18: I did check this myself but feel free to double check my math 😂

refiner/app/routes/demo.py
```diff
@@ -45,7 +58,16 @@ async def demo_upload(file_path: Path = Depends(_get_demo_zip_path)):
...
-            {"unrefined_eicr": eicr_xml, "refined_eicr": refined_data}
+            {
+                "unrefined_eicr": eicr_xml,
+                "refined_eicr": refined_data,
+                "stats": [
+                    f"eCR file size reduced by {
+                        _get_file_size_difference_percentage(eicr_xml, refined_data)
+                    }%",
+                    "Found X observations relevant to the condition(s)",
+                ],
```
jakewheeler, line 69: I was debating whether or not we want to simply return an array of strings to render in the client, like I'm doing here, or return raw numbers, something more like:
```
"stats": {
    "file_size_reduction_percentage": 55,
    "observations_found": 10,
}
```

I ended up going the array of strings route since I don't think we have another use case for these values currently. I could certainly be swayed to take the alternative approach I outlined above if that makes more sense. We own both the client and server so we aren't locked into either approach if it does need to be modified later.

refiner/app/routes/demo.py
robertmitchellv: it checks out for me based on what i was looking up!

refiner/app/routes/demo.py
robertmitchellv: i can't think of any reason as to why we'd want to do any kind of client side math with the numbers so string feels right unless we have some material reason to do otherwise

robertmitchellv (approved):
everything is working for me locally; thanks for this! i think this is a great start and it'll be interesting to see how the user facing stats will evolve ✨

PR #82, opened 2025-05-07
chore: Enable ruff type annotation checking rules

### LINKED ISSUE #81: [TASK] Enable ruff type annotation rules
## 🚧 What needs to be done

After discussion with the team, we decided we'd like to enable the following type annotation related rules in our ruff config:

- [ANN001](https://docs.astral.sh/ruff/rules/missing-type-function-argument/)
- [ANN002](https://docs.astral.sh/ruff/rules/missing-type-args/)
- [ANN201](https://docs.astral.sh/ruff/rules/missing-return-type-undocumented-public-function/)

## 🤔 Why it needs to be done

The goal of this work is to better enable us to write code that consistency includes type annotations. Currently, there is nothing in place to enforce this.

## 🏗️ Acceptance Criteria

- [ ] Rules mentioned above have been enabled in the ruff config
- [ ] Test files are excluded when these rules are checked
- [ ] All ruff errors (post making this change) have been addressed and 0 remain

## ⚙️ Technical Notes

N/A

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #81

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,9 @@ target-version = "py313"
 
 [tool.ruff.lint]
 select = [
+  "ANN001", # flake8 type annotations
+  "ANN002",
+  "ANN201",
   "F",    # Pyflakes
   "E4",   # Pydocstyle errors
   "E7",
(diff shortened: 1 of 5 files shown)


robertmitchellv (approved):
great to have these 🤖 📈 thanks!

PR #84, opened 2025-05-08
style: add Merriweather font

### LINKED ISSUE #77: [TASK] Add Merriweather font
## 🚧 What needs to be done

The [Figma designs](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=221-178&m=dev) make use of the [Merriweather font](https://fonts.google.com/specimen/Merriweather) in addition to the Public Sans font, which was added to the project in a previous ticket.

In this ticket, we should pull in Merriweather and make use of it where the designs say to do so.

## 🤔 Why it needs to be done

We want to ensure the client matches what is described in the Figma designs.

## 🏗️ Acceptance Criteria

- [ ] Merriweather has been pulled in and can be rendered in the client
- [ ] Any text that should use Merriweather has been updated to match the designs

## ⚙️ Technical Notes

N/A

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The primary goal of this PR is to add the Merriweather font and apply it to the `eCR Refiner` and `Focus on what matters.` elements on the home page so that it much more closely resembles the designs.

Additionally, I've updated the way we're managing fonts to now use [https://fontsource.org/](https://fontsource.org/) instead of downloading them directly from Google Fonts and configuring them manually in the project. This allows us to manage our fonts the same way we do our other client-side dependencies.

## 🔗 Related Issue

Fixes #77

## ✅ Acceptance Criteria

- [x] Merriweather has been added and used for expected elements

## 🧪 How to test

1. Pull down this branch and ensure client dependencies have been installed
2. Take a look at [http://localhost:8081/](http://localhost:8081/) and see that the two expected elements are being rendered as Merriweather

## ℹ️ Additional Information

N/A


client/tsconfig.app.json
```diff
@@ -20,7 +20,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
-    "noUncheckedSideEffectImports": true,
+    "noUncheckedSideEffectImports": false,
```
jakewheeler, line 23: I was getting an error from these two lines:
```ts
import '@fontsource-variable/merriweather';
import '@fontsource-variable/public-sans';
```
TS considers this a side-effect since we are only importing "global" config and not actually making use of anything specific from these packages. `"noUncheckedSideEffectImports": true` was a Vite default that I don't usually have enabled in TS projects. I don't expect we'll have a ton of this in the project in general, so I'm comfortable leaving this setting off.

robertmitchellv (approved):
thanks for adding this!

PR #88, opened 2025-05-09
fix: Add client-side routing fallback

### LINKED ISSUE #87: [TASK] Fix client-side routing
## 🚧 What needs to be done

During work on #77 it was noticed that client-side routing, when served by the FastAPI server, had not been fully implemented. The initial implementation does not appropriately handle client-side routing, such as navigating directly to the `/demo` page. We need to fix this so the production build works as expected.

We will need to create a piece of middleware that ensures the server will fall back to allowing the client to handle its own routing.

## 🤔 Why it needs to be done

- Attempting to navigate directly to client-side only routes (such as to `/demo`) doesn't work as expected. Right now the server tries to handle this instead of allowing our SPA app to handle it

## 🏗️ Acceptance Criteria

- [ ] Middleware added to allow for SPA routing fallbacks
- [ ] Navigating directly to client-side pages (`/demo`) works as expected
- [ ] Navigating directly to non-existent client pages (`/missing`) returns the basic "not found" page

## ⚙️ Technical Notes

These changes should not have any impact on the Docker compose environment used for development.

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes an issue where the server was attempting to handle routing that should be handled by React Router on the client. I have added a small piece of middleware that will now return `index.html` on routes that are not intended for the server and will instead allow the client to handle them.

Note that this issue would only happen in a production build since the server serves the static files in that scenario. The development setup should not be impacted.

## 🔗 Related Issue

Fixes #87 

## ✅ Acceptance Criteria

- [x] Production Docker image functions as expected
- [x] Routing directly to client routes, such as `/demo`, works in the browser
- [x] API routes continue to work as they did previously
- [x] All styling is still being applied in the browser

### Dev setup

1. Navigate into the `client` directory and run `npm run build` to create the `dist` directory with the built files
2. Move `dist` into `refiner` so that it sits next to `app`
3. In your browser, navigate to [http://localhost:8080/](http://localhost:8080/) and you should see that the server is serving the site
4. Try navigating to [http://localhost:8080/demo](http://localhost:8080/demo) directly - this was previously broken

### Production Docker build

1. At the top level, run a docker build - `docker build -t prod . --no-cache`
2. Start up a container using the `prod` image - `docker run -p 8080:8080 prod`
3. Steps 3 and 4 from above should work

## ℹ️ Additional Information

N/A


refiner/app/main.py
```diff
@@ -251,13 +252,9 @@ def _get_clinical_services(condition_codes: str) -> list[dict]:
 
 
 app.include_router(router)
-
-# When running the application in production we will mount the static client files from the
-# "dist" directory. This directory will typically not exist during development since the client
-# runs separately in its own Docker container.
-if is_production:
```
jakewheeler, line 258: On second thought, I don't know how useful it is to only do this on `is_production`. In a production environment, this is the only thing that will happen. In a dev environment, what does it hurt if this is enabled?

We will typically be using `localhost:8081` in dev to work with the client and get all of our nice dev features (like fast refresh) but I find it useful to be able to run a client build, move the `dist` into `refiner`, and navigate to `localhost:8080` (the server) and check if it's serving the built client files correctly. Even if we aren't using it regularly it feels helpful to be able to quickly check if a "prod" setup works without building a Docker image.

refiner/app/middleware.py
```diff
@@ -0,0 +1,35 @@
+from pathlib import Path
+
+from fastapi import Request
+from fastapi.responses import FileResponse, Response
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+
+
+class SPAFallbackMiddleware(BaseHTTPMiddleware):
```
jakewheeler, line 8: We can put this elsewhere if it makes more sense in the new structure we're moving towards. I figured putting this in its own new file here could make merging it in pretty painless.

refiner/app/main.py
robertmitchellv: i 💯 percent agree with this. i would prefer to be able to do it this way; dev on `8081` and prod on `8080` would work out pretty nicely. additionally, it will make setting up a build action workflow for the container image to host on ghcr something we can reach for more quickly

refiner/app/main.py
```diff
@@ -280,13 +281,9 @@ def _get_clinical_services(condition_codes: str) -> list[dict]:
...
-    app.mount(
-        "/",
-        StaticFiles(directory="dist", html=True, check_dir=is_production),
-        name="dist",
-    )
+app.mount(
+    "/dist",
```
robertmitchellv, line 285: oh so this is how we're able to mount what we build in client to the app; that's neat

refiner/app/middleware.py
robertmitchellv: yeah i think it sitting next to `main` is a good call. would be good to chat about where you see this sitting in the structure tho.

robertmitchellv: to be clear i don't think it should sit next to `main` in the _new_ structure, but right now since we know that change will be merged soon it's good that it's in its own file next to `main`

robertmitchellv (approved):
tested everything out; dev worked exactly as expected and "prod" with the built container worked exactly as expected. great job 🚀 

and how many `monmothma(38).zip`s do you have now 🤣

refiner/app/main.py
jakewheeler: Yep! The `Dockerfile` will generate the built client and place it in the appropriate location.

PR #94, opened 2025-05-15
chore: build Docker image and push to ghcr

### LINKED ISSUE #26: [TASK] Create GitHub Action to build and push Docker image
## 🚧 What needs to be done

We will need the ability to build the Refiner's Docker image and push it to a container registry.

In this ticket, we should create a GitHub Action to allow us to manually build a new Docker image. The GitHub Action should allow for the following arguments:

- `branch` - branch we want to build from (string, default to `main`)
- `version` - the application version to be used as the Docker image tag (string, no default)
- `tag latest` - whether or not we should also tag this build as `latest` (boolean, default to `false`)

Once the build completes, the image should be pushed to our container registry of choice (TBD on this, probably GHCR and/or ECR?).

NOTE: We may need to modify the arguments listed above as we determine what our build process looks like, but this should give us a good place to start.

## 🤔 Why it needs to be done

We want to be able to create new versions of the application on-demand.

## Acceptance Criteria

- [ ] GitHub Action created to build and push Docker image to container registry
- [ ] Action arguments function as expected
- [ ] Uploads of the new image are available in the container registry after the job runs

## ⚙️ Technical Notes

Is there anything we can borrow from another DIBBs team's build process?

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds two GitHub Actions:

1. `docker-build-ghcr.yml` - Upon merge to `main`, builds a container image tagged `main` and `latest`, and pushes it to the refiner's GitHub Container Registry. This job can also be triggered manually to run on any branch. You can also provide a version number if you want to do that. If you don't use a version number, the image tag will be a sanitized version of your branch's name.
2. `build.yml` - This job will run as part of `pipeline.yml` and will build, but not push, the Docker image. This is primarily just a check to make sure the code in the feature branch is buildable without issues.

## 🔗 Related Issue

Fixes #26

## ✅ Acceptance Criteria

- [ ] GitHub Action created to build and push Docker image to container registry
- [ ] Uploads of the new image are available in the container registry after the job runs
- [ ] Should build and push automatically on merge to `main`

## 🧪 How to test

1. See a successful pipeline run [here](https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/15074417789)
2. See a successful build and push [here](https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/15070907506)

## ℹ️ Additional Information

When this PR is merged we should see both `main` and `latest` in ghcr.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+name: Docker build
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Test Docker build
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: false
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/docker-build-ghcr.yml
+++ b/.github/workflows/docker-build-ghcr.yml
@@ -3,7 +3,7 @@ name: Build and push Refiner image to GHCR
 on:
   push:
     branches:
-      - "**"
+      - main
 
   workflow_dispatch:
     inputs:
--- a/.github/workflows/docker-build-ghcr.yml
+++ b/.github/workflows/docker-build-ghcr.yml
@@ -3,7 +3,7 @@ name: Build and push Refiner image to GHCR
 on:
   push:
     branches:
-      - main
+      - "**"
 
   workflow_dispatch:
     inputs:
@@ -64,7 +64,7 @@ jobs:
           fi
 
       - name: Build and push Docker image
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v6
         with:
           context: .
           push: true
--- a/.github/workflows/docker-build-ghcr.yml
+++ b/.github/workflows/docker-build-ghcr.yml
@@ -4,17 +4,16 @@ on:
   push:
     branches:
       - main
-    paths:
-      - "Dockerfile"
-      - "**/*.py"
-      - ".github/workflows/docker-ghcr.yml"
 
   workflow_dispatch:
     inputs:
       ref:
-        description: "Git ref (branch or tag) to build"
+        description: "Git ref (branch or tag) to build from"
         required: false
         default: main
+      version_tag:
+        description: "Optional Docker tag to use instead of branch name (`1.2.3`)"
+        required: false
 
 jobs:
   build-and-push:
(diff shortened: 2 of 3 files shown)


.github/workflows/docker-build-ghcr.yml
```diff
@@ -0,0 +1,71 @@
...
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
robertmitchellv, line 31: i'm glad we're using `buildx` for these builds 🏗️ 

i think we should keep the building instructions similar in our docs; for `refiner/README.md` [here](https://github.com/CDCgov/dibbs-ecr-refiner/blob/f094ab15c2bd0e752503151c75f56df002a7d9aa/refiner/README.md?plain=1#L42) we should do something like:

```bash
docker buildx build -t dibbs-ecr-refiner .
```

and if that doesn't work; making sure no snapshots are used and everything is built from scratch (sure there's a better way to phrase this):

```bash
docker buildx build --no-cache -t dibbs-ecr-refiner .
```

jakewheeler: My understanding is that `docker build` [is an alias for `docker buildx build`](https://docs.docker.com/reference/cli/docker/buildx/build/). I'm not against changing it at all -  just wasn't sure if this aliasing is a newer thing and it might not have been this way in the past and it made sense to be explicit about using `buildx`.

.github/workflows/docker-build-ghcr.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Log into GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Make repo owner lowercase
```
jakewheeler, line 40: `github.repository_owner` is `CDCgov` and it's expected that we provide `cdcgov` which is why I added this step. The job will break if its not lowercase.

.github/workflows/build.yml
```diff
@@ -0,0 +1,23 @@
+name: Docker build
```
jakewheeler, line 1: [Stole this from viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/.github/workflows/container-validation.yaml#L97) since it seems to be working well over there 🙂

.github/workflows/docker-build-ghcr.yml
robertmitchellv: oh `buildx` used to be separate from the legacy `build` so this is just me not knowing it was now aliased 😆

.github/workflows/docker-build-ghcr.yml
```diff
@@ -0,0 +1,71 @@
...
+        id: repo
+        run: |
+          echo "owner=$(echo '${{ github.repository_owner }}' | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
+
+      - name: Determine Docker tags
+        id: vars
+        run: |
```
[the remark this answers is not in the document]
robertmitchellv: i think the sanitize step is a good one; this seems like a good approach to me

robertmitchellv (approved):
thanks for putting this together Jake 🥇

.github/workflows/docker-build-ghcr.yml
jakewheeler: cool, yeah I figured that could be the case!

PR #98, opened 2025-05-20
feat: Install and configure react-uswds

### LINKED ISSUE #97: [TASK] Install and configure react-uswds
## 🚧 What needs to be done

With the release of [react-uswds version 10.0.1](https://github.com/trussworks/react-uswds/releases/tag/10.0.1) we are now able to install and configure the design system in the Refiner's client project.

In this ticket we should install the design system, configure it appropriately to work with the project, and convert any custom components to react-uswds components (where it makes sense to do so).

## 🤔 Why it needs to be done

- Align with typical DIBBs styling expectations
- Use pre-existing components instead of making our own
- Limit dev time required for client work

## 🏗️ Acceptance Criteria

- [ ] react-uswds installed and configured
- [ ] App looks the same as it did previously
- [ ] Custom components have been swapped out in favor of react-uswds components, where applicable

## ⚙️ Technical Notes

There will be conflicts with our existing styling that will need to be addressed as a part of this work. For example, `border-1` will now use the react-uswds class, which is undesirable.

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #97 

Fixes #
- #

### DIFF
--- a/client/src/pages/Demo/Layout.tsx
+++ b/client/src/pages/Demo/Layout.tsx
@@ -8,7 +8,7 @@ interface ContainerProps {
 
 export function Container({ color, children, className }: ContainerProps) {
   const defaultStyles =
-    'items-center gap-6 rounded-lg border-1 border-dashed px-20 py-8';
+    'items-center gap-6 rounded-lg border-thin border-dashed px-20 py-8';
   return (
     <div
       className={classNames(defaultStyles, className, {
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,6 +11,7 @@
         "@fontsource-variable/merriweather": "^5.2.2",
         "@fontsource-variable/public-sans": "^5.2.5",
         "@tailwindcss/vite": "^4.1.4",
+        "@trussworks/react-uswds": "^10.0.1",
         "classnames": "^2.5.1",
         "react": "^19.0.0",
         "react-dom": "^19.0.0",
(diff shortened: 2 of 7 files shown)


client/src/pages/Demo/Layout.tsx
```diff
@@ -8,7 +8,7 @@ interface ContainerProps {
 
 export function Container({ color, children, className }: ContainerProps) {
   const defaultStyles =
-    'items-center gap-6 rounded-lg border-1 border-dashed px-20 py-8';
+    'items-center gap-6 rounded-lg border-thin border-dashed px-20 py-8';
```
jakewheeler, line 11: USWDS also has a `border-1` utility which was overriding Tailwind's and was not at all what we want 🫠

client/vite.config.ts
```diff
@@ -11,6 +16,17 @@ export default defineConfig({
     environment: 'jsdom',
     setupFiles: 'tests/setup.ts',
   },
+  css: {
+    preprocessorOptions: {
+      scss: {
+        quietDeps: true, // tons of deprecation warnings caused by uswds
```
jakewheeler, line 22: Not a fan of this at all but the client's output is quite literally unusable if we leave deprecation warnings enabled. It's worth removing this and taking a look at the container logs to get a sense of what I mean. I am open to an alternative to avoid doing this if there are any ideas.

client/vite.config.ts
```diff
@@ -11,6 +16,17 @@ export default defineConfig({
...
+        loadPaths: uswdsIncludePaths,
+      },
+    },
+    modules: {
+      localsConvention: 'camelCaseOnly',
```
jakewheeler, line 27: This allows us to use hyphens in our class names and camel case in our TS code. 

For example, if we had `button.module.scss` with this styling:

```scss
.black-button {
    background-color: black;
    color: white;
}
```

in our React code, we can use it like this:
```ts
import styles from './button.module.scss';
...
...

<button className={styles.blackButton}>My button</button>
```

robertmitchellv: is this worth putting in our eng doc in slack?

robertmitchellv (approved):
thanks for adding the uswds library now that it's been updated. i tested everything out locally and it all looks similar except for that button color but this is a great start 🙏🏻

PR #105, opened 2025-05-21
refactor: New SQL backend for refiner

### LINKED ISSUE #70: [TASK] Switch out old SQL backend for new data model
## 🚧 What needs to be done

We need to allow for the refiner to be able to query and unpack data from the new SQL backend's data model:

```sql
-- table for storing groupers with associated codes
CREATE TABLE IF NOT EXISTS groupers (
    condition TEXT PRIMARY KEY,           -- SNOMED code for the condition
    display_name TEXT,                    -- the display name of the condition
    loinc_codes TEXT DEFAULT '[]',        -- JSON array of LOINC codes
    snomed_codes TEXT DEFAULT '[]',       -- JSON array of SNOMED codes
    icd10_codes TEXT DEFAULT '[]',        -- JSON array of ICD-10 codes
    rxnorm_codes TEXT DEFAULT '[]'        -- JSON array of RxNorm codes
);

-- table for storing user-defined filters
CREATE TABLE IF NOT EXISTS filters (
    condition TEXT PRIMARY KEY,           -- Links to condition_code in groupers
    display_name TEXT,                    -- the display name of the condition
    ud_loinc_codes TEXT DEFAULT '[]',     -- JSON array of user-defined LOINC codes
    ud_snomed_codes TEXT DEFAULT '[]',    -- JSON array of user-defined SNOMED codes
    ud_icd10_codes TEXT DEFAULT '[]',     -- JSON array of user-defined ICD-10 codes
    ud_rxnorm_codes TEXT DEFAULT '[]',    -- JSON array of user-defined RxNorm codes
    included_groupers TEXT DEFAULT '[]'   -- JSON array of grouper condition codes
);
```

## 🤔 Why it needs to be done

We need to be able to persist user defined filters and maintain updates to the reportable condition groupers from the TES. The data model that we've decided to move forward with will allow for both of these things with a simple structure.

This will give us the ability to access the main RCTC trigger codes' SNOMED concepts, the human readable name as well as all associated codes and potentially custom user defined codes (as well as including other groupers if desired).

To make this possible we will need to modify existing code to both query and unpack this new data model.

## 🏗️ Acceptance Criteria

- [x] add the new schema
- [x] add a `seed_db.py` script
- [x] add a script to check the database's data (this will be a scripted set of queries; could even be a unit test for the db)

**Potentially later tickets:**
- [ ] if possible, think about migrations
- [ ] if possible, think about updating/seeding with a route

## ⚙️ Technical Notes

- most of this has been figured out and will be aligned into the main `refine.py` function.

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This is an update to the data model for the SQLite3 database that we seed from the APHL Terminology Exchange Service (TES).

There is currently a new script for seeding the database and for checking its outputs with some sample queries so you can see the shape of the data. There is still more work to do on how we will fold in the `filters` table so for now we're just defaulting to single condition `groupers` data.

All database related modules are now folded under `app/db/` and the terminology related services are still in `app/services/terminology.py`. All of the business logic in `app/services/refine.py` has been updated but there is still work to do in connection to updating the refine workflow.

## 🔗 Related Issue

Fixes #70 
- #70 

## 🧪 How to test

1. try running tests locally with `pytest`
2. try spinning up the containers with `docker compose up -d` and then running the demo `.http` file in `docs/refiner/`
3. navigate to the demo page to make sure everything is working as expected

## ℹ️ Additional Information

- [x] add the new schema
- [x] add a seed_db.py script
- [x] add a script to check the database's data (this will be a scripted set of queries; could even be a unit test for the db)

## Appendix

> [!NOTE] 
> `lxml` XML element types in this codebase
>
> When working with XML in Python using `lxml`, you will see both `_Element` and `etree.Element` in our code.

- **Type hints:** We use `_Element` (from `lxml.etree`) for type annotations, because it is the actual type of XML element objects. This is important for static analysis and tools.
- **Element construction:** We use `etree.Element(...)` in the code to create elements. This is a factory function that returns an `_Element` instance.
 
**Why?**
 
- `etree.Element` is *not* a class, but a function. Using it as a type hint is incorrect and may confuse type checkers.
- `_Element` is the correct type for variables that hold XML elements.
 
See also: [lxml _Element API docs](https://lxml.de/api/lxml.etree._Element-class.html)

This pattern will change in `lxml` 6.0; check the [PR](https://github.com/lxml/lxml/pull/405) for more information.


.github/workflows/tests.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -U pip
-          pip install -r dev-requirements.txt -r requirements.txt
+          pip install -r requirements-dev.txt -r requirements.txt
```
jakewheeler, line 38: Personal preference but I'd love for this type of change to be its own PR in the future unless it needed to be done here. No action needed and not a big deal at all - I just figured I'd mention it since I'm looking at it!

refiner/app/api/v1/ecr.py
```diff
@@ -88,14 +87,8 @@ async def refine_ecr_from_zip(
         if sections_to_include:
             sections = refine.validate_sections_to_include(sections_to_include)
 
-        # process conditions if provided
-        clinical_services = None
-        if conditions_to_include:
-            clinical_services = list(get_clinical_services(conditions_to_include))
-            clinical_services = create_clinical_services_dict(clinical_services)
```
jakewheeler, line 95: Glad to see this go away 🎉

refiner/scripts/README.md
```diff
@@ -9,8 +9,12 @@ Before you can run the seed script, you'll need to acquire a TES API key. Make a
 ## Running the script
 
 1. Get set up to run the Refiner locally ([see here](../README.md#running-from-python-source-code))
-2. Ensure dependencies have been installed (`pip install -r requirements.txt -r dev-requirements.txt`)
+2. Ensure dependencies have been installed (`pip install -r requirements.txt -r requirements-dev.txt`)
 3. Update (or create) your `scripts/.env` file to include your TES API key. Ex: `TES_API_KEY=xxxx....`
```
jakewheeler, line 13: Do we want to outline all required and optional environment variables here?

- `TES_API_URL` (required)
- `TES_API_KEY` (required)
- `API_SLEEP_INTERVAL` (optional) - defaults to `1.0`

Could we provide a default value for `TES_API_URL`? Any reason not to?

.github/workflows/tests.yml
robertmitchellv: yeah, i think that's fair. i just noticed after looking at a lot of other major python projects that it's always `requirements-dev.txt` and not the other way around so it was bugging me 😆

refiner/scripts/README.md
robertmitchellv: i suppose we could include the URL but when i was updating this i was also asking myself:

> in what way do we either expect or encourage users to deploy this if they are not approaching this project with code contributions in mind?

i could create a "sample env" file and a shell script that will create a `.env` if one doesn't already exist. i can borrow the script i made for the viewer here i suppose but i'm still sort of wondering what we want the documentation to communicate and for what reason. i think it provides a bit too much of not enough if that makes sense?

refiner/scripts/seed_terminology_db.py
```diff
@@ -0,0 +1,345 @@
+import json
```
jakewheeler, line 1: I needed to make sure I had `TES_API_URL` in my `.env` but otherwise this worked perfectly in my testing.

refiner/scripts/README.md
jakewheeler: I don't think you need to take it that far currently. I think as long as I or @lina-roth know how we need our `.env` file configured, that's good enough for now. My thought on this is certainly subject to change down the road, depending on how the project progresses and who we think might be running this script that isn't one of us on the core team.

refiner/app/README.md
```diff
@@ -60,3 +60,22 @@ Interactive API documentation is available at:
 The application uses JSON assets for configuration:
 - `refiner_details.json`: Section processing rules
 - Sample request/response examples for OpenAPI documentation
+
+## Appendix
+
+> [!NOTE]
```
robertmitchellv, line 66: i added info about how we're handling type hints with `lxml` and what we do inside the functions themselves to the readmes at:

- `/app/README.md`
- `/app/services/README.md`

and also as a comment in:

- `/app/services/refine.py`

refiner/app/services/refine.py
```diff
@@ -369,95 +337,155 @@ def _get_observations(
...
     # loop less, search less, and aim for simplicity
 
-    return observations
+    try:
+        # process each condition code
+        for code in condition_codes.split(","):
+            try:
+                grouper_row = grouper_ops.get_grouper_by_condition(code)
+                if grouper_row:
+                    processed = ProcessedGrouper.from_grouper_row(grouper_row)
```
jakewheeler, line 357: huh, this is interesting. Is this required because it's essentially acting as the constructor for the `dataclass`? Anything else worth knowing here about how this works?

refiner/app/services/refine.py
```diff
@@ -471,30 +499,40 @@ def _extract_observation_data(
...
+        "display_text": display_text,
+        "code": code,
+        "code_system": code_system,
         "is_trigger_code": is_trigger_code,
     }
     return data
 
 
-def _create_or_update_text_element(observations: list[etree.Element]) -> etree.Element:
+def _create_or_update_text_element(observations: list[_Element]) -> _Element:
```
jakewheeler, line 538: Thank you for fixing these 🙏

refiner/scripts/seed_terminology_db.py
jakewheeler: Side note but this makes the tables so much simpler 👍

refiner/app/db/schema.sql
```diff
@@ -0,0 +1,20 @@
+-- table for storing groupers with associated codes
```
jakewheeler, line 1: Is there anything we need to consider right now in terms of future DB migrations? Do you think that sounds like a later problem once we have things more nailed down?

refiner/app/services/refine.py
robertmitchellv: yeah the `from_grouper_row` method on `ProcessedGrouper` dataclass acts as a custom constructor because we have to transforms the database row (with json strings for various codeable concepts) into an object with a combined set of codes (in a `Set`). there isn't a 1:1 between the search results, the db row input, and the `ProcessedGrouper` dataclass that has all codes from all systems in a `Set` (without display name).

effectively we have the "raw" data as the `TypedDict` to represent what we're getting from the db and then the dataclass that processes that raw data into the frozen object that makes it easier to search and create xpaths.

refiner/app/db/schema.sql
robertmitchellv: we definitely will need to figure out migrations but at this point i don't exactly know how the `filters` table will leverage what is in `groupers`. i feel like `groupers` is a pretty solid table and we have an effective way to seed it but there's still more thinking that needs to be done on an ERD that satisfies some of the new designs and the relationship between the filters and conditions.

refiner/scripts/seed_terminology_db.py
```diff
@@ -0,0 +1,345 @@
...
+                continue
+
+            codes[code_type].extend(
+                [
+                    CodeableConcept(
+                        code=concept["code"], display=concept.get("display", "")
+                    )
+                    for concept in concepts
+                ]
+            )
```
jakewheeler, line 283: How against would you be swapping this out for this?
```suggestion
for concept in concepts:
    codeable = CodeableConcept(
        code=concept["code"],
        display=concept.get("display", "")
    )
    codes[code_type].append(codeable)
```

Maybe less pythonic but it took me a minute to understand the original 😅 I think the formatting of it was also throwing me off for some reason.

robertmitchellv: yeah, i actually like this better. do you want to try and delete your local db and run the seed script and the other check_db script to make sure the outputs line up? couldn't hurt to just do `pytest tests/` as well

jakewheeler (approved):
Excellent work on this! So glad to finally get this merged in 🥳

PR #111, opened 2025-05-23
fix: properly escape JSON payload in workflow dispatch event

alismx:
## Changes Proposed

- This change allows variables to be correctly interpreted.

## Testing

- Review logs to confirm that the dispatch event returns a success status code (204).


jakewheeler (approved):
Thanks Alis!

PR #113, opened 2025-05-29
feat: add dropdown to demo success page to select eICR to view

jakewheeler:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #
- #


refiner/app/api/v1/demo.py
```diff
@@ -189,19 +189,28 @@ async def demo_upload(
...
                             )
                         }%",
                     ],
+                },
+            )
+
+        return JSONResponse(
+            content=jsonable_encoder(
+                {
+                    "conditions": conditions,
```
jakewheeler, line 213: There will only be a single condition in the array until we swap out the sample data for the newer data @robertmitchellv is working on.

PR #115, opened 2025-05-30
chore: update "how to test" with better defaults

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Updating this PR template so that the commented out **How to test** section contains two common examples of how we might want to test the application. This will make it easier to continue to have better instructions without having to type them over and over. That way _how to test_ doesn't end up looking like _run docker, check application_

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -49,7 +49,7 @@ Unit tests
 
 or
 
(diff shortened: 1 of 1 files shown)


jakewheeler (approved):
Thanks for making this change, Robert!

PR #133, opened 2025-06-09
test: Run mypy as part of the GitHub Actions pipeline

### LINKED ISSUE #121: [TASK] Run mypy as part of the GitHub Actions pipeline
## 🚧 What needs to be done

The engineering team would like to add [mypy](https://mypy-lang.org/) to the project and run it as a GitHub Action as part of the `pipeline.yml` Action so that it happens on every PR.

## 🤔 Why it needs to be done

This will:
1. Give developers more confidence that their changes are correct and are not breaking existing APIs
2. Lessen the burden on PR reviewers since they will not need to look for types that seem out of place

## 🏗️ Acceptance Criteria

- [ ] Top-level `pyproject.toml` has been updated to include a mypy config (see technical notes)
- [ ] `requirements-dev.txt` includes `mypy==1.15.0` (or latest version)
- [ ] mypy only runs against Python files in the `refiner/app/` directory
- [ ] All existing errors pointed out by mypy have been addressed
- [ ] GitHub Action has been created and runs as a part of `pipeline.yml` so that it runs on every PR

## ⚙️ Technical Notes

This config should do what we want when added to the top-level `pyproject.toml`:
```toml
[tool.mypy]
files = ["app/**/*.py"]
```

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR:
1. Adds and configures `mypy` to run in the `app` directory within the `refiner`
2. Fixes all current typing issues within `app`
3. Adds a GitHub Action to run as a part of the pipeline on every PR 

## 🔗 Related Issue

Fixes #121 

## 🧪 How to test

1. Reinstall the Python dependencies (or restart your Docker Compose setup, which should handle this)
2. From within `refiner`, try running `mypy`. You should see `Success: no issues found in 24 source files` (all *.py files in `app`)

Other things to check:
1. All automated tests should pass
2. App continues to work in the browser as it did previously
3. GitHub Action reports failures (see the [sample failure here](https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/15537686314/job/43740628194) where I introduced a few typing issues as a test)

## ℹ️ Additional Information

I had to `cast` in a few places in `refine.py` because of `lxml`'s typing. Is there a better way to handle these scenarios?

### DIFF
--- a/refiner/app/services/refine.py
+++ b/refiner/app/services/refine.py
@@ -590,7 +590,10 @@ def _create_or_update_text_element(observations: list[_Element]) -> _Element:
         row = etree.SubElement(table_element, "tr")
         for key in headers[:-1]:  # Exclude the last header as it's for the boolean flag
             td = etree.SubElement(row, "td")
-            td.text = cast(str | None, data[key.lower().replace(" ", "_")])
+            value = data[key.lower().replace(" ", "_")]
+            if value is not None and not isinstance(value, str):
+                value = str(value)
+            td.text = value
 
         # add boolean flag for matching condition code (always TRUE since we only keep matches)
         td = etree.SubElement(row, "td")


refiner/app/main.py
```diff
@@ -54,7 +54,7 @@ async def _lifespan(_: FastAPI):
 # set service_path in app state
 # add open api configuration
 app.state.service_path = "/api"
-app.openapi = lambda: create_custom_openapi(app)
+app.openapi = lambda: create_custom_openapi(app)  # type: ignore
```
jakewheeler, line 57: Not 100% sure what the issue is here exactly. I went ahead and ignored this since this isn't something we'll be changing often and is not part of the core application. What we have seems in-line with the FastAPI docs so I'm a bit confused.

I would be happy to fix this if anyone has an idea about how to type this correctly.

refiner/app/services/refine.py
```diff
@@ -124,10 +124,13 @@ def get_reportable_conditions(root: _Element) -> list[dict[str, str]] | None:
...
-            ".//cda:section[cda:code/@code='55112-7']"
-            "//cda:entry/cda:organizer[cda:code/@code='RR11']",
-            namespaces=namespaces,
+        coded_info_organizers = cast(
+            list[_Element],
+            root.xpath(
+                ".//cda:section[cda:code/@code='55112-7']"
+                "//cda:entry/cda:organizer[cda:code/@code='RR11']",
+                namespaces=namespaces,
+            ),
```
jakewheeler, line 133: Without casting we get a `mypy` error like this:
```sh
app/services/refine.py:153: error: Item "bool" of "bool | float | str | bytes | list[_Element | _ElementUnicodeResult | _PyElementUnicodeResult | _ElementStringResult | str | bytes | tuple[str | bytes | None, str | bytes | None]]" has no attribute "xpath"
```
There are a LOT of possibilities for what this type could be. I went through each of the `.xpath()` lines and determined that `list[_Element]` appears to be what we are typically expecting.

Let me know if you think this is wrong and/or if you have a different way in mind about how we could handle these.

refiner/app/main.py
robertmitchellv: i was curious about this and it seems like what some people do is create a wrapper function with return types like this:

```python
def custom_openapi() -> dict[str, object]:
    return create_custom_openapi(app)

app.openapi = custom_openapi
```

there's also a class based way, which i didn't like as much and some others that seemed more complicated.

from what i can see `.openapi` is a method that doesn't take any arguments.

also, i think that lambdas are likely a bad fit if we want to do more strict type checking. python is fine with it but mypy doesn't like it since there isn't a way to hint at the type (at least that i'm aware of). 

i also think that the type hint for the custom function in `app/core/app/openapi` that we import is "partially unknown"; it should probably match this wrapper i think, e.g., `-> dict{str, object]`

refiner/app/services/file_io.py
```diff
@@ -128,6 +129,19 @@ async def read_xml_zip(file: FileUpload) -> XMLFiles:
...
+                    details={
+                        "files_found": [
+                            f
+                            for f in z.namelist()
+                            if not (f.startswith("__MACOSX/") or f.startswith("._"))
+                        ],
+                        "required_files": ["CDA_eICR.xml", "CDA_RR.xml"],
+                    },
+                )
+
```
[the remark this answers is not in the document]
robertmitchellv: oh good catch; this seems right to me

refiner/app/services/refine.py
```diff
@@ -576,7 +590,7 @@ def _create_or_update_text_element(observations: list[_Element]) -> _Element:
         row = etree.SubElement(table_element, "tr")
         for key in headers[:-1]:  # Exclude the last header as it's for the boolean flag
             td = etree.SubElement(row, "td")
-            td.text = data[key.lower().replace(" ", "_")]
+            td.text = cast(str | None, data[key.lower().replace(" ", "_")])
```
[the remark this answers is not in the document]
robertmitchellv: yeah, cast doesn't seem right here. i like the idea of checking first. i can't imagine how anything inside of `td.text` is anything other than a string or `None`. the xml docs have already been read and at the point of this function being called have already been split by section.

were you thinking something like this?

```python
value = data[key.lower().replace(" ", "_")]
if value is not None and not isinstance(value, str):
    value = str(value)
td.text = value
```

refiner/app/main.py
jakewheeler: mypy still seems to not like this because we're monkeypatching the `openapi` method (which as far as I can tell from the docs, this is what FastAPI wants and expects us to do).

When I try to go with your suggestion this is the error I receive:
```sh
app/main.py:63: error: Cannot assign to a method  [method-assign]
```

Here is another alternative I've seen suggested, which mypy seems to have no issue with, but also feels hacky 🫠 
```python
setattr(app, "openapi", lambda: create_custom_openapi(app))
```

refiner/app/services/refine.py
jakewheeler: Awesome, thanks for double checking me!

refiner/app/main.py
robertmitchellv: that's kind of bewildering tbh. i think i'm good to just tell mypy to pipe down then 🤣

robertmitchellv (approved):
thanks for adding this! i'm going to need to see if i can get my LSP using our mypy rules now 🏆

PR #138, opened 2025-06-11
feat: eCR file upload

### LINKED ISSUE #49: [USER STORY] Test eCR File Upload
## 👤💭 User Story

As an informatician/eCR Lead, I want to upload my own eCR test file, so that I can make sure the Refiner will refine it how I want.

## 🏗️ Acceptance Criteria

- [ ] Allows user to upload a single eICR/RR as a .zip file
- [ ] Does checks to make sure a single, valid eICR/RR pair is uploaded as a .zip **- What do we need to check for?**

## 🎨 Designs
(ignore header navigation, pizza tracker, step x of x)

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a feature to allow demo users to upload their own .zip file containing an `CDA_eICR.xml` and `CDA_RR.xml` pair for processing through the Refiner.

## 🔗 Related Issue

Fixes #49 

## ✅ Acceptance Criteria

- [x] Allows user to upload a single eICR/RR as a .zip file
- [x] Does checks to make sure a single, valid eICR/RR pair is uploaded as a .zip
  - [x] Valid file extension
  - [x] Valid file name
  - [x] Is not empty
  - [x] Max compressed size
  - [x] Max uncompressed size
  ~- [x] Zip file contains only `CDA_eICR.xml` and `CDA_RR.xml`~
  - [x] Zip file can be processed

## 🧪 How to test

Testing the application:

1. Run `docker compose up`
4. You should be able to navigate to the upload page. On the left side, you'll be able to click the button and select a file from your file system. Once complete, you can change the file or upload the file. When you upload the file it will take you to the reportable conditions screen

## Example
Kapture 2025-06-16 at 10 02 26

## ℹ️ Additional Information

Do you think the check to require a zip file to contain ONLY `CDA_eICR.xml `and `CDA_RR.xml` is too restrictive? Should I loosen it to check that one and only one of each of these files exist instead? Other thoughts?

There are some additional file checks already happening in `read_xml_zip()`.


refiner/app/api/v1/demo.py
```diff
@@ -146,8 +151,108 @@ def _get_refined_ecr_output_dir() -> Path:
...
+
+    try:
+        file_content = await file.read()
+        with ZipFile(io.BytesIO(file_content)) as zf:
+            # Zip must contain only the two expected files
+            file_count = _get_zip_file_count(zf)
+            if file_count > MAX_ALLOWED_FILE_COUNT:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail=f".zip must contain only 2 files: CDA_eICR.xml and CDA_RR.xml. Found {file_count} files.",
```
jakewheeler, line 224: Is this requirement too strict? If we do want to keep it, maybe we should list the file names found to give the user more details as to why it failed? 

What do you think?

robertmitchellv: i think this is too strict because a typical `.zip` from AIMS will have these files:

- `CDA_eICR.xml`
- `CDA_RR.xml`
- `CDA_eICR.html`
- `CDA_RR.html`
- potentially other files, like a json file or even the FHIR bundle if they've turned it on in RCKMS

i think as long as the `.zip` contains the files we want we should proceed. especially because we'd want users to be able to take any random zip from AIMS and try it out if they have one

refiner/app/api/v1/demo.py
```diff
@@ -20,6 +20,11 @@
 FILE_NAME_SUFFIX = "refined_ecr.zip"
 file_store: dict[str, dict] = {}
 
+# File uploads
+MAX_ALLOWED_UPLOAD_FILE_SIZE = 10 * 1024 * 1024  # 10 MB
+MAX_ALLOWED_UNCOMPRESSED_FILE_SIZE = MAX_ALLOWED_UPLOAD_FILE_SIZE * 5  # 50 MB
```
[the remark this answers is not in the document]
robertmitchellv: i'm good with 10mb as the limit for now

refiner/app/api/v1/demo.py
jakewheeler: Totally forgot about the HTML files 😄 Great point! I think I'm just going to remove this entire case since we already validate this as a part of `read_xml_zip()`.

robertmitchellv (approved):
thanks for making that change! it's working really well and it's illuminating some other issues to tackle next 🙏🏻

PR #171, opened 2025-07-01
refactor: Move from SQLite to PostgreSQL

### LINKED ISSUE #169: [USER STORY] Migrate from SQLite to PostgreSQL
## 👤💭 User Story

As a user, I want to be able to store data to and retrieve data from the Refiner. I expect the application to be able to seamlessly provide updates over time without losing my data.

## Summary

In this ticket, we want to migrate off of SQLite and on to PostgreSQL since we plan to make use of [AWS RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/) in a production environment. We do not currently have a live environment, so the main focus of this ticket will be modifying the Refiner's local Docker compose setup in order to add PostgreSQL.

## 🏗️ Acceptance Criteria

- [ ] Find and install SQL client to use with FastAPI
- [ ] Update database access code to use new SQL client across all packages
    - [ ] `refiner/app/db`
    - [ ] `refiner/scripts`
    - [ ] `refiner/tests`
- [ ] Add and configure `postgres:alpine` image to `docker-compose.yml` ([Example](https://github.com/skylight-hq/phinvads-go/blob/main/docker-compose.yml), [Docker blog post](https://www.docker.com/blog/how-to-use-the-postgres-docker-official-image/))
- [ ] Test that seeding works as expected
- [ ] Remove `refiner/services/terminology.db` from the project

## 📜 Additional context

Even if we do not implement this here, we should start thinking about what database migrations might look like in the future and how we will manage them.

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR removes the `terminology` SQLite database and instead makes use of PostgreSQL. The Postgres container will now run as a part of the Docker Compose setup.

### Seeding

Unfortunately with this change, we no longer have a pre-seeded database checked in that we can immediately start querying from. Instead, we now have to seed the PostgreSQL database before we begin running integration tests.

This is being handled in two steps:

1. Seeding locally will generate a `refiner/scripts/seed-data.sql` file - this file is checked in
2. The `scripts/seed-data.sql` file is used by `refiner/tests/integration/conftest.py` which runs prior to the integration tests running. A new step setup step has been added to find this file from within the `db` container and use it to populate the database with data

Same as before, we do not need to make calls to the TES from the pipeline. This setup will likely continue to evolve as we change the data model, change the way we seed, etc.

### SQLite specific tests

As I mentioned to the team a couple of times, `test_db.py` was giving me some trouble in doing a straight conversion from SQLite over to Postgres. I've removed these tests for now since we will be developing integration tests once the new data model has landed.

## 🔗 Related Issue

Fixes #169 

## ✅ Acceptance Criteria

- [x] Application works as it did previously
- [x] Tests are passing
- [x] `scripts/check_terminology_db.py` works as it did previously
- [x] `scripts/query_terminology_db.py` works as it did previously
- [x] `scripts/seed_terminology_db.py` works as it did previously + generates `scripts/seed-data.sql`

## 🧪 How to test

1. Stop your running containers
4. Update your `refiner/scripts/.env` file to include `DB_URL=postgresql://postgres:refiner@localhost:5432/refiner`
5. Run `docker compose up` to bring your containers back up, you'll see a `db` container running. `psycopg` should get installed automatically
6. Seed your new database by running `python refiner/scripts/seed_terminology_db.py`
7. You should see a new `scripts/seed-data.sql` file get created - feel free to ignore this

### Check that scripts work

Both of these scripts should work as they did previously:

1. Try running `python refiner/scripts/check_terminology_db.py`
3. Try running `python refiner/scripts/query_terminology_db.py`

### Check that the app works

1. Navigate to [http://localhost:8081/](http://localhost:8081/) which should look the same as before
3. Click on `Testing` and run through the demo workflow using the `Use test file` button to make use of the default data
4. You should see the app continue to work as it did previously

## ℹ️ Additional Information

Feel free to try logging into the dev DB with any Postgres client you want (I like [pgAdmin](https://www.pgadmin.org/)). You should see a database called `refiner` that has two tables: `filters` and `groupers` - identical to what we had in SQLite.

- Did I miss anything?
- We will want to add DB-focused integration tests once we have the new data model implemented, an updated seed script, and connection pooling.


.pre-commit-config.yaml
```diff
@@ -10,8 +10,8 @@ repos:
     # - prevent large file commits
     -   id: check-added-large-files
         args:
-        # 5mb == 5120kb
-        -   --maxkb=5120
+        # 20mb == 20480kb
+        -   --maxkb=20480
```
jakewheeler, line 14: `seed-data.sql` is huge 😭

refiner/scripts/check_terminology_db.py
```diff
@@ -1,8 +1,9 @@
 import json
```
jakewheeler, line 1: I'd appreciate some extra scrutiny on the script files since these needed the most substantial changes since they were very SQLite specific and now need to be Postgres specific. The functionality appeared to be unchanged to me during my testing but let me know if you notice anything wrong.

refiner/requirements.txt
```diff
@@ -2,6 +2,7 @@ chardet
 fastapi>=0.109.1
 httpx
 lxml
+psycopg[binary]
```
jakewheeler, line 5: We'll need to add the dependency for connection pooling later, which will come in #179.

refiner/scripts/seed_terminology_db.py
```diff
@@ -333,10 +337,40 @@ def populate_groupers_and_filters(self) -> None:
         self.logger.info("Finished populating groupers and filters tables")
 
 
+def dump_postgres_db_from_url(conn_str: str, output_file: str):
```
jakewheeler, line 340: I am noticing that `pg_dump` orders the text of `seed-data.sql` differently every time it runs, meaning that git will see changes when there really aren't any.

Should we git ignore `seed-data.sql` and put the file used to seed the integration test DB somewhere else? Updating the integration test data will have to be done explicitly, but maybe that is preferable to `seed-data.sql` constantly showing changes every time one of us re-seeds?

.gitignore
```diff
@@ -35,6 +35,7 @@ coverage.xml
 # Static files
 dist
 refined-ecr
+**/scripts/seed-data.sql
```
jakewheeler, line 38: ☝️ `refiner/scripts/seed_terminology_db.py` will still produce this file but it's no longer being checked in. The integration tests will now populate the database using `refiner/tests/integration/seeding/seed-data.sql`, which we need to explicitly update instead of it being "automatic" as it was before.

refiner/scripts/seed_terminology_db.py
jakewheeler: Here is the ticket! https://github.com/CDCgov/dibbs-ecr-refiner/issues/189

refiner/scripts/seed_terminology_db.py
```diff
@@ -333,10 +337,40 @@ def populate_groupers_and_filters(self) -> None:
...
+    """
+    # Grab user and database from connection string
+    match = re.match(r"postgresql://([^:]+):[^@]+@[^/]+/([^?]+)", conn_str)
+    if not match:
+        raise ValueError("Invalid connection string format.", conn_str)
+
+    user, dbname = match.groups()
+
+    # Compose the command
+    dump_cmd = f"docker-compose exec db pg_dump -U {user} {dbname} > {output_file}"
```
robertmitchellv, line 356: i think that we've snuck in a dependency that isn't tracked; this is probably coming from the homebrew version of `docker-compse` (could you confirm with `which docker-compose`?)

meaning that we are relying on the python library `docker-compose` and not the go addition to the docker cli `docker compose`

it didn't work on my system because i don't have this installed in my env and the seeding failed. when i changed it i was able to get things working correctly.

```suggestion
dump_cmd = f"docker compose exec db pg_dump -U {user} {dbname} > {output_file}"
```

jakewheeler: Whoops 😬 Sorry about that! You're right that `compose` is supposed to now go through the `docker` binary. I have the old binary on my system that still works which is why I didn't catch this. 

Thank you!

PR #178, opened 2025-07-02
docs: Refiner data model v0 proposal

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The goal of this PR is to give the team an opportunity to shape the initial Refiner data model. Ideally, feedback will be provided and we will be able to come to a consensus on a path forward from here.

Implementation, seed script changes, etc. will come in another ticket/PR once we all agree and merge this.

This will close out #183 upon merging.

## Note

Using https://dbdiagram.io/ was an arbitrary choice on my end. It's easy enough to construct a diagram, see changes reflected immediately, and share it which is why I went with it. We can change software at a later point if there's desire for that.


docs/refiner/data-model/README.md
```diff
@@ -0,0 +1,93 @@
...
+  loinc_codes jsonb
+  snomed_codes jsonb
+  icd10_codes jsonb
+  rxnorm_codes jsonb
+}
+
+Table tes_rs_grouper {
+  grouper_id integer
+  snomed_code text
+}
```
jakewheeler, line 59: Did we say we were going to link this to `filter`?

docs/refiner/data-model/normalized-tes-with-denormalized-cache.dbml
```diff
@@ -0,0 +1,191 @@
+////
+// DIBBS eCR Refiner - Automated Data Pipeline Architecture
```
robertmitchellv, line 2: i might've gone a bit overboard with this but after our sync on trying to just get something that is mostly postgres only first and then dial up the complexity if the performance isn't working i came up with this basic idea that i think accounts for the TES updates, the two different triggers for aggregate data (from my research we can join/aggregate jsonb) to sort of end up with that cache table for the app to use we talked about. i haven't written out the SQL and the triggers since this is just an idea but it should all be handled in the database. let me know if you think it addresses the things we need and what parts we can use for the final version of the data model.

i ended up hitting some conceptual problems when we introduce updates to the TES into both the application and the data model. questions around what updates mean for the app, for filters, for versions and how that is managed by the database and the UI. i think this model accounts for it but we can maybe chat before standup if it's confusing

PR #186, opened 2025-07-07
fix(ui): Update styles and better a11y support

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Brings in the following changeset:

1. 💬 **fix(ui):** Make sure styles match design system; 🏷️ e377be8
    - 📝 After some feedback from Kyle, I've gone and made the styles more
    accessible for moving around the application using the keyboard and
    making sure that things don't collide with people who use Touch or Mouse
    inputs. Inspiration for the focus trick is to leverage a suggestion from
    @LeaVerou mentioned here: https://x.com/LeaVerou/status/1045768279753666562
    This also updates the colors to match the Figma design which was missed in
    the earlier changeset.
1. 💬 **feat(ui):** Add support for skipping to content 🏷️ 9c2afb0

## 🔗 Related Issue

- #164

## ✅ Acceptance Criteria

- [x] Header navigation between configurations page and testing page for all pages incl. testing flow
- [ ] ~~Remove landing page~~
- [ ] ~~Remove "Return to landing page" buttons in testing flow~~
- [ ] ~~Configurations page title and subtitle copy~~
- [ ] ~~Empty table (or no table if easier to add later)~~

## 🧪 How to test

You can load this up locally and use the keyboard to navigate things. Since
it's a11y-related, it'll be hard to test automatically. For now, we can
leverage this animated GIF below.

<details open>
<summary> 📹Screen-recording(s) </summary>

a11y-for-navigation-example

</details>

## ℹ️ Additional Information

This came back as feedback from Kyle. I took it as an opportunity to better
support some of the a11y styles around focusing as well.

### DIFF
--- a/client/src/styles/index.scss
+++ b/client/src/styles/index.scss
@@ -2,7 +2,8 @@
 @forward 'uswds';
 
 /**
-  * @mixin Sets outline to a better complimentary and WCAG AA rated contrast
+  * @mixin
+  * @desc Sets outline to a better complimentary and WCAG AA rated contrast
   * raio of `5.93:1` which works for graphic and large text.
   * source: https://webaim.org/resources/contrastchecker/?fcolor=82B4C9&bcolor=14333D
   */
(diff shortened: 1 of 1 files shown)


client/src/components/Layout/index.tsx
```diff
@@ -47,6 +53,7 @@ export function Footer() {
             href="https://www.cdc.gov"
             target="_blank"
             rel="noreferrer noopener"
+            className="inline-block"
```
rogeruiz, line 56: Necessary to get the outline styles to be render for the CDC logo due to the `inline` CSS defaults for `<a />` and `` elements.

jakewheeler: Another good change! 🎉

PR #187, opened 2025-07-09
feat(ui): Add `<ConfigurationsTable />` component

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

1. 💬 chore(ui): Bring in more colors from design system 🏷️ dc28404
1. 💬 feat(ui): Add component <ConfigurationsTable /> 🏷️ 2d60cc6

## 🔗 Related Issue

- #164

## ✅ Acceptance Criteria

- [ ] ~~Header navigation between configurations page and testing page for all pages incl. testing flow~~
- [ ] ~~Remove landing page~~
- [ ] ~~Remove "Return to landing page" buttons in testing flow~~
- [ ] ~~Configurations page title and subtitle copy~~
- [x] Empty table (or no table if easier to add later)

## 🧪 How to test

There are unit tests associated with the `<ConfigurationsTable />` component
and the `<StatusPill />` component. You can also load up the app locally and
see the Configurations page has a table component now that matches the designs.

## ℹ️ Additional Information

The table is using sample data from the original designs. This means that when
writing the retrieval from the database, we may need to refactor parts of this
component and add data-fetching. As part of this ticket, I want to add keyboard
navigation to the Table component to be able to meet accessibility needs when
JavaScript is enabled. I have a rough sketch, but am worried that it might
extend this work longer and should probably just be done in a separate ticket.

### DIFF
--- a/client/src/components/ConfigurationsTable/index.tsx
+++ b/client/src/components/ConfigurationsTable/index.tsx
@@ -1,0 +1,61 @@
+import { Table as UswdsTable } from '@trussworks/react-uswds';
+import { StatusPill } from '../StatusPill';
+
+interface HeaderProps {
+  items: string[];
+}
+
+interface TableBodyProps {
+  data: Array<TableItem>;
+}
+
+interface TableItem {
+  name: string;
+  status: string;
+  id: string;
+}
+
+interface ConfigurationsTableProps {
+  columns: string[];
+  data: Array<TableItem>;
+}
+
+function TableHeader({ items }: HeaderProps) {
+  return (
+    <thead>
+      <tr>
+        {items.map((name, idx) => {
+          return <th key={idx} scope="col">{name}</th>;
+        })}
+      </tr>
+    </thead>
+  );
+}
+
+function TableBody({ data }: TableBodyProps) {
+  return (
+    <tbody>
+      {data.map((row, idx) => {
+        return (
+          <tr key={idx}>
+            <th scope="row">{row.name}</th>
+            <td>
+              <StatusPill status={row.status} />
+            </td>
+          </tr>
+        );
+      })}
+    </tbody>
+  );
+}
+export function ConfigurationsTable({
+  columns,
+  data,
+}: ConfigurationsTableProps) {
+  return (
+    <UswdsTable stackedStyle="headers">
+      <TableHeader items={columns} />
+      <TableBody data={data} />
+    </UswdsTable>
+  );
+}
--- a/client/src/components/StatusPill/index.test.tsx
+++ b/client/src/components/StatusPill/index.test.tsx
@@ -1,0 +1,26 @@
+import { describe, it } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { BrowserRouter } from 'react-router';
+import { StatusPill } from '.';
+
+const renderComponentView = (status: string) =>
+  render(
+    <BrowserRouter>
+      <StatusPill status={status} />
+    </BrowserRouter>
+  );
+
+describe('Status pill component', () => {
+  it('should render with status on', async () => {
+    renderComponentView('on');
+
+    expect(screen.getByTestId('status-pill')).toBeInTheDocument();
+expect(screen.getByTestId('status-pill')).toHaveAttribute('data-configuration-status', 'on')
+  });
+  it('should render with status off', async () => {
+    renderComponentView('off');
+
+    expect(screen.getByTestId('status-pill')).toBeInTheDocument();
+expect(screen.getByTestId('status-pill')).toHaveAttribute('data-configuration-status', 'off')
+  });
+});
--- a/client/src/components/StatusPill/index.tsx
+++ b/client/src/components/StatusPill/index.tsx
@@ -1,0 +1,15 @@
+interface StatusPillProps {
+  status: string;
+}
+
+export function StatusPill({ status }: StatusPillProps) {
+  return (
+    <span
+      data-configuration-status={status}
+      data-testid='status-pill'
+      className="rounded-full px-2 py-1 text-white font-bold"
+    >
+      Refiner {status}
+    </span>
+  );
+}
--- a/client/src/pages/Configurations/index.tsx
+++ b/client/src/pages/Configurations/index.tsx
@@ -1,8 +1,42 @@
 import { Title } from '../../components/Title';
 import { Button } from '../../components/Button';
 import { Search } from '../../components/Search';
+import { ConfigurationsTable } from '../../components/ConfigurationsTable';
+
+
 
 export function Configurations() {
+  const tableData = {
+    columns: ['Reportable condition', 'Status'],
+    data: [
+      {
+        name: 'Chlamydia trachomatis infection',
+        status: 'on',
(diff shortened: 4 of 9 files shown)


client/src/components/ConfigurationsTable/index.tsx
```diff
@@ -0,0 +1,65 @@
+import { Table as UswdsTable } from '@trussworks/react-uswds';
+import { StatusPill } from '../StatusPill';
+
+interface HeaderProps {
+  items: string[];
+}
+
+interface TableBodyProps {
+  data: Array<TableItem>;
```
jakewheeler, line 9: Bit of a nitpick but we can keep this consistent with the way we type arrays elsewhere (such as the `items: string[]` above).
```suggestion
  data: TableItem[];
```

client/src/components/ConfigurationsTable/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+interface TableItem {
+  name: string;
+  status: string;
+  id: string;
+}
+
+interface ConfigurationsTableProps {
+  columns: string[];
+  data: Array<TableItem>;
```
jakewheeler, line 20:
```suggestion
  data: TableItem[];
```

client/src/tailwind.css
```diff
@@ -43,13 +51,67 @@
...
+  @apply bg-gray-200;
+}
+
+.usa-table tbody th {
+  width: 400px;
+}
+
+.usa-table tbody td {
+  width: 720px;
+}
```
jakewheeler, line 117: We should probably consider an alternative approach to hard-coding pixel values. I noticed this is breaking the view on more narrow displays.
image

We could consider using [breakpoints](https://tailwindcss.com/docs/responsive-design) or maybe we make the table's parent container wider (reduce left and right margin) and allow the table to fill 100% of the width? What do you think?

client/src/components/StatusPill/index.tsx
```diff
@@ -0,0 +1,15 @@
+interface StatusPillProps {
+  status: string;
```
jakewheeler, line 2: Would it make sense to only allow `on` and `off`?

```suggestion
  status: 'on' | 'off';
```

client/src/tailwind.css
```diff
@@ -18,6 +18,14 @@
...
     @apply !m-0;
   }
+
+  [data-configuration-status='on'] {
+    @apply bg-state-success-dark;
+  }
+
+  [data-configuration-status='off'] {
+    @apply bg-state-error-dark;
+  }
```
jakewheeler, line 28: Since these styles are specific to `StatusPill` I'd suggest making the styling local to that component. I think a good way to do this would be to delete this entirely and handle the background color based on the value given to `StatusPill`. We can use `classNames` to make this all very neat.

```tsx
import classNames from 'classnames';

interface StatusPillProps {
  status: 'on' | 'off';
}

export function StatusPill({ status }: StatusPillProps) {
  const styles = classNames('rounded-full px-2 py-1 font-bold text-white', {
    'bg-state-success-dark': status === 'on',
    'bg-state-error-dark': status === 'off',
  });
  return (
    <span
      data-configuration-status={status}
      data-testid="status-pill"
      className={styles}
    >
      Refiner {status}
    </span>
  );
}
```

It looks like this may also eliminate the need for `data-configuration-status`? What do you think?

client/src/tailwind.css
```diff
@@ -43,13 +51,67 @@
...
+  */
+
+.usa-table {
+  @apply border-gray-cool-30 w-full rounded-sm border-1;
+  box-shadow: 0px 4px 8px 0px rgba(0, 0, 0, 0.15);
+}
+
+.usa-table--borderless thead th {
+  @apply text-blue-cool-90 bg-gray-cool-10 border-none py-3 uppercase;
+  line-height: 26px;
```
jakewheeler, line 94: Tailwind allows us to use [custom values ](https://tailwindcss.com/docs/line-height#using-a-custom-value)so we can do something like this for `line-height`:

```suggestion
  @apply text-blue-cool-90 bg-gray-cool-10 border-none py-3 uppercase leading-[26px];
```

client/src/pages/Configurations/index.tsx
```diff
@@ -1,8 +1,40 @@
 import { Title } from '../../components/Title';
 import { Button } from '../../components/Button';
 import { Search } from '../../components/Search';
+import { ConfigurationsTable } from '../../components/ConfigurationsTable';
 
 export function Configurations() {
+  const tableData = {
+    columns: ['Reportable condition', 'Status'],
+    data: [
```
jakewheeler, line 9: Looks good!

client/src/tailwind.css
```diff
@@ -43,13 +51,67 @@
...
+  * README:
+  * Overriding USWDS styles here with a mixture of raw CSS and applying
+  * Tailwind classes as needed. This makes having the minimal amount of classes
+  * in the components that leverage the USWDS Table HTML to not require having
+  * a lot of Tailwind class names.
+  */
+
+.usa-table {
+  @apply border-gray-cool-30 w-full rounded-sm border-1;
+  box-shadow: 0px 4px 8px 0px rgba(0, 0, 0, 0.15);
```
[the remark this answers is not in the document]
rogeruiz: hmm yeah. I like this approach a lot more. I thought that the TailwindCSS was a bit harder to read, but it's really the `0` not requiring a suffix that threw me off.

client/src/tailwind.css
rogeruiz: 🚀 nice! thanks for this. This makes more sense and removes the need for any additional attributes for these components. I'll refactor the whole component and tests locally but incorporate your suggestions from the review.

client/src/components/StatusPill/index.tsx
rogeruiz: [de46652](https://github.com/CDCgov/dibbs-ecr-refiner/pull/187/commits/de4665247a5de577b83aa0508b53527f07c270c9)

client/src/tailwind.css
rogeruiz: Resolved by [de46652](https://github.com/CDCgov/dibbs-ecr-refiner/pull/187/commits/de4665247a5de577b83aa0508b53527f07c270c9). I did leave the width of the columns to be balanced for now. I will need to do some more research to figure out how to best set the width of the first column, but I wanted to get the code into `main` and not hold up this code review to perfect the column widths.

client/src/components/StatusPill/index.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  it('should render with status off', async () => {
+    renderComponentView('off');
+
+    expect(screen.getByTestId('status-pill')).toBeInTheDocument();
+    expect(screen.getByTestId('status-pill')).toHaveAttribute(
+      'data-configuration-status',
+      'off'
+    );
+  });
+});
```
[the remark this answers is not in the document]
rogeruiz: Resolved by [de46652](https://github.com/CDCgov/dibbs-ecr-refiner/pull/187/commits/de4665247a5de577b83aa0508b53527f07c270c9)

PR #190, opened 2025-07-09
docs: add updates for searching by version for both rs and condition groupers in the TES

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This is just to make sure that we document the pattern that we'll end up leveraging for grabbing specific versions of the ValueSet resources associated with either the reporting specification or condition grouper.

## 🔗 Related Issue

Fixes #
- #

## ✅ Acceptance Criteria

This is just to make sure we track the different things that work for the TES API.

## 🧪 How to test

So long as you have an API key for the TES and you follow the instructions in the README in `docs/tes/` you should be able to use the `.http` file to run all of the requests.

### DIFF
--- a/docs/tes/explore-tes.http
+++ b/docs/tes/explore-tes.http
@@ -1,85 +1,174 @@
 ### 1. Basic Server Information
+
 # Get CapabilityStatement - Returns metadata about the FHIR server's capabilities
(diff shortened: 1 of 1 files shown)


jakewheeler (approved):
Working as expected for me. Thanks Robert!

PR #191, opened 2025-07-10
build: Support `just` command-runner

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This brings in the Just command-runner to be used with this project. For this
patch, it's only focused on `client/` commands and wraps `npm run ...` with
`just client run ...` or `just c r ...` to run the NPM commands from any
directory in the project.

## 🔗 Related Issue

- n/a

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

After successfully installing `just` on your machine, you can run the following
commands to get started using the commands that this patch introduces.

- `just help`
- `just client help`

## ℹ️ Additional Information

This work isn't trying to use Just for everything. Use your discretion when
adding commands that you may find useful. Use modules to split commands into
sub-commands if it's ergonomic. Read more about Just in their friendly manual:
[https://just.systems/man/en/](https://just.systems/man/en/).

I'm excited to see what others come up with! 🚀

### DIFF
--- a/.justscripts/just/client.just
+++ b/.justscripts/just/client.just
@@ -1,0 +1,20 @@
+alias h := _default
+alias help := _default
+
+alias d := dev
+alias r := run
+
+@_default:
+    just --list client
+
+set working-directory := '../../client'
+
+[doc('Bring up development environment using Docker Compose')]
+dev:
+    docker-compose -f '../../docker-compose.yaml' up
+
+
+[doc('Run arbitrary `npm run *` commands within `client/`')]
+run *ARGS:
+    @test  '{{ ARGS }}' == '' && npm run || echo
+    @for action in {{ ARGS }}; do npm run $action; done
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # DIBBs eCR Refiner
 
+> [!TIP]
+> This project leverages `just` as a command runner. To learn more about `just`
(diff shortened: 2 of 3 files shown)


.justscripts/just/client.just
```diff
@@ -0,0 +1,20 @@
...
+alias help := _default
+
+@_default:
+    just --list client
+
+alias d := dev
+
+[doc('Bring up development environment using Docker Compose')]
+dev:
+    docker-compose -f {{ absolute_path('../docker-compose.yaml') }} up
```
robertmitchellv, line 13: similar to the postgres seeding it seems like we're using the python `docker-compse` library that we don't have in our requirements or requirements-dev files. i think we should use the go addition to the docker cli instead (`docker compose`). also would it be too much trouble to have this run in detached mode? something like:

```suggestion
    docker compose -f {{ absolute_path('../docker-compose.yaml') }} up -d
```

rogeruiz: I've updated the command to be a wrapper instead with the required ARGS parameters for the recipe. Now the commands can be run like.

- `just client dev up -d`
- `just client dev down`
- `just client dev restart`
- `just client dev --help`
- etc

robertmitchellv: this is awesome! thanks Roger! 🏆

PR #195, opened 2025-07-15
feat(api): Make /api/healthcheck database connection aware

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

With the addition of #171, we should have a health check endpoint that is
database aware and reports health based on whether it's possible to connect to
the database URL that exists in the environment.

## 🔗 Related Issue

- #169

## ✅ Acceptance Criteria

When the database is not available, the application health check endpoint should
return a 503 Service Unavailable and JSON body stating the health of the
database and the API.

## 🧪 How to test

Run the `refiner/` code application without a database present. Checking the API
endpoint for `api/healthcheck` should return a `503` along with the JSON stating
the status of the API and the database.

Running the Docker Compose with the PostgreSQL database running will return a
status of `200` along with JSON stating the status of the API and the database.


refiner/app/main.py
```diff
@@ -22,16 +26,31 @@
 
 # define health check endpoint at the service level
 @router.get("/healthcheck")
-async def health_check() -> dict[str, str]:
+async def health_check() -> JSONResponse:
     """
     Check service health status.
 
     Returns:
         dict[str, str]: Service status response:
```
jakewheeler, line 34:
```suggestion
        JSONResponse: Service status response:
```

jakewheeler (approved):
Added one docs update suggestion but otherwise this looks good. Thanks Roger!

PR #205, opened 2025-07-16
feat: trigger-based data pipeline in PostgreSQL working example

### LINKED ISSUE #188: [SPIKE] Investigate Trigger-Based Data Refinement Pipeline in PostgreSQL
## 🚧 What needs to be done

This spike is to research and validate the proposed trigger-based data processing architecture outlined in the `normalized-tes-with-denormalized-cache.dbml` document. The goal is to understand the feasibility, performance implications, and potential challenges of using a chain of PostgreSQL triggers to automatically populate a denormalized cache table from normalized source-of-truth tables.

## 🤔 Why it needs to be done

The proposed architecture relies heavily on database triggers (`trigger_update_parent_on_child_change` and `trigger_update_refinement_cache`) to ensure data consistency and performance. Before committing to this design, we need to verify that it will be robust, maintainable, and scalable for our needs. This investigation will help us identify any potential impediments and ensure the Python backend can integrate with it smoothly.

## ⏰ Timebox

A timebox of **3 days** should be sufficient to research the core concepts, build a small proof-of-concept, and document the findings.

## 📝 To-do list

- [ ] **Define objectives and goals**
  - [ ] Document the key questions this spike aims to answer about performance, maintainability, and transactional integrity.
  - [ ] Define the success criteria for the proof-of-concept.
- [ ] **Research + Analysis**
  - [ ] Research best practices and common pitfalls for using PL/pgSQL triggers for data propagation and aggregation.
  - [ ] Analyze potential real-world impediments, such as trigger recursion, debugging complexity, and performance bottlenecks under load.
  - [ ] Investigate strategies for initial data seeding and handling large-scale batch updates from the TES API source.
  - [ ] Evaluate strategies for testing the trigger logic both in isolation (unit tests) and as part of an integration test suite.
  - [ ] Consider how database migrations and schema changes would be managed with this trigger-heavy architecture.
- [ ] **Prototype Development (Proof of Concept)**
  - [ ] Create a local PostgreSQL database with the schema defined in the DBML file.
  - [ ] Implement the two core triggers (`trigger_update_parent_on_child_change` and `trigger_update_refinement_cache`) using PL/pgSQL.
  - [ ] Mock up test data for `tes_reporting_spec_groupers`, `tes_condition_groupers`, `tes_condition_grouper_references`, and `configurations`.
  - [ ] Write a test script to simulate the data flow:
    - [ ] Insert a single child row and its reference, and verify the parent and cache tables are updated correctly.
    - [ ] Insert multiple child rows for the same parent and verify the aggregation works as expected.
    - [ ] Simulate a user adding a `configuration` and verify the `refinement_cache` is updated for that specific jurisdiction.
- [ ] **Documentation of Findings**
  - [ ] Summarize the pros and cons of this trigger-based approach.
  - [ ] Document the results of the POC, including the trigger code and test scripts.
  - [ ] Provide a recommendation on whether to proceed with this architecture and any suggested modifications.
- [ ] **Review and Finalize**
  - [ ] Review the findings and recommendation with the team.

## 📜 Additional context

The backend application is built with Python. The findings should consider the interaction between the database logic and the application layer. For example, how does the application get notified of errors that occur within a trigger? How should transactions be managed from the Python side to ensure atomicity?

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This is the beginning of the new `/database` service. Currently we're using `docker compose` to spin up PostgreSQL but we will eventually move to a process where a Dockerfile will build an image with flat files (from the TES) for seeding. This sets up the env, has a schema, the two triggers, some ValueSet resources for the initial test data (that i'll likely use to continue this work with seeding in subsequent tickets), and the script that adds data to test the triggers (from the ValueSet resources).

## 🔗 Related Issue

Fixes #188 
- #188 

## 🧪 How to test

The `README.md` in `/database` will explain how to test; there is both a unit test and a `seed.py` script that will sort of do an integration test of the triggers.

### DIFF
--- a/database/docker-compose.yaml
+++ b/database/docker-compose.yaml
@@ -1,0 +1,25 @@
+services:
+  postgres-dev:
+    image: postgres:16
+    restart: unless-stopped
+    # these variables are read from the .env file
+    environment:
+      POSTGRES_DB: ${POSTGRES_DB}
+      POSTGRES_USER: ${POSTGRES_USER}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
+    ports:
+      - "${POSTGRES_PORT}:5432"
+    volumes:
+      # The '1_', '2_', and '3_' prefixes ensure they run in the correct order.
+      - ./schema/01-base-schema.sql:/docker-entrypoint-initdb.d/1_schema.sql
+      - ./triggers/01-aggregate-base-groupers.sql:/docker-entrypoint-initdb.d/2_triggers.sql
+      - ./triggers/02-update-refinement-cache.sql:/docker-entrypoint-initdb.d/3_cache_trigger.sql
+      - postgres_data:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+volumes:
+  postgres_data:
--- a/database/schema/01-base-schema.sql
+++ b/database/schema/01-base-schema.sql
@@ -1,0 +1,88 @@
+-- PART 1: user identity and jurisdiction management
+CREATE TABLE IF NOT EXISTS jurisdictions (
+    id VARCHAR PRIMARY KEY,
+    name VARCHAR UNIQUE NOT NULL,
+    state_code VARCHAR(2),
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS users (
+    id SERIAL PRIMARY KEY,
+    email TEXT UNIQUE NOT NULL,
+    jurisdiction_id VARCHAR NOT NULL REFERENCES jurisdictions(id),
+    full_name VARCHAR,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    last_login TIMESTAMP WITH TIME ZONE
+);
+
+-- PART 2: the "source of truth"--normalized relational model
+CREATE TABLE IF NOT EXISTS tes_condition_groupers (
+    canonical_url VARCHAR NOT NULL,
+    version VARCHAR NOT NULL,
+    display_name VARCHAR NOT NULL,
+    loinc_codes JSONB,
+    snomed_codes JSONB,
+    icd10_codes JSONB,
+    rxnorm_codes JSONB,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (canonical_url, version)
+);
+
+CREATE TABLE IF NOT EXISTS tes_reporting_spec_groupers (
+    canonical_url VARCHAR NOT NULL,
+    version VARCHAR NOT NULL,
+    display_name VARCHAR,
+    snomed_code VARCHAR NOT NULL,
+    loinc_codes JSONB,
+    snomed_codes JSONB,
+    icd10_codes JSONB,
+    rxnorm_codes JSONB,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (canonical_url, version),
+    UNIQUE (snomed_code, version)
+);
+
+CREATE TABLE IF NOT EXISTS tes_condition_grouper_references (
+    id SERIAL PRIMARY KEY,
+    parent_grouper_url VARCHAR NOT NULL,
+    parent_grouper_version VARCHAR NOT NULL,
+    child_grouper_url VARCHAR NOT NULL,
+    child_grouper_version VARCHAR NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (parent_grouper_url, parent_grouper_version) REFERENCES tes_condition_groupers(canonical_url, version) ON DELETE CASCADE,
+    FOREIGN KEY (child_grouper_url, child_grouper_version) REFERENCES tes_reporting_spec_groupers(canonical_url, version) ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS configurations (
+    id SERIAL PRIMARY KEY,
+    jurisdiction_id VARCHAR NOT NULL REFERENCES jurisdictions(id),
+    child_grouper_url VARCHAR NOT NULL,
+    child_grouper_version VARCHAR NOT NULL,
+    display_name_override VARCHAR,
+    version VARCHAR,
+    loinc_codes JSONB,
+    snomed_codes JSONB,
+    icd10_codes JSONB,
+    rxnorm_codes JSONB,
+    is_active BOOLEAN DEFAULT FALSE,
+    created_by INTEGER REFERENCES users(id),
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (child_grouper_url, child_grouper_version) REFERENCES tes_reporting_spec_groupers(canonical_url, version) ON DELETE CASCADE
+);
+
+-- PART 3: the "serving layer"--denormalized runtime cache
+CREATE TABLE IF NOT EXISTS refinement_cache (
+    snomed_code VARCHAR NOT NULL,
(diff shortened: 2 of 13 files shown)


database/tests/test_database.py
```diff
@@ -0,0 +1,36 @@
...
+def test_database_connection(db_cursor: psycopg.Cursor):
+    """
+    Tests that a connection to the database can be successfully established.
+    """
+
+    assert db_cursor is not None
+    assert not db_cursor.closed
+
+
+def test_tables_exist(db_cursor: psycopg.Cursor):
```
robertmitchellv, line 13: these tests are sort of purposefully basic. the real test is the `seed.py` script but this sets me up for working on the full seeding of the db

database/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+      POSTGRES_DB: ${POSTGRES_DB}
+      POSTGRES_USER: ${POSTGRES_USER}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
+    ports:
+      - "${POSTGRES_PORT}:5432"
+    volumes:
+      # The '1_', '2_', and '3_' prefixes ensure they run in the correct order.
+      - ./schema/01-base-schema.sql:/docker-entrypoint-initdb.d/1_schema.sql
+      - ./triggers/01-aggregate-base-groupers.sql:/docker-entrypoint-initdb.d/2_triggers.sql
+      - ./triggers/02-update-refinement-cache.sql:/docker-entrypoint-initdb.d/3_cache_trigger.sql
```
jakewheeler, line 16: We've already chatted about it and I'm sure we'll get to this very soon, but a DB migration tool would help us a lot already with this sort of thing.

robertmitchellv: i'm sort of on the fence between Flyway, Liquibase, and Sqitch but i do want to nail down one of these as the tool we go with for migrations--do any of these feel like more natural fits to you from your experience?

database/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
+services:
+  postgres-dev:
+    image: postgres:16
+    restart: unless-stopped
+    # these variables are read from the .env file
```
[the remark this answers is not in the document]
robertmitchellv: i created a separate .env file here. do you think i should do something like `.env.sample`? my thought was that it wouldn't matter for now until we build a Dockerfile and that image has connection details that need to be actually correct. i think you could put anything in `.env` and docker compose would just use those right?

the other thing i talked with Roger about is maybe having a `.envrc` at the project root to connect the different `.env` files or something like that? we were talking about this in the context of just earlier this week

database/docker-compose.yaml
robertmitchellv: this looks cool too; let's add it to the list to evaluate

database/triggers/01-aggregate-base-groupers.sql
```diff
@@ -0,0 +1,90 @@
+-- =============================================================================
+-- TRIGGER 1: aggregation--when child `tes_reporting_spec_groupers`
+-- are linked to a parent `tes_condition_grouper`, a trigger fires
+-- to aggregate all unique codes (LOINC, SNOMED, ICD-10, RxNorm)
+-- from the children into the parent's `jsonb` columns
+-- =============================================================================
+
+CREATE OR REPLACE FUNCTION update_parent_on_child_change()
```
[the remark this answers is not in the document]
jakewheeler: This is decently readable overall IMO!

database/triggers/01-aggregate-base-groupers.sql
```diff
@@ -0,0 +1,90 @@
...
+        WHERE canonical_url = parent_url_to_update AND version = parent_version_to_update;
+    END IF;
+
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+-- create a trigger that fires when the relationship between groupers changes
+CREATE TRIGGER trigger_update_parent_on_ref_change
+AFTER INSERT OR UPDATE OR DELETE ON tes_condition_grouper_references
```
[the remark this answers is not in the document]
robertmitchellv: that's actually something that i'm working on right now. i might add it to this PR tbh but i think we need a `trigger_error_log` table so we know if any trigger didn't fire and why. i think this might be better than just outputting to the console since it'll probably get us to the point at which we might elect to use some of the AWS tooling for this instead. at least we'd be able to decide what solution is the right fidelity matching our tool

database/seed.py
```diff
@@ -0,0 +1,238 @@
...
+                tables = [
+                    "refinement_cache",
+                    "configurations",
+                    "users",
+                    "jurisdictions",
+                    "tes_condition_grouper_references",
+                    "tes_reporting_spec_groupers",
+                    "tes_condition_groupers",
+                ]
+                cur.execute(f"TRUNCATE {', '.join(tables)} RESTART IDENTITY CASCADE;")
```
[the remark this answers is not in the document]
robertmitchellv: i think this `seed.py` is pretty brittle and is like, very tightly coupled to a happy path. i wonder why it did that?? can you tell me what happens if you do this:

```bash
docker compose down -v
```

and then restart with `docker compose up --build`; does it still do that?

database/triggers/01-aggregate-base-groupers.sql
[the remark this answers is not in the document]
jakewheeler: Awesome, that seems like a good place to start.

database/schema/01-base-schema.sql
```diff
@@ -0,0 +1,88 @@
...
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    last_login TIMESTAMP WITH TIME ZONE
+);
+
+-- PART 2: the "source of truth"--normalized relational model
+CREATE TABLE IF NOT EXISTS tes_condition_groupers (
+    canonical_url VARCHAR NOT NULL,
+    version VARCHAR NOT NULL,
+    display_name VARCHAR NOT NULL,
+    loinc_codes JSONB,
```
jakewheeler, line 25: I'm realizing in almost all situations we are actually using only arrays. The only place I noticed where we do want a full JSON object is `refinement_cache.source_details`. 

Do you think we should switch all of the other `JSONB` columns to `TEXT[]` instead? This feels like it'd be a good move since we'd get some extra safety around what we're expecting the shape of the data to look like and it'd also make various other things easier (adding constraints, using Postgres functions to directly manage the arrays, querying, etc.).

What do you think?

database/schema/01-base-schema.sql
```diff
@@ -0,0 +1,88 @@
...
+    id VARCHAR PRIMARY KEY,
+    name VARCHAR UNIQUE NOT NULL,
+    state_code VARCHAR(2),
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TABLE IF NOT EXISTS users (
+    id SERIAL PRIMARY KEY,
+    email TEXT UNIQUE NOT NULL,
```
jakewheeler, line 12: I wonder if we should consider limiting length in situations like this? 🤔  

Also, I believe `VARCHAR` == `TEXT` in situations where we don't specify a length. Should we swap this out for `TEXT` when we want unlimited length just so it's super clear what the intention is?

database/triggers/01-aggregate-base-groupers.sql
```diff
@@ -0,0 +1,61 @@
...
+
+    -- if we found a parent, proceed with the update
+    IF parent_url_to_update IS NOT NULL THEN
+        -- call the helper function for each code type to get the aggregated list
+        UPDATE tes_condition_groupers
+        SET
+            loinc_codes = get_aggregated_child_codes(parent_url_to_update, parent_version_to_update, 'loinc_codes'),
+            snomed_codes = get_aggregated_child_codes(parent_url_to_update, parent_version_to_update, 'snomed_codes'),
+            icd10_codes = get_aggregated_child_codes(parent_url_to_update, parent_version_to_update, 'icd10_codes'),
+            rxnorm_codes = get_aggregated_child_codes(parent_url_to_update, parent_version_to_update, 'rxnorm_codes')
```
jakewheeler, line 44: Love it 🔥

database/schema/01-base-schema.sql
robertmitchellv: i think the test data i used in `seed.py` would make the case that a text array would be better, but when we seed those tables with data we'll be using objects, e.g.,:

```json
{"code": "1234", "display": "display name"},
{"code": "1235", "display": "display name"},
{"code": "1236", "display": "display name"},
```

let me know if i'm wrong about this; i thought that text arrays were more like lists and json be was more like a dictionary (thinking from python)

my thinking was that the UI will be using the objects to have that 1:1 connection between the codes and the names. but i think you're right about using a text array for the cache table--especially for `aggregated_codes` since it will just be a flat text array of codes.

i also don't think any of the triggers would need to be altered too much for that. if it sounds good let me know and i can make those changes

database/schema/01-base-schema.sql
jakewheeler: > i agree that it seems best to use TEXT for anything that doesn't have a data validation related element to it; like state code, or jurisdiction id (once we get the list from APHL), etc. in a case where we want to enforce validation i think doing something like VARCHAR(n) would make sense and for everything else TEXT since one could wonder if maybe the (n) is missing because of an accident?

💯  I think you're correct. From what I've been reading it seems to function exactly the same. It should make things more clear to readers without causing any performance impact.

database/schema/01-base-schema.sql
robertmitchellv: ok, this should be good to go now; we're using `TEXT[]` in the cache table now

database/schema/01-base-schema.sql
robertmitchellv: all the `VARCHAR`s have been changed to `TEXT` save the few like `jurisdiction_id` and `state_code`. i updated the seed data for jurisdiction as well

jakewheeler (approved):
Good work Robert! 🎉

PR #208, opened 2025-07-17
style: Nav links should display on small screens

jakewheeler:
This PR is a small fix to make sure that the `Configurations` and `Testing` links remain visible on narrow screens. Without this fix they'll disappear when the screen gets too small.

I also included some styling to ensure the pill component's text no longer wraps when displayed on smaller screens.

Here's an example:
Kapture 2025-07-17 at 13 38 08


rogeruiz (approved):
lgtm 🌈

PR #219, opened 2025-07-21
feat: normalize unrefined eicr and refined eicrs for better diff comparison

### LINKED ISSUE #177: [TASK] Update formatting of refined/unrefined eICR for diff comparison #176
## 🚧 What needs to be done

We need to update our code so that the formatting of the unrefined eICR and condition refined eICR matches 

## 🤔 Why it needs to be done

Without this change, it looks like 80% of the document was updated when it was not. 

When we refine an eICR for a condition we serialize it so that we can make the necessary changes, but when we do this is changes some of the formatting. So when we compare the unserialized unrefined eICR to the refined eICR in the diff tool it looks like way more changed than actually was 

## 🏗️ Acceptance Criteria

- [ ] Comparison of an unrefined eICR and a refined eICR will only show the changes the refiner made (not formatting changes)
- [ ] Update the demo endpoint to make the necessary changes to the unrefined and refined eICR so that the comparison is accurate 

## ⚙️ Technical Notes

This is what it currently looks like in the diff tool, and since we actually arent or shouldnt be making updates to the first  lines and others like it we need to fix this 

Conversations from dev chat 

- Seems like we could serialize the original eicr in the same way and then have a more similar document to diff. Or we could change how we refine and not serialize the whole document, just identify the pieces that are being updated, serialize and change those pieces and slice it all back together.

- ah right; this is what it is sometimes a pain with the unit tests where we have to use the normalize_xml functions to compare things right?

- Maybe we move that function into a formatter service and make use of it for the demo response?

- ok this is actually a pain because we only remove comments from processed sections, instead of the whole document, and so then if i remove them from the whole unrefined eicr it now looks like we are adding in comments with the refiner. this is going to probably need some more work around it. im going to leave it where im at now and just focus on the new diff tool and make a story to update the files we are comparing and what sort of formatting we are doing to each

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary

Update demo.py to normalize (format and remove all comments) refined and unrefined eicrs for better diff comparison on the front end 

## 🔗 Related Issue

Fixes #177 
- #177

## ✅ Acceptance Criteria

- [ ] Comparison of an unrefined eICR and a refined eICR will only show the changes the refiner made (not formatting changes)
- [ ] Update the demo endpoint to make the necessary changes to the unrefined and refined eICR so that the comparison is accurate

## 🧪 How to test

- Hard to test without diff tool implemented yet, but can deploy and add in these lines to success.tsx to see what it looks like but also demoed what this looks like last week 

`import ReactDiffViewer, { DiffMethod } from 'react-diff-viewer-continued';

      <ReactDiffViewer
          oldValue={unrefinedEicr}
          newValue={selectedCondition.refined_eicr}
          splitView={true}
          showDiffOnly={false}
          compareMethod={DiffMethod.WORDS_WITH_SPACE}
          leftTitle={'Unrefined eICR'}
          rightTitle={'Refined eICR'}
      />`

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -20,6 +20,7 @@
     "@trussworks/react-uswds": "^10.0.1",
     "classnames": "^2.5.1",
     "react": "^19.0.0",
+    "react-diff-viewer-continued": "^4.0.6",
     "react-dom": "^19.0.0",
     "react-router": "^7.5.2",
     "react-xml-viewer": "^2.0.5",
--- a/refiner/app/api/v1/demo.py
+++ b/refiner/app/api/v1/demo.py
@@ -17,7 +17,7 @@ from ...core.exceptions import (
     XMLValidationError,
     ZipValidationError,
 )
-from ...services import file_io, refine
+from ...services import file_io, format, refine
 
 # Keep track of files available for download / what needs to be cleaned up
 REFINED_ECR_DIR = "refined-ecr"
@@ -330,13 +330,21 @@ async def demo_upload(
         # Store the combined zip
         _update_file_store(output_file_name, output_file_path, token)
 
+        # normalize/format condition refined eicrs and unrefined eicr for cleaner diff comparison
+        for cond in conditions:
+            raw_xml = cond["refined_eicr"]
+            # normalize and overwrite (or store to a new key)
+            cond["refined_eicr"] = format.normalize_xml(raw_xml)
+
+        normalized_unrefined_eicr = format.normalize_xml(original_xml_files.eicr)
+
         return JSONResponse(
             content=jsonable_encoder(
                 {
                     "message": "Successfully processed eICR with condition-specific refinement",
                     "conditions_found": len(conditions),
                     "conditions": conditions,
-                    "unrefined_eicr": original_xml_files.eicr,
+                    "unrefined_eicr": normalized_unrefined_eicr,
                     "processing_notes": [
                         "Each condition gets its own refined eICR",
                         "Sections contain only data relevant to that specific condition",
--- a/refiner/app/services/format.py
+++ b/refiner/app/services/format.py
@@ -1,0 +1,28 @@
+from lxml import etree
+from lxml.etree import _Element
+
+
+def normalize_xml(xml: str) -> str:
+    """
+    Normalize XML string for comparison: parse, strip out all comments, then re-serialize with consistent pretty-printing.
+    """
+
+    # Parse into an Element
+    root: _Element = etree.fromstring(xml)
+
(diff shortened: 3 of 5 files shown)


client/package.json
```diff
@@ -20,6 +20,7 @@
     "@trussworks/react-uswds": "^10.0.1",
     "classnames": "^2.5.1",
     "react": "^19.0.0",
+    "react-diff-viewer-continued": "^4.0.6",
     "react-dom": "^19.0.0",
     "react-router": "^7.5.2",
     "react-xml-viewer": "^2.0.5",
```
jakewheeler, line 26: Are we good to remove `react-xml-viewer` now?

refiner/app/services/format.py
```diff
@@ -0,0 +1,34 @@
...
+
+    # Reject non‐string inputs
+    if not isinstance(xml, str):
+        raise ValueError(f"Expected XML as str, got {type(xml).__name__!r}")
+
+    # Parse into an Element
+    root: _Element = etree.fromstring(xml)
+
+    # Force mypy to treat this as a list of Elements
+    comment_nodes: list[_Element] = root.xpath("//comment()")  # type: ignore[assignment]
```
jakewheeler, line 18: We use `cast` in a few other parts of the code base where do something similar. This should work for you here too:
```suggestion
    comment_nodes = cast(list[_Element], root.xpath("//comment()"))
```

We'll need to add `from typing import cast` the top of the file 🫡

refiner/app/api/v1/demo.py
```diff
@@ -330,13 +330,21 @@ async def demo_upload(
         # Store the combined zip
         _update_file_store(output_file_name, output_file_path, token)
 
+        # normalize/format condition refined eicrs and unrefined eicr for cleaner diff comparison
+        for cond in conditions:
+            raw_xml = cond["refined_eicr"]
+            # normalize and overwrite (or store to a new key)
+            cond["refined_eicr"] = format.normalize_xml(raw_xml)
```
jakewheeler, line 337: Do you think it'd make sense to move the `format.normalize_xml()` up to the loop above where we're building `conditions` instead of looping through and updating all of the conditions again here?

Something like this:
```python
            conditions.append(
                {
                    "code": condition_info["code"],
                    "display_name": condition_info["displayName"],
               --> "refined_eicr": format.normalize_xml(condition_refined_eicr),
                    ...
```

lina-roth: Oh thats a good point. I think I just assumed I should put it after where the upload to the file store is but yeah I see now that what youre saying will work. I will make that change

PR #222, opened 2025-07-23
refactor: seeding new PostgreSQL data model

### LINKED ISSUE #182: [TASK] Implement data model
## 🚧 What needs to be done

With the MVP data model decided by the team in #178 we now want to go ahead and implement this model so that the application can make use of it.

1. SQL scripts are written based on data model in #178 
2. SQL scripts are applied to DB
3. App code is updated as needed
4. Seeding script is updated as needed
5. Tests are updated as needed

## 🤔 Why it needs to be done

This implementation will represent the initial version (or close to it) of the data model used by users in the MVP.

## 🏗️ Acceptance Criteria

See requirements above.

## ⚙️ Technical Notes

We do not necessarily need to consider future migrations as a part of this ticket.

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This sets up seeding of the database in the `database/` directory. Some of the things that have been implemented:

- A process that will download flat files from the TES based on their grouper type and version (will work for versions we have not yet seen before too)
- A process that will check if the flat files are new (the `manifest.json` contains the hash for the files with metadata for when we pulled the data down from the TES)
- A database seeding script that works from the flatfiles to populate data in PostgreSQL
- A Dockerfile where the assets and scripts can be built into the image (seeding happens when the container first starts up
- Some unit and integration tests as well as a sanity check on the database
## 🔗 Related Issue

Fixes #182 
- #182 

## 🧪 How to test

The `database/README` will have all of the details to get this working. If something there isn't working please let me know so I can update 🙏🏻 

## ℹ️ Additional Information

There is still time to think through other additions we will need to make when connecting the server code to this database so anything that feels like it should come after this work, let's keep a tally of those thoughts 💭


.pre-commit-config.yaml
```diff
@@ -10,8 +10,8 @@ repos:
     # - prevent large file commits
     -   id: check-added-large-files
         args:
-        # 20mb == 20480kb
-        -   --maxkb=20480
+        # 30mb == 35,840kb
+        -   --maxkb=35840
```
robertmitchellv, line 14: i imagine this won't be a popular addition but we need it for the json files we're downloading from the TES.

database/data/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
```
robertmitchellv, line 1: if there are any other details you think we should capture in the manifest let me know; this is just a first pass for how to check if anything has changed when updating data from the TES. eventually i think a cron style github action would be great for this code; it could run every sunday and if there are changes a PR can be opened (i really want to take a stab at this; sounds fun)

.pre-commit-config.yaml
jakewheeler: Do these files need to be versioned with git? I wonder if we could put them in blob storage or something. It's kind of nice to have them living here but I see the `database/data` folder is already 70MB 💀

database/README.md
```diff
@@ -1,86 +1,143 @@
...
+Run the pipeline script to download the latest ValueSet data from the TES source API. This will populate the `./data` directory with the JSON files needed for seeding.
 
-Use Docker Compose to build the PostgreSQL container. This command will also create the database schema and apply the triggers from the `schema/` and `triggers/` directories.
+> [!NOTE]
+> You will need an API key and it will need to be in your `.env` file in order for the scripts to work.
 
 ```bash
-docker compose up -d
+# in the database/ directory
+python pipeline/detect_changes.py
```
jakewheeler, line 70: What am I doing wrong here? 🤔 
```sh
(venv) ➜  database git:(robert/182-seed-new-model) python pipeline/detect_changes.py 
  File "/<removed>/dibbs-ecr-refiner/database/pipeline/detect_changes.py", line 10, in <module>
    from pipeline.fetch_api_data import run_fetch_pipeline
ModuleNotFoundError: No module named 'pipeline'
```

database/pipeline/detect_changes.py
```diff
@@ -0,0 +1,140 @@
...
+    Args:
+        filepath: The path to the file to be hashed.
+
+    Returns:
+        The hex digest of the file's SHA256 hash.
+    """
+
+    sha256 = hashlib.sha256()
+    with open(filepath, "rb") as file:
+        while chunk := file.read(8192):
```
jakewheeler, line 42: What's the significance of `8192`? Is this arbitrary?

Maybe worth turning this into a variable with a meaningful name so it's obvious.

database/README.md
```diff
@@ -1,86 +1,143 @@
...
+- `tests/`: Contains integration and unit tests for the database logic and pipeline scripts.
+- `triggers/`: Contains the SQL trigger definitions that handle ongoing, incremental data updates after the initial seed.
 
 ## Core Concepts
 
-The data pipeline works in two stages, automated by database triggers:
+The database is designed to pre-calculate and cache refined code sets, ensuring high performance for the main Refiner application. While database triggers are in place to handle incremental updates during normal operation, the initial setup is performed by a dedicated seeding script for reliability and speed.
 
-1.  **Aggregation (Trigger 1):** When child `tes_reporting_spec_groupers` are linked to a parent `tes_condition_grouper` (or when the data in a child grouper is updated), this trigger fires. It aggregates all unique codes (LOINC, SNOMED, etc.) from all children into the parent's `jsonb` columns.
+### The `refinement_cache` Table
```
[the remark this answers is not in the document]
jakewheeler: > blank slate configurations

Can you speak to this more? I just wanted to be sure I'm understanding what you mean here. I see one config was seeded and one item in `refiner_cache` also exists which seems good to me, but I think you're asking about something different?

database/scripts/check_seeded_db.py
```diff
@@ -0,0 +1,186 @@
...
+                console,
+                "No Orphaned References",
+                """
+                    SELECT COUNT(*)
+                    FROM tes_condition_grouper_references ref
+                    LEFT JOIN tes_reporting_spec_groupers child
+                    ON ref.child_grouper_url = child.canonical_url AND ref.child_grouper_version = child.version
+                    WHERE child.canonical_url IS NULL;
+                """,
+                lambda res: res[0][0] > 0,
```
jakewheeler, line 128: I feel like this is kind of confusing at a glance. Do you think using [dict_row](https://www.psycopg.org/psycopg3/docs/advanced/rows.html#row-factories) like we do in the app would make things more clear? Pretty minor thing overall, just figured I'd mention it!

.pre-commit-config.yaml
robertmitchellv: i don't think the individual files will get much bigger, but there will be other versions dropped in here in the future. i'd love to keep the flat files in git so that the manifest and files can stay tracked but if it turns out that won't work, we can try something else

database/README.md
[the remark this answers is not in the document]
robertmitchellv: what i mean by a "blank slate configuration" is similar to what we did in the sqlite3 db, where we had a sort of empty configuration but it had its own condition grouper as the only real data in that row

ideally when we connect the app to this db, i'd want to be able to upload files (other than mon mothma) to test but there needs to be a difference between a sort of basic condition grouper style filtering and one that is set up to run in prod on the lambda.

but this begs an important question: do we let the testing suite work if there isn't a configuration set up for a condition? if yes, what is that specific pattern and expected outcome. if no, what should we do if a user uploads an eICR/RR pair that is for conditions there are no configurations for?

i think this is the big thing that needs to be answered before we connect this db work to our app imo

database/README.md
robertmitchellv: i can also update the docs to use the module style too

database/pipeline/detect_changes.py
robertmitchellv: i just added some comments to explain, but this is just to process the file in 8kb chunks as a way to handle larger files. the example for hashing files i looked at used 8kb so that value was arbitrarily chosen and could be tweaked if we need. you can take a look at the "Note" in the docstring and the comments to explain--i sort of forgot about this, thanks for flagging!

jakewheeler: Perfect, looks good!

database/README.md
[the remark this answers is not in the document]
jakewheeler: This is a great point. Perhaps our seeding script creates a config per condition for our test jurisdiction so that we can continue to run files through. Or at the very least, 2 configs for our monmothma 2 condition test file. That said, I don't know if a solution necessarily needs to be determined and implemented in this PR.

PR #223, opened 2025-07-23
feat: refining process is usable both sync and async

### LINKED ISSUE #179: [TASK] Application should use database connection pool
## 🚧 What needs to be done

The refiner app code currently does not do any connection pooling. It opens and closes a single connection. In order to ensure the app is production ready we will want to switch over to a connection pooling setup.

## 🤔 Why it needs to be done

Opening and closing connections is a heavy process. We want to avoid this when possible and instead grab an open, available connection from the pool.

## 🏗️ Acceptance Criteria

- [ ] Pooling has been configured
- [ ] App has been restructured as needed
- [ ] Tests updated as needed

## ⚙️ Technical Notes

Please refer to [this page](https://www.psycopg.org/psycopg3/docs/advanced/pool.html) as a reference when implementing.

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The primary goal of this PR is to ensure the FastAPI application makes use of the async connection pool when performing database queries. Additionally, we want to ensure that code called in the Lambda _does not_ do this, due to the way Lambda handles DB connections.

In order to achieve this, the main thing I needed to do was create two different refining process functions:
- `refine_sync` - The refining function to be called in the Lambda
- `refine_async` - The refining function to be called within the app (just `/demo` for now, testing suite later)

While the `refine_eicr` function is used inside these functions, it is now "private" and these two functions are the entry point into the refining process.

A few other things that were modified in this PR:
- A small bug fix (I think, at least 😅 ) - [see here
](https://github.com/CDCgov/dibbs-ecr-refiner/pull/223/files#r2226247562)
- Improved a lot of the typing in the `refine.py` file. It should now be a lot more clear as to what data you're working with
- Deleted the `/ecr` and `/file_io` routes since they are unused by the application
- `/healthcheck` uses the DB pool
- Integration tests now use the `/demo/upload` route when checking file uploads

## 🔗 Related Issue

Fixes #179 

## ✅ Acceptance Criteria

- [x] App uses async connection pool when talking to the DB
- [x] Lambda uses a one-off connection, can call refiner

## 🧪 How to test

1. Run `docker compose up`
4. You should be able to be able to run through the demo exactly as before without issue

## ℹ️ Additional Information

- I ran a manual diff check on the refined output of `main` vs. this branch using the default demo files and it was 100% the same. Feel free to manually double check me 😄


refiner/app/api/v1/demo.py
```diff
@@ -318,8 +288,8 @@ async def demo_upload(
         full_zip_output_path = _create_zipfile_output_directory(refined_zip_output_dir)
 
         # Add eICR + RR file as well
-        refined_files_to_zip.append(("CDA_eICR.xml", xml_files.eicr))
-        refined_files_to_zip.append(("CDA_RR.xml", xml_files.rr))
+        refined_files_to_zip.append(("CDA_eICR.xml", original_xml_files.eicr))
+        refined_files_to_zip.append(("CDA_RR.xml", original_xml_files.rr))
```
jakewheeler, line 292: I believe my change is what was originally intended here. Please correct me if I'm wrong!

refiner/app/db/operations.py
```diff
@@ -89,6 +72,58 @@ def get_grouper_by_condition(self, condition: str) -> GrouperRow:
...
+            WHERE condition = %s
+        """
+
+    def _validate_condition(self, condition: str) -> None:
+        if not condition or not isinstance(condition, str):
+            raise InputValidationError(
+                message="Invalid condition code", details={"condition": condition}
+            )
+
+    def _parse_row(self, row: Any, condition: str) -> GrouperRow:
```
jakewheeler, line 115: psycopg offers [tools to provide better typing ](https://www.psycopg.org/psycopg3/docs/api/types.html) but it's a little hard to do right now without a workaround since all of the `groupers` `*_codes` columns should be `jsonb` but are currently `text`.

Should I leave it alone for now? I would like to get DB data types in order once things are more solidified.

refiner/app/services/refiner/helpers.py
```diff
@@ -0,0 +1,102 @@
+import logging
```
jakewheeler, line 1: I will happily accept alternative naming suggestions for this file 😂

refiner/app/db/operations.py
robertmitchellv: i think we can potentially leave this for when we fold the new db into the actual refiner work. i don't know how you feel about this but i can imagine a couple of tickets for this, e.g.,

1. Achieve parity between the two dbs for required things:
  - fold in any of the outstanding tables that are required for downloads, sessions, login, etc, and make sure there's nothing we're missing in `database/`
  - make sure there are configurations that match the test data--we can even pick a few of the new test files to add configurations to as well so we can test the files locally (the ones Geo forwarded)
  - [optional] hit a coverage goal on the unit tests (can be 70% or 80% for now until we start a more firm target and add that to our workflows)
2. cut over to new db
  - swap out the dbs and update the typing
  - make sure the project root's `docker compose` workflow spins everything up correctly
  - run the test data we have configurations for through the testing suite
  - do some QA
  
what do you think? since we're going to be changing quite a bit it feels better to wait

jakewheeler: That seems reasonable! I don't think it makes sense to spend too much more time on things that we don't ever plan to ship.

docs/refiner/refiner-demo-requests.http
robertmitchellv, line 1: i think if we're going to remove a lot of these "make a request decoupled from the UI" types of files, which i agree with btw, we should also potentially move the data-model work into the `database/` directory maybe in a `database/docs/` directory to explain some of the early ideas around what `database/` has turned into. this was something i considered doing with my last PR but i decided not to touch it yet. what do you think?

refiner/app/services/refiner/helpers.py
robertmitchellv: yeah this is a bit confusing, but i'm not sure any of my 💡 would be any better. i suppose some things that come to mind:

for the directory:
- `processing/`
- `engine/`
- `pipeline/`

`refine.py` still works imo

`helpers.py` could be more precise with like `grouper_utils.py` or something similar depending on how specific you want to get; e.g., leave it in one file, split it up for context

i think it's tough thinking of a name tho since we need to move this a level down

refiner/app/services/refiner/refine.py
```diff
@@ -44,6 +46,17 @@
     "Matching Condition Code",
 ]
 
+
+@dataclass
```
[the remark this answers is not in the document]
robertmitchellv: you mean move them into `core/` or something else?

refiner/app/services/refiner/helpers.py
jakewheeler: Could we call this service `ecr/` since we're planning to process RRs as well? 🤔 This feels the most obvious to me if I'm trying to remember where the refiner code lives. Do you think that works?

I like `grouper_utils.py`. I think that's just enough description for what it's doing.

So after these changes we could have:
- `services/ecr/refine.py`
- `services/ecr/grouper_utils.py`

What do you think? Is this better?

robertmitchellv: initially i don't think i liked `ecr` as the name but after considering it a bit more i think this gives us some flexibility to decide whether or not down the road we want to do something like:

- `ecr/eicr-refine.py`
- `ecr/rr-refine.py`
- `ecr/rr-determination.py`

given how different the lambda and the testing suite take shape, it might be nice to have these services separated out a bit more. and i think `ecr` is general and specific enough to work. so i'm good with this; thanks Jake!

docs/refiner/refiner-demo-requests.http
robertmitchellv: i suppose that's true. `database/` might change for sure, but we still have the TES extraction/checksum process. this comment was also before the conversations we had about the changes we decided to make so it's also a lot less of a worry for now. i'm good with things as they stand

Issue #210, opened 2025-07-18
[USER STORY] Configuration Detail Page Navigation


kplaneaux:
## 👤💭 User Story

As a user, I want to navigate between steps of setting up my configuration, so that I can get to the step I need to

## 🏗️ Acceptance Criteria
 
- [ ] Individual steps are clickable links
- [ ] Each step is a new page or sub-frame(?) that contains configuration building, configuration testing, and configuration activation

## 🎨 Designs
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1BIQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a1d24b74fded2869faff1ac34c24e0af50889fc0/image.png" alt="image.png" />
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd1RIQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--dffb77ea66355da72cf2601c37158b45845e6718/image.png" alt="image.png" />
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2JIQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f8ff1d467655527d54c77894a7d56bd64e3f7162/image.png" alt="image.png" />

Issue #181, opened 2025-07-07
[USER STORY] Configuration Detail Page


kplaneaux:
## 👤💭 User Story

As a user, I want to view reportable condition details, so that I can see how my eCRs with that condition are being refined

## 🏗️ Acceptance Criteria
 
- [ ] Setting up a new condition sends the user to the condition detail page 
- [ ] Clicking on a condition row in the table on the configurations page takes a user to the condition detail page - **Ask Michelle about hover state**
- [ ] Page has the condition name as title

## 🎨 Designs
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM0tIQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--652741a9a42a3f10b1450c57ed7d30f7682b7361/image.png" alt="image.png" />
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1hnQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--67333dd708ebb2a8c8ceea09cbdc4925c0bfcc91/image.png" alt="image.png" />

PR #226, opened 2025-07-28
feat: Add configuration detail page

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR addresses the following:

- Routes were added for config building, testing, and activation
- Configurations table row can be clicked (via mouse or keyboard) to navigate to the config builder page
- Navigation bar for the config workflow has been created
    - Visible at the top of each config building step
- Small a11y fix for CDC logo link in the footer 
- Basic UI tests for the new pages
- Delete unused `LandingPageLink` component

## 🔗 Related Issue

Fixes:
- #181
- #210 

## 🧪 How to test

Run the unit tests:

```sh
just client test
```

Manually testing the application:

1. Run `docker compose up`
4. Upon logging in you should be see the configurations table
5. Use your mouse or keyboard to click one of the configurations
6. You will be navigated to the config building page. Use the links at the top and/or the purple button to navigate to the next page
7. You'll notice the URL changing as you navigate between the various different routes

Example:
Kapture 2025-07-29 at 17 36 00

## ℹ️ Additional Information

- How do we feel about this file structure? Does the inner-page nesting make sense? How about the "local" components that are shared but only between a group of pages?


client/src/components/Layout/index.tsx
```diff
@@ -63,6 +63,9 @@ export function Footer() {
             className="inline-block"
           >
             <img src={CdcLogo} alt="" />
+            <span className="sr-only">
+              CDC - U.S. Centers for Disease Control and Prevention
+            </span>
```
jakewheeler, line 68: I was running some Axe checks as I was working and happened to notice this so I went ahead and fixed it.

rogeruiz: thanks for catching this!

rogeruiz (approved):
lgtm 🌈

PR #245, opened 2025-07-30
feat: Allow users to download refined files from an S3 bucket

### LINKED ISSUE #221: [TASK] Refined file downloads should use AWS S3
## 🚧 What needs to be done

The refiner UI allows users to upload an eCR and then download the refined files that are produced. However, there are several issues with the current implementation that need to be addressed so that this feature can continue to exist and be usable in a production environment.

## 🤔 Why it needs to be done

In order to make these feature resilient and reliable, we should store the generated files in AWS S3. This will allow us to:
1. Create an S3 lifecycle policy to delete files after a set period of time
2. Encrypt the files
3. Provide signed download links back to the user, which will only allow the uploader to access the file
4. Allow containers to scale if needed without breaking this functionality (files are currently tracked in-memory)

## 🏗️ Acceptance Criteria

- [ ] Application performs required checks prior to file upload
- [ ] Simulate S3 locally as part of the refiner's `docker-compose.yaml` setup (take a look at [localstack](https://github.com/localstack/localstack))
- [ ] Application should create a refined zip and store them in S3
- [ ] Upload uploading, the application should produce a signed S3 link that can be provided to the user for download
- [ ] We can likely remove the `/download` route from `demo.py` since the returned download link will be a signed S3 link
- [ ] Remove code from `demo.py` and elsewhere related to handling file cleanup

## ⚙️ Technical Notes

Localstack should spin up with the rest of the docker compose infrastructure. We may need some S3 related environment variables added to the application.

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

Users are able to download their refined files once they reach the "success" page of the demo experience. However, the current solution is flimsy and will not function if more than a single application container is running due to storing references to the files in memory. Additionally, files are stored in the container itself, which could introduce various other issues.

This PR modifies the refined file download functionality so that instead of files being written to and stored temporarily within the running container, they are now written to S3 and the user is provided with a pre-signed download link that they can access within the Refiner UI in order to get the file. There should be no impact to UX.

I have added [localstack](https://github.com/localstack/localstack) to the project's `docker-compose.yaml` file, which provides us with a way to locally write files to S3 and download them from S3. While necessary for the application to function locally after this change, localstack will allow for simple integration testing and will be useful in some of our upcoming work related to configuration storage.

**This should close out at least a few of the security issues in the security tab as well!**

## 🔗 Related Issue

Fixes #221 

## 📺 Demo

Kapture 2025-07-31 at 11 19 45

## 🧪 How to test

Unit tests

4. Run the unit tests with `pytest tests/`

The most import tests to look at are the ones under `TestDemo` in `integration/test_demo.py`.

Testing the application:

1. Shut down any running containers and then run `docker compose up`
4. Run through the demo using the sample file or uploading your own
5. On the "success page" try downloading the refined files using the `Download results` button. It should work exactly as it did before.

## ℹ️ Additional Information

N/A

### DIFF
--- a/client/src/pages/Testing/Success.tsx
+++ b/client/src/pages/Testing/Success.tsx
@@ -23,7 +23,7 @@ export function Success({
     conditions[0]
   );
 
-  async function downloadFile(presignedUrl: string) {
+  function downloadFile(presignedUrl: string) {
     try {
       const link = document.createElement('a');
       link.href = presignedUrl;
@@ -77,7 +77,7 @@ export function Success({
           </div>
           <div>
             <div className="flex flex-col items-start gap-3">
-              <Button onClick={() => void downloadFile(presignedDownloadUrl)}>
+              <Button onClick={() => downloadFile(presignedDownloadUrl)}>
                 Download results
               </Button>
               {downloadError ? (


rogeruiz:
Should we update the main README with the necessary variables that were added to the `docker-compose.yaml`?

jakewheeler:
> Should we update the main README with the necessary variables that were added to the `docker-compose.yaml`?

Yes, definitely - thanks for noticing that!

refiner/app/services/aws/s3.py
```diff
@@ -0,0 +1,52 @@
...
+
+        presigned_url = s3_client.generate_presigned_url(
+            "get_object",
+            Params={"Bucket": bucket_name, "Key": key},
+            ExpiresIn=expires,
+        )
+
+        return presigned_url
+
+    except ClientError:
```
robertmitchellv, line 48: since we will end up doing more with S3 do you think it would be good to add custom models for the different kinds of exceptions? something like:

```python
class AWSError(Exception):
    """Base class for AWS-related errors"""
    def __init__(self, message: str, aws_error_code: str = None):
        self.message = message
        self.aws_error_code = aws_error_code
        super().__init__(self.message)

class S3Error(AWSError):
    """Base class for S3-related errors"""
    pass

class S3UploadError(S3Error):
    """Raised when an S3 upload operation fails"""
    pass

class S3DownloadError(S3Error):
    """Raised when an S3 download operation fails"""
    pass

class S3PermissionError(S3Error):
    """Raised when an S3 operation fails due to insufficient permissions"""
    pass

class S3BucketNotFoundError(S3Error):
    """Raised when a specified S3 bucket doesn't exist"""
    pass

class LambdaError(AWSError):
    """Base class for Lambda-related errors"""
    pass

class LambdaInvocationError(LambdaError):
    """Raised when a Lambda invocation fails"""
    pass
```

docker-compose.yaml
```diff
@@ -68,5 +73,22 @@ services:
...
+      - "4566:4566"
+    environment:
+      - SERVICES=s3
+      - DEBUG=1
+      - AWS_ACCESS_KEY_ID=refiner
+      - AWS_SECRET_ACCESS_KEY=refiner
+    volumes:
+      - localstack-data:/var/lib/localstack
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./localstack/s3-init.sh:/etc/localstack/init/ready.d/s3-init.sh
```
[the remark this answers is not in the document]
robertmitchellv: this is perfect since i really want to also add a bucket to test out writing our configs 🚀

refiner/app/services/aws/s3.py
robertmitchellv: yeah, the way you're describing how i might use them is correct--i'm just anticipating the logging and how we want to approach things. i was thinking about the way we've been approaching the raising of errors so far and wondering if we should continue in that vein or do something else? we're sort of maintaining a middle layer abstraction of what the error means for the app outside of what it means as the error itself. so maybe this is a larger question for eng sync. so this isn't anything that should hold us back for now.

PR #227, opened 2025-07-28
feat(ui): Add Configuration

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

These changes are for adding Configurations to the `<ConfigurationsTable />`
component with a modal and button.

<details>

<summary>24 commits later ... (click to view them)</summary>

1. 💬 feat(ui): Add Configurations modal flow 🏷️ a7b8fdf
1. 💬 feat(ui): Setup new condition flow 🏷️ e382f71
1. 💬 feat(ui): Refactor tests for Configurations Page 🏷️ 36a3c95
1. 💬 chore(ui): Lint & format files 🏷️ 63ef52e
1. 💬 test(ui): Skipping tests and bringing in Mocks 🏷️ 0685469
1. 💬 chore: Lint & format files 🏷️ 66ade1e
1. 💬 Merge branch 'main' into rogeruiz/165-add-configuration-ui 🏷️ 590a314
1. 💬 Apply suggestions from code review; 🏷️ 6126563
1. 💬 chore(ui): Remove mocks for Vitest & clean up 🏷️ 9dc7170
1. 💬 fix(test): Bring in a mock for Tabbable & Vitest 🏷️ 7a1b179
1. 💬 fix(test): Moves mock to setup, & adds more tests 🏷️ 4d796a0
1. 💬 refactor(ui): Incorporate styles into `<Button />` 🏷️ a2909fa
1. 💬 Merge branch 'main' into rogeruiz/165-add-configuration-ui 🏷️ b3a2224
1. 💬 Merge branch 'main' into rogeruiz/165-add-configuration-ui 🏷️ f897b5f
1. 💬 test(ui): Complete tests for Configuration Modal 🏷️ ea53bd1
1. 💬 chore: Lint & format files 🏷️ 306d8f4
1. 💬 fix(test): Update tests properly; 🏷️ e5a3201
1. 💬 chore: Resolving conflicts & updating from `main`; 🏷️ fa5e0bd
1. 💬 fix(test): Refactor the test to pass for now; 🏷️ ac3f408
    - 📝 I'll circle back in a smaller PR to re-implement this test and
      functionality a bit more.
1. 💬 fix(ui): Remove unused styles; 🏷️ c4af827
1. 💬 fix: Working through feedback; 🏷️ deb4264
1. 💬 fix: Remove redundant tests; 🏷️ c460eeb
1. 💬 Apply suggestions from code review; 🏷️ 1a4b5a5
[...]

PR #248, opened 2025-07-31
feat: Add toast component

jakewheeler:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #
- #

### DIFF
--- a/client/src/pages/Configurations/index.tsx
+++ b/client/src/pages/Configurations/index.tsx
@@ -2,6 +2,7 @@ import { Title } from '../../components/Title';
 import { Button } from '../../components/Button';
 import { Search } from '../../components/Search';
 import { ConfigurationsTable } from '../../components/ConfigurationsTable';
+import { toast } from 'react-toastify';
 
 enum ConfigurationStatus {
   on = 'on',
@@ -42,6 +43,19 @@ export function Configurations() {
 
   return (
     <section className="mx-auto p-3">
+      <Button
+        onClick={() =>
+          toast(<p>Hi</p>, {
+            autoClose: 3000,
+            position: 'bottom-left',
+            closeOnClick: true,
+            closeButton: false,
+            pauseOnFocusLoss: false,
+          })
+        }
+      >
+        Show toast
+      </Button>
       <div className="flex flex-col gap-4 py-10">
         <Title>Your reportable condition configurations</Title>
         <p>
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -16,6 +16,7 @@
         "react": "^19.1.1",
         "react-dom": "^19.1.1",
         "react-router": "^7.5.2",
+        "react-toastify": "^11.0.5",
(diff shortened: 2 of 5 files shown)


client/src/pages/Configurations/index.tsx
```diff
@@ -56,7 +59,17 @@ export function Configurations() {
...
+          className="m-0!"
+          onClick={() =>
+            showToast({
+              heading: 'New configuration created',
+              body: 'Human immunodeficiency virus infection',
+            })
+          }
+        >
+          Set up new condition
+        </Button>
```
jakewheeler, line 72: This will get updated once the other pages are in place. Just using this as an example for now.

client/src/pages/Configurations/index.test.tsx
```diff
@@ -30,4 +33,13 @@ describe('Configurations', () => {
...
   });
+
+  it('should render an error and success toast when the "Set up new configuration" button is clicked', async () => {
+    const user = userEvent.setup();
+    renderPageView();
+    await user.click(screen.getByText('Set up new condition'));
+    expect(
+      await screen.findAllByText('New configuration created')
+    ).toHaveLength(2);
+  });
```
rogeruiz, line 44: Is this test always going to be accurate? Should we test the component in a different way? I sorta like this as an example of how to test, but worry it's too specific to this PR and not how the app will ultimately function.

client/src/pages/Configurations/index.tsx
```diff
@@ -56,7 +59,22 @@ export function Configurations() {
...
+            showToast({
+              heading: 'New configuration created',
+              body: 'Human immunodeficiency virus infection',
+            });
+            showToast({
+              heading: 'New configuration created',
+              body: 'Human immunodeficiency virus infection',
+              variant: 'error',
+            });
+          }}
```
rogeruiz, line 74: Same as the other comment. but not a show stopper for now. I think I can use the Toast for the work introduced in #227 after it all gets integrated in a new small PR. What do you think?

rogeruiz (approved):
lgtm 🌈

Looks good. I left some comments about how it's currently coded, but I'm okay with introducing these small changes for now to get them into the default branch to wire things up with later PRs.

client/src/pages/Configurations/index.test.tsx
jakewheeler: You're 100% correct. This test likely won't stick around for long since we'll (hopefully very soon) be using these toasts in a few places around the app to demonstrate something happened. I mainly just wanted a test to show the toasts work in the meantime 😄

client/src/pages/Configurations/index.tsx
jakewheeler: Yep, I think that's a good plan! There's no real need for this code other than as an example to show how it's used. Once you integrate it into your work, it would make sense to me to delete this code and probably delete (or modify) the test I wrote to make more sense with your integration.

PR #250, opened 2025-08-01
refactor: remove refinement cache table, triggers, and functions from data model

### LINKED ISSUE #249: [TASK] Revist the data model give new constraints
## 🚧 What needs to be done

Design a data model (and supporting SQL queries) that enables, given a SNOMED condition code, the retrieval of all codes grouped under its corresponding condition grouper—either all codes at once or filtered by context (e.g., only diagnosis, immunizations, medications, etc.).

## 🤔 Why it needs to be done

We are moving away from a cache-table and trigger-based approach due to new architectural constraints (e.g., air-gapped Lambda services). The new model needs to support on-demand, context-based code retrieval for a given SNOMED condition code.

## 🧩 Background & Source Relationships

- At the source (APHL's TES), a **condition grouper** is a parent that references:
  - **Reporting specification groupers** (children, each with a 1:1 relationship to a SNOMED condition code; each grouper groups codeable concepts by code system, e.g., SNOMED, LOINC, ICD10, RxNorm)
  - **Additional context groupers** (children holding codes for specific contexts, such as diagnosis, immunizations, medications, etc.)
- Condition groupers themselves do not hold codes, but only reference their reporting specification and additional context groupers.
- For example, a COVID condition grouper will point to reporting specification groupers for COVID SNOMED codes and the codes for different contexts.

## 🎯 Desired Capability

- Given a SNOMED condition code, be able to query for:
  - All codes that fall under the related condition grouper (across all child groupers)
  - Codes filtered by specific context (e.g., `WHERE context = 'diagnosis'`)

## 🏗️ Acceptance Criteria

- [ ] Data model supports these relationships and query patterns
- [ ] Pure SQL queries provided for:
  - Retrieving all codes for a given SNOMED condition code
  - Retrieving codes for a given SNOMED condition code filtered by context
- [ ] Solution avoids triggers and cache tables

## ⚙️ Technical Notes

- The solution should avoid complexity; queries should be as simple as possible
- Data will be sourced from flat files (FHIR ValueSets) and imported
- Architecture requires "air-gapped" access (e.g., AWS Lambda cannot read from public DB)
- Prefer database-agnostic SQL if possible, or specify RDBMS requirements

## 📜 Additional context

- Architectural details (Lambda, air gap) do not affect the data model itself, but motivate the move away from cache tables/triggers.
- Focus is on enabling flexible, context-based reporting and analytics queries.

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This PR refactors the database model to support on-demand, context-based retrieval of codes for a given SNOMED condition code, in alignment with new architectural constraints (e.g., air-gapped Lambda services). It removes cache tables and triggers, aggregates codes by condition grouper, and introduces a flexible schema for jurisdictional configuration and activation. Support for "additional context groupers" is deferred until TES 2.0.0 Additional Context Grouper ValueSets have full implementation.

## 🔗 Related Issue

Fixes #249 
- #249 

## ✅ Acceptance Criteria

- [x] Data model supports direct relationships between SNOMED condition codes and grouped codes (LOINC, SNOMED, ICD10, RxNorm)
- [x] Pure SQL queries provided for:
    - Retrieving all codes for a given SNOMED condition code from the `conditions` table
    - Retrieving codes for a given SNOMED condition based on its "active" configuration
- [x] No triggers or cache tables used
- [x] Data is sourced from flat files (FHIR ValueSets)
- [x] Model and queries are extensible for future context groupers

#### 1. Follow the README instructions to set up and seed the database:

- Start the database:

```bash
docker compose build --no-cache
docker compose up -d
```

- Seeding the database is done automatically when you start the container.

#### 1. Find a Condition by a Child SNOMED Code

Since conditions in the `conditions` table have an array of `child_rsg_snomed_codes` we can very easily start with our input, which will be SNOMED condition codes from the RR's coded information organizer (either one or many as a list) and we can retrieve the parent condition grouper's complete list of ValueSets as `jsonb` with both the code and display name by code system:

```sql
-- Find a specific version of a condition by a child SNOMED code.
-- This uses the && operator (overlap) to efficiently search inside the text array.

SELECT
  display_name,
  version,
  canonical_url,
  child_rsg_snomed_codes,
  -- The jsonb_pretty function formats the JSON output for better readability.
  jsonb_pretty(loinc_codes) as loinc_codes,
  jsonb_pretty(snomed_codes) as snomed_codes,
  jsonb_pretty(icd10_codes) as icd10_codes,
  jsonb_pretty(rxnorm_codes) as rxnorm_codes
FROM
  conditions
WHERE
  -- The '&&' operator checks if the child_rsg_snomed_codes array
  -- has any elements in common with the array provided.
  -- This is a very fast way to check for membership.
  child_rsg_snomed_codes && ARRAY['840539006', '772828001']
  -- You can optionally filter to a specific version of the condition definition.
  AND version = '2.0.0';
```

#### 2. View All Configurations with Their Details

This query shows all configurations with their included conditions and any jurisdiction-specific additions:

```sql
-- Get an overview of all configurations
-- Shows which conditions each configuration includes and any custom additions

SELECT
    c.id as configuration_id,
    c.name,
    c.description,
    c.jurisdiction_id,
    c.created_at,
    
    -- Show which conditions are included
    jsonb_pretty(c.included_conditions) as included_conditions,
    
    -- Show jurisdiction-specific additions (only if they exist)
    CASE 
        WHEN jsonb_array_length(c.loinc_codes_additions) > 0 
        THEN jsonb_pretty(c.loinc_codes_additions) 
        ELSE NULL 
    END as loinc_additions,
    
    CASE 
        WHEN jsonb_array_length(c.snomed_codes_additions) > 0 
        THEN jsonb_pretty(c.snomed_codes_additions) 
        ELSE NULL 
    END as snomed_additions,
    
    CASE 
        WHEN jsonb_array_length(c.icd10_codes_additions) > 0 
        THEN jsonb_pretty(c.icd10_codes_additions) 
        ELSE NULL 
    END as icd10_additions,
    
    CASE 
        WHEN jsonb_array_length(c.rxnorm_codes_additions) > 0 
        THEN jsonb_pretty(c.rxnorm_codes_additions) 
        ELSE NULL 
    END as rxnorm_additions

FROM configurations c
ORDER BY c.jurisdiction_id, c.created_at DESC;
```

#### 3. View Active and Historical Activations

This query shows the activation status for each jurisdiction and SNOMED code combination, including historical activations:

```sql
-- Get activation status overview
-- Shows current and historical activations with their lifecycle

SELECT
    a.jurisdiction_id,
    a.snomed_code,
    a.configuration_id,
    c.name as configuration_name,
    a.activated_at,
    a.deactivated_at,
    
    -- Show activation status
    CASE 
        WHEN a.deactivated_at IS NULL THEN 'ACTIVE'
        ELSE 'DEACTIVATED'
    END as status,
    
    -- Show how long the activation was/has been active
    CASE 
        WHEN a.deactivated_at IS NULL 
        THEN EXTRACT(EPOCH FROM (NOW() - a.activated_at)) / 86400 || ' days (ongoing)'
        ELSE EXTRACT(EPOCH FROM (a.deactivated_at - a.activated_at)) / 86400 || ' days'
    END as duration,
    
    -- Show a preview of the computed codes structure
    jsonb_pretty(
        jsonb_build_object(
            'loinc_count', jsonb_array_length(a.computed_codes->'loinc_codes'),
            'snomed_count', jsonb_array_length(a.computed_codes->'snomed_codes'),
            'icd10_count', jsonb_array_length(a.computed_codes->'icd10_codes'),
            'rxnorm_count', jsonb_array_length(a.computed_codes->'rxnorm_codes')
        )
    ) as code_counts

FROM activations a
JOIN configurations c ON a.configuration_id = c.id
ORDER BY 
    a.jurisdiction_id, 
    a.snomed_code, 
    a.activated_at DESC;
```

#### 4. Primary Runtime Query - Get Active Configuration for SNOMED Code

This is the primary query the Refiner application will use at runtime. It's optimized for fast lookups of currently active configurations:

```sql
-- Fast runtime lookup for active configurations
-- This is the primary query pattern for the Refiner application

SELECT
    a.jurisdiction_id,
    a.snomed_code,
    c.name as configuration_name,
    c.id as configuration_id,
    a.computed_codes
FROM activations a
JOIN configurations c ON a.configuration_id = c.id
WHERE 
    a.jurisdiction_id = 'SDDH'
    AND a.snomed_code = '840539006'
    AND a.deactivated_at IS NULL  -- Only get currently active
LIMIT 1;
```

This query is quite complex so I think a good discussion point would be whether or not we decide to stick with one query for this or if we'd rather do two queries and then handle the more complicated parts in python, which could make things a lot simpler. And it's important to note that we'll still need to implement the packaging of this object that combines the `conditions` table data with the `configuration_version` data for the AWS Lambda piece too. I welcome feedback on this 🙇🏻


database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,121 @@
...
-CREATE TABLE IF NOT EXISTS tes_condition_grouper_references (
+-- a type to represent the lifecycle of a configuration version
+-- 'draft': in progress, not ready for use
+-- 'active': vetted, approved, and will be used by the activation logic
+-- 'archived': no longer in use, kept for historical purposes
+CREATE TYPE configuration_status AS ENUM ('draft', 'active', 'archived');
+
+-- this table stores each specific, immutable version of a configuration
+-- every time a user saves a change, a new row is created here
+CREATE TABLE configuration_versions (
```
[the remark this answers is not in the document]
robertmitchellv: i'm still on the fence because of this reason:

- **Configuration ID** `=` Issue/PR number (immutable, permanent identifier)
- **Configuration name** `=` Issue/PR title (can be renamed anytime)
- **Configuration versions** `=` Individual commits/updates to that issue/PR
- **Labels** `=` Tags that can be added/removed without creating new versions

we want to be able to easily go from a configuration 💡 to all of its versions while allowing a user to change its name at any point.

but, this doesn't necessarily need to be implemented right away but a structure like this makes it a bit easier. and if the json payload is in the `activations` table, this isn't as important of a change since it won't affect the main query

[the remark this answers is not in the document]
robertmitchellv: this is where i landed:

```sql
-- simplified configurations table - version-centric approach
-- * each row represents a specific, immutable configuration version (like github issues/PRs)
-- * configuration ID becomes the user-facing version number (e.g., "configuration #47")
-- * benefits:
--   - eliminates complex joins
--   - simpler activation lookups
--   - more intuitive user experience ("work with config #47")
--   - matches github's issue numbering paradigm
CREATE TABLE configurations (
    id SERIAL PRIMARY KEY,                    -- This IS the user-facing version number
    jurisdiction_id TEXT NOT NULL REFERENCES jurisdictions(id),
    name TEXT NOT NULL,                       -- Can be renamed anytime without versioning
    description TEXT,

    -- configuration data: references to base conditions by composite key
    -- must reference at least one condition to ensure configurations have meaningful clinical context
    included_conditions JSONB NOT NULL DEFAULT '[]'::jsonb,
    -- structure: [{"canonical_url": "...", "version": "..."}, ...]

    -- jurisdiction-specific code additions (beyond what's in the base conditions)
    loinc_codes_additions JSONB DEFAULT '[]'::jsonb,
    snomed_codes_additions JSONB DEFAULT '[]'::jsonb,
    icd10_codes_additions JSONB DEFAULT '[]'::jsonb,
    rxnorm_codes_additions JSONB DEFAULT '[]'::jsonb,

    -- optional: track relationships between configurations for audit/history
    -- e.g., "configuration #48 was cloned from #47"
    derived_from_id INTEGER REFERENCES configurations(id),

    created_at TIMESTAMPTZ DEFAULT NOW(),

    -- business rule: configurations must reference at least one condition
    -- this ensures every configuration has a meaningful clinical foundation
    CONSTRAINT configurations_must_have_conditions
    CHECK (jsonb_array_length(included_conditions) > 0)
);
```

database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,143 @@
...
+    deactivated_at TIMESTAMPTZ NULL,
+
+    -- pre-computed aggregated codes (replaces complex runtime query)
+    -- structure matches ProcessedGrouper for easy consumption by terminology service
+    computed_codes JSONB NOT NULL,
+    -- example: {"loinc_codes": [...], "snomed_codes": [...], "icd10_codes": [...], "rxnorm_codes": [...]}
+
+    -- S3 synchronization tracking for Lambda-based horizontal scaling
+    s3_synced_at TIMESTAMPTZ NULL,
+    s3_object_key TEXT NULL,
```
jakewheeler, line 109: This is a good idea 👍

database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,143 @@
...
+--   - eliminates complex joins
+--   - simpler activation lookups
+--   - more intuitive user experience ("work with config #47")
+--   - matches github's issue numbering paradigm
+CREATE TABLE configurations (
+    id SERIAL PRIMARY KEY,                    -- This IS the user-facing version number
+    jurisdiction_id TEXT NOT NULL REFERENCES jurisdictions(id),
+    name TEXT NOT NULL,                       -- Can be renamed anytime without versioning
+    description TEXT,
+
```
jakewheeler, line 60: Do we want to capture section code info now or hold off? I know we discussed this but I can't recall where we landed.

database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,143 @@
...
+-- * this table stores the "activation state" of configurations for specific SNOMED codes
+-- * key optimization: computed_codes contains pre-aggregated data matching ProcessedGrouper structure
+-- * benefits:
+--   - runtime queries become simple lookups instead of complex aggregations
+--   - predictable performance regardless of configuration complexity
+--   - easy to sync with S3 for horizontal scaling via Lambda
+--   - full audit trail of activation history
+CREATE TABLE activations (
+    jurisdiction_id TEXT NOT NULL REFERENCES jurisdictions(id),
+    snomed_code TEXT NOT NULL,                -- From conditions.child_rsg_snomed_codes
```
jakewheeler, line 94: In our seed data, why does code `840539006` go here? Why is `186747009` not "activated" even though it's part of the COVID19 grouper?

database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,143 @@
...
+-- * configuration ID becomes the user-facing version number (e.g., "configuration #47")
+-- * benefits:
+--   - eliminates complex joins
+--   - simpler activation lookups
+--   - more intuitive user experience ("work with config #47")
+--   - matches github's issue numbering paradigm
+CREATE TABLE configurations (
+    id SERIAL PRIMARY KEY,                    -- This IS the user-facing version number
+    jurisdiction_id TEXT NOT NULL REFERENCES jurisdictions(id),
+    name TEXT NOT NULL,                       -- Can be renamed anytime without versioning
```
jakewheeler, line 58: If this can be renamed, should we still have an `updated_at` column?

database/schema/100-base-schema.sql
```diff
@@ -1,88 +1,143 @@
...
+-- simplified configurations table - version-centric approach
+-- * each row represents a specific, immutable configuration version (like github issues/PRs)
+-- * configuration ID becomes the user-facing version number (e.g., "configuration #47")
+-- * benefits:
+--   - eliminates complex joins
+--   - simpler activation lookups
+--   - more intuitive user experience ("work with config #47")
+--   - matches github's issue numbering paradigm
+CREATE TABLE configurations (
+    id SERIAL PRIMARY KEY,                    -- This IS the user-facing version number
```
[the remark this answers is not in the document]
robertmitchellv: ah right, that's a good catch, that's not what i was intending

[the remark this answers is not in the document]
robertmitchellv: i think you're right about the uuids. i'm going to add that. i was going a bit too fast here and was going to use another composite key but maybe we can just use the uuids instead of these more complex composite keys? i can always add validation to enforce some of the things the composite keys were doing while leaving the pk up to the uuid algo

database/schema/100-base-schema.sql
robertmitchellv: yeah i think that would be a good addition; it's still helpful i think

database/schema/100-base-schema.sql
robertmitchellv: yeah, i think i'll just use the same column name as the parameter name to the refine function just to make it extra clear

database/schema/100-base-schema.sql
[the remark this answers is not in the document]
jakewheeler: Yes that sounds perfect 👍

database/schema/100-base-schema.sql
robertmitchellv: i hadn't really thought too deeply about it tbh. but when a user decides to activate a configuration i'm expecting:

- the server will begin the process of greating that json object that looks like `ProcessedGrouper` to write to the `activations` table and it will also attempt to write it to s3. once it finishes, we'll want to check that both are in sync and want to document when things are synced probably. if all the checks pass then we'll tell the user it was successful. if there was an error it would tell them something went wrong. so at least there will either be a previously well tested config in prod or no config in prod if there's a problem.

i feel like a kind of popup with folder sending packets to a globe type thing would work

jakewheeler: This aligns with what I was thinking too. If any step in this operation fails the entire thing will fail.

database/schema/100-base-schema.sql
jakewheeler: As we discussed, we can probably revisit this.

PR #251, opened 2025-08-04
SPIKE: automatically generate React hooks for data fetching

### LINKED ISSUE #256: [SPIKE] Automatically generate React data fetching hooks from FastAPI routes
## 🚧 What needs to be done

Refiner team engineers currently need to manually configure client-side requests within the React app each time a new FastAPI route is created. This can be time-consuming and error prone since the TypeScript types need to manually stay in sync with the data returned by the FastAPI route.

We should investigate how we can automatically generate React hooks that can be used to fetch or mutate data coming from the server.

## 🤔 Why it needs to be done

This will save engineers time on a regular basis as we build out new client-facing features. This will also help to prevent typing errors.

## ⏰ Timebox

~3 days

## 📝 To-do list

- [ ] Research and choose a data fetching tool
- [ ] Find a codegen tool compatible with the data fetching tool
- [ ] Configure the codegen tool to consume FastAPI's OpenAPI spec and generate types
- [ ] Create scripts to handle the codegen
- [ ] Configure development environment to watch for file changes and re-run codegen
- [ ] Provide proof-of-concept examples to show that the method is viable
  - [ ] Demonstrate error and loading states
- [ ] Explore mocking the data fetching hooks for testing

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The goal of this PR is to research and understand how we might go about automatically generating React hooks in order to fetch data from the API and use it in the client. There were a couple of main motivators for this:

1. Every time we add a new feature to the FastAPI we will also need to write the client code to make use of it
2. We want the interface types between the client and server to always stay in sync

Doing these couple of things manually will take a lot of extra development effort. It is also error-prone.

### What this PR does

Please refer to the updated `client/README.md` file in this PR for the full overview, but in short, this PR sets up everything needed to automatically update or create React hooks when FastAPI routes are modified on the server.

For example, if I create a route to fetch all configurations for a given jurisdiction, I will receive a hook `useGetConditions` on the client side I can simply use to fetch data from that new endpoint.

(example)
```tsx
const { data, isLoading, isError } = useGetConfigurations() 
```

Similarly, if I create a route that allows me to create a new configuration, I'll receive a hook that provides me with a function to mutate that data on the server.

(example)
```tsx
const { mutate } = useCreateConfiguration() 
```

## 🔗 Related Issue

Fixes #256 

## 🧪 How to test

Unit tests

4. Activate a virtual environment using your tool of choice
5. Make sure that both `requirements.txt` and `requirements-dev.txt` are installed
6. Run the unit tests with `pytest -vv tests/`
7. Additionally you can check coverage with `pytest --cov=app tests`

Testing the application:

1. Stop all running containers and then run `docker compose up`
2. You'll see a message like this from the client container:
```
client-1           | [0] 🍻 Start orval v7.11.2 - A swagger client generator for typescript
client-1           | [0] Running afterAllFilesWrite hook...
client-1           | [0] 🎉 fastapi - Your OpenAPI spec has been converted into ready to use orval!
```
3. In the client, take a look at `pages/Configurations/index.ts` - this is a good example to show how the data is being fetched to give to the `ConfigurationsTable` component.


.pre-commit-config.yaml
```diff
@@ -20,8 +20,10 @@ repos:
         -   --unsafe
     # - ensure files end with newline
     -   id: end-of-file-fixer
+        exclude: ^refiner/openapi\.json$
     # - remove trailing whitespace
     -   id: trailing-whitespace
+        exclude: ^refiner/openapi\.json$
```
jakewheeler, line 26: This is a generated file so we want it left alone.

client/src/pages/Configurations/index.tsx
```diff
@@ -2,46 +2,14 @@ import { Title } from '../../components/Title';
...
-        id: 'syphilis-config-id',
-      },
-      {
-        name: 'Viral hepatitis, type A',
-        status: ConfigurationStatus.on,
-        id: 'viral-hep-config-id',
-      },
-    ],
-  };
+  if (isLoading || !data?.data) return 'Loading...';
```
jakewheeler, line 12: This `data.data` situation is a little odd. Typically when you use TanStack Query on its own, the fetched data from the server goes into the `data` variable. However, due to the way Orval generates these hooks, the original `data` is actually an `AxiosResponse`. This response has a few different properties, such as `status`, `headers`, etc. It also includes its own `data`, which is the data coming from the server that we're attempting to fetch.

I don't think this is necessarily a bad thing. Having access to the full response could potentially be helpful in certain situations. It looks a little weird but we'll just need to keep in mind that `data.data` is what we actually want most of the time.

client/src/pages/Testing/Success.tsx
```diff
@@ -2,20 +2,21 @@ import { useState } from 'react';
...
-  conditions: Condition[];
-  unrefinedEicr: string;
-  downloadToken: string;
-}
+type SuccessProps = Pick<
+  RefinedTestingDocument,
+  'conditions' | 'refined_download_token' | 'unrefined_eicr'
+>;
+
+type Condition = SuccessProps['conditions'][0];
```
jakewheeler, line 14: This is a good example of no longer needing to try to keep backend types in line with frontend types. The `RefinedTestingDocument` is produced by Orval and is based off of the `RefinedTestingDocument` model defined in our FastAPI.

client/src/pages/Testing/index.tsx
```diff
@@ -33,43 +28,29 @@ export default function Demo() {
...
-    }
-  }
-
   function reset() {
     setView('run-test');
-    setUploadResponse(null);
+    resetState();
   }
 
+  if (isPending) return 'Loading...';
```
jakewheeler, line 52: I've been using these loading placeholders but we will swap these out for some sort of spinner or something when one is selected.

refiner/app/api/v1/configurations.py
```diff
@@ -0,0 +1,30 @@
+from fastapi import APIRouter
+
+from ...db.configurations.model import Configuration
+
+router = APIRouter(prefix="/configurations")
+
+
+@router.get(
```
jakewheeler, line 8: I used this as an example of grabbing a user's configs from the server and displaying them in the client via the generated `useGetConfigurations` hook.

This route will, of course, be updated in the future to pull from the database instead of using hard-coded values.

refiner/tests/integration/test_configurations.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+base_route = "/api/v1/configurations/"
+
+
+@pytest.mark.integration
+@pytest.mark.asyncio
+async def test_get_configurations_returns_expected_data(authed_client):
```
jakewheeler, line 8: This test can be updated to use the database when that time comes.

client/src/pages/Configurations/index.tsx
rogeruiz: Can we have a better convention around naming the variables here? I suggest `response` so that it looks like `response.data`.

rogeruiz: We should probably update the variable to something like `response` instead of `data` so that it can be referenced like `!response?.data`. That feels like it would align best with the default behavior of `AxiosReponse`. What do you think?

refiner/openapi.json
```diff
@@ -0,0 +1 @@
...Conditions Found"}, "conditions": {"items": {"$ref": "#/components/schemas/Condition"}, "type": "array", "title": "Conditions"}, "unrefined_eicr": {"type": "string", "title": "Unrefined Eicr"}, "processing_notes": {"items": {"type": "string"}, "type": "array", "title": "Processing Notes"}, "refined_download_token": {"type": "string", "title": "Refined Download Token"}}, "type": "object", "required": ["message", "conditions_found", "conditions", "unrefined_eicr", "processing_notes", "refined_download_token"], "title": "RefinedTestingDocument", "description": "Model for the response when uploading a document in the testing suite."}, "User": {"properties": {"id": {"type": "string", "title": "Id"}, "username": {"type": "string", "title": "Username"}}, "type": "object", "required": ["id", "username"], "title": "User", "description": "Model for a logged-in user."}, "ValidationError": {"properties": {"loc": {"items": {"anyOf": [{"type": "string"}, {"type": "integer"}]}, "type": "array", "title": "Location"}, "msg": {"type": "string", "title": "Message"}, "type": {"type": "string", "title": "Error Type"}}, "type": "object", "required": ["loc", "msg", "type"], "title": "ValidationError"}}}}
```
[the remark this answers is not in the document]
rogeruiz: To avoid Diffs GitHub and locally, we should add all of our generated files to a `.gitattributes` file where we mark them as `-diff` to avoid showing them whenever `git diff` is run or when PRs get submitted. I believe GitHub will avoid the diff-view for files marked this way. [More info here ⬅️](https://git-scm.com/docs/gitattributes#_marking_files_as_binary)

client/src/pages/Configurations/index.tsx
jakewheeler: Yeah, I think that makes sense. Unfortunately I don't think Orval allows us to change the name during generation. However, we could generally plan to use the hooks like this:
```tsx
const { data: response, isLoading } = useGetConfigurations();
```

So that we can use it like you mentioned. Do you think that works? If so, I can make this update.

refiner/openapi.json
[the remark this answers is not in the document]
jakewheeler: think I would prefer to still see a diff just so that we can all easily tell when a generated file is being created/updated. I think if we all know that there's no need to review the content of the files too closely that's probably enough. What do you think? I can check in with the team as well.

PR #258, opened 2025-08-04
Nuel/pr image scan

EmmanuelNwa247:
# Changes Proposed 
## "Fix: Update existing security scan comments instead of creating new ones
This update improves the PR security scanning workflow to reduce comment spam. The current workflow successfully scans both Lambda and Refiner container images on every pull request, providing detailed vulnerability reports. However, it creates a new comment for each commit pushed to the same PR branch, resulting in multiple redundant comments that clutter the PR interface.
This fix modifies the workflow to update the existing security scan comment instead of creating new ones, maintaining the same security coverage while providing a cleaner experience.

- Prevents comment spam on PRs.
- Updates existing comments instead of creating new ones.
- Adds timestamp for last updated info.


.github/workflows/pr-security-scan.yml
```diff
@@ -97,12 +99,37 @@ jobs:
             message += `**Scan Time**: ${new Date().toISOString()}\n`;
             message += `**Status**: ${conclusion === 'success' ? 'Completed' : 'Failed'}\n\n`;
             message += `**View detailed results**: Navigate to Security tab → Code scanning alerts\n`;
-            message += `**Filter by**: \`pr-scan-refiner-app\` or \`pr-scan-refiner-lambda\``;
+            message += `**Filter by**: \`pr-scan-refiner-app\` or \`pr-scan-refiner-lambda\`\n\n`;
+            message += `*Last updated: ${new Date().toLocaleString()}*`;
```
jakewheeler, line 103: I'm EST and this is showing a few hours into the future for me (~11:40pm instead of ~7:40pm). Should we use UTC so it's the same for everybody? `new Date().toUTCString()`

EmmanuelNwa247: Great Catch, Jake. Yes, we can use UTC.

EmmanuelNwa247: @jakewheeler, I have gone ahead to update the workflow file to output UTC time zone

PR #262, opened 2025-08-07
chore: Implement API logging

### LINKED ISSUE #93: [TASK] Implement API logging
## 🚧 What needs to be done

While the server does produce some of its own logging, such as when HTTP requests are made or when errors occur, we do not currently have a logging implementation that allows developers to easily add their own logs.

We should create a logging setup that includes:

- A standard format with important metadata
- The ability for developers to easily use the logger within any API route
- The ability to produce various category of logs (`INFO`, `WARNING`, `ERROR`, etc.)
- Logging to `stdout` 

Please note that this work is only intended to cover the API. It is likely we will need to include logging within the eventual Lambda function as well, but we should save that for a different ticket.

## 🤔 Why it needs to be done

- Improve ability to diagnose problems
- Track server usage history
- Understand what was happening when

## 🏗️ Acceptance Criteria

- [ ] A standard format with important metadata has been configured
- [ ] Developers can make use of the logger in any request handler
- [ ] Developers are able to specify what type of error to log
- [ ] Logs go to `stdout`
- [ ] Initial logs added where it makes sense

## ⚙️ Technical Notes

We should look research to figure out what the recommend approach to logging is in FastAPI. It's possible we may need to create a logging middleware or something similar.

[structlog](https://www.structlog.org/en/stable/getting-started.html) could be a good place to start.

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR configures a logger with standard fields that can be used throughout the FastAPI application. I've configured the logger to log every request to the API. The request logs produced contain traceable fields, such as a `request_id` and a `timestamp`.

I've made use of [python-json-logger](https://nhairs.github.io/python-json-logger/latest/) so that all logs are formatted as JSON. This allow strong compatibility with external logging solutions, such as Splunk or AWS CloudWatch. Logs will be able to be easily searched by their fields.

Here's an example:

```sh
refiner-service-1  | {"app": "refiner", "env": "local", "message": "Request start", "request_id": "7bf2572c-1134-4b9d-986c-711d75572879", "path": "/api/v1/demo/upload", "method": "POST", "timestamp": "2025-08-08T15:25:35.529228+00:00"}
refiner-service-1  | {"app": "refiner", "env": "local", "message": "Processing demo file", "upload_file": "mon-mothma-two-conditions.zip", "request_id": "7bf2572c-1134-4b9d-986c-711d75572879"}
refiner-service-1  | {"app": "refiner", "env": "local", "message": "Request end", "request_id": "7bf2572c-1134-4b9d-986c-711d75572879", "path": "/api/v1/demo/upload", "status_code": 200, "method": "POST", "timestamp": "2025-08-08T15:25:35.695945+00:00", "duration_ms": 166.99}
```

This example shows the initial request, something happening in a route (running the demo file), and the request ending. The `request_id` will allow us to trace all actions that occurred during a single request.

The logger can be accessed via any request handler by simply injecting it as a dependency. Here's an example of how you'd do that:

```python
@router.get('/example')
async def example(logger: Logger = Depends(get_logger)):
    logger.info('Log something here!')
    ...
```

> [!NOTE]
> All logs include the `request_id` by default if one exists. Otherwise it'll log as `"request_id": "unknown"`. No need to remember to log it yourself!

## 🔗 Related Issue

Fixes #93 

## 🧪 How to test

1. Restart your containers (to install `python-json-logger`)
2. Log in and click around the app so that requests are sent to the server
3. Take a look at the output from `refiner-service-1` - you'll see logs being produced

### DIFF
--- a/refiner/app/services/ecr/refine.py
+++ b/refiner/app/services/ecr/refine.py
@@ -1,4 +1,3 @@
-import logging
 from dataclasses import dataclass
 from typing import Literal, TypedDict, cast
 
@@ -24,8 +23,6 @@ from .grouper_utils import (
 # NOTE:
 # CONSTANTS AND CONFIGURATION
 # =============================================================================
-
-log = logging.getLogger(__name__).error
 
 # read json that contains details for refining and is the base of what drives `refine`
 REFINER_DETAILS = read_json_asset("refiner_details.json")


rogeruiz (approved):
lgtm 🌈 

Yay for logging!

PR #298, opened 2025-08-22
fix: use correct auth callback URI when user is logging in (demo site)

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The refiner web app runs behind a load balancer that uses TLS and TLS is terminated at the gateway, meaning the app itself does not make use of TLS. However, one issue we've discovered is that when `request.url_for("auth_callback")` is called, the app inserts `http://` instead of `https://` into the URI (since the server is not using HTTPS). This was causing an issue because our auth service is using HTTPS, not HTTP, so the URI was incorrect. This was preventing us from being able to log in correctly.

This PR implements a simple fix to swap `http` with `https` in a non-local environment. This fix is making the assumption the auth service we use will always be HTTPS.

This fix should be deployed to the demo environment for testing.


EmmanuelNwa247 (approved):
LGTM

PR #315, opened 2025-09-02
(Backend) Add/update/delete custom codes

### LINKED ISSUE #310: [TASK] (Backend) CRUD operations for custom codes
## 🚧 What needs to be done

The frontend work for custom codes is being completed in #151.

In this ticket we need to modify the backend so that users can:
- Create custom codes
- Read custom codes
- Update custom codes
- Delete custom codes

These actions must be hooked up to their corresponding components in the client as a part of this ticket. 

The `/configurations/:id` route will need to be updated to ensure custom codes are part of the initial response when a user navigates to the config builder page.

## 🤔 Why it needs to be done

Users need to be able to perform the described actions so that custom codes can be used as part of a configuration.

## 🏗️ Acceptance Criteria

- [ ] New custom code can be added via the UI and saved to a configuration in the database
- [ ] An existing custom code associated with a configuration can be modified via the UI and saved to the database
- [ ] An existing custom code associated with a configuration can be deleted via the UI
- [ ] The `/configurations/:id` route has been modified so that custom code data is included as a part of the response. The UI has been updated to use this data

## ⚙️ Technical Notes

TBD

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds the ability for users to modify their configurations by adding, editing, and deleting custom codes.

## 🔗 Related Issue

Fixes #310 

## 🧪 How to test

Unit tests

`npm run test` in the client to run the new client-side tests

`pytest tests/` in refiner to run the server-side tests

Testing the application:

1. Run `docker compose up`
4. Create a new configuration
5. Navigate to `Custom codes`, which should have a `0` next to it
6. Click `Add code` to add a new code
7. Once added, try to edit that code with the `Edit` link
8. Once edited, try to delete that code with the `Delete` link

>[!NOTE]
> You should not be able to add a code with an existing code/system pair. Nothing will happen if you try to. You should also not be able to edit a code that could conflict with an existing code/system pair. This will result in the user receiving an error toast.

## ℹ️ Additional Information

N/A

### DIFF
--- a/refiner/app/db/configurations/model.py
+++ b/refiner/app/db/configurations/model.py
@@ -13,6 +13,9 @@ class DbConfigurationCondition(BaseModel):
     version: str
 
 
+# TODO: Revisit this to see if we can figure out how to reduce overlap with other types.
+# This is a "custom_code" on the configuration row in the configurations table.
+# This is one object in the `custom_codes` list.
 class DbConfigurationCustomCode(BaseModel):
     """
     Custom code associated with a Configuration.


client/eslint.config.js
```diff
@@ -43,6 +43,7 @@ export default tseslint.config(
       '@typescript-eslint/no-unsafe-return': 'off',
       '@typescript-eslint/no-unsafe-argument': 'off',
       '@typescript-eslint/no-misused-promises': 'off',
+      '@typescript-eslint/no-redundant-type-constituents': 'off',
```
jakewheeler, line 46: FYI - I've been continuously modifying our ESLint config due to Orval's generated code not 100% complying with the default rules. Not an issue, just something to be aware of.

refiner/app/db/configurations/db.py
```diff
@@ -309,3 +309,100 @@ async def get_total_condition_code_counts_by_configuration_db(
...
+
+    custom_codes = config.custom_codes
+
+    exists = any(
+        (c.code == custom_code.code and c.system == custom_code.system)
+        for c in custom_codes
+    )
+
+    if not exists:
+        custom_codes.append(custom_code)
```
jakewheeler, line 337: I tried to take a different path this time and am seeing that doing the heavy lifting in Python instead of SQL is probably the way to go in many situations. This makes the query way simpler.

PR #274, opened 2025-08-15
feat(ui): Add condition code set `<Drawer />` component

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Borrows a component from `cdcgov/dibbs-query-connecter` for the new `<Drawer />`
component which is the interface for the _Add condition code sets_ form for the
`<ConfigBuild />` page.

## 🔗 Related Issue

Related to #172

## ✅ Acceptance Criteria

- [x] Code set for given reportable condition is added by default
- [x] Can select other reportable condition code sets to add in side-bar
- [x] Condition(s) added toast message
- [x] Code sets added to side bar
- [x] Can delete code sets from "CODE SETS" bar
- [x] Search bar in sidebar optional if high effort

## 🧪 How to test

Click around the application.

refiner-add-codeset-drawer

PR #321, opened 2025-09-04
refactor: use cursor from async database connection

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR makes an update to `AsyncDatabaseConnection` so that it only provides a `get_connection` and no longer provides a `get_cursor`. 

The reason for this change is because our usage of `cursor` has become more complex over time. We used to only return the `dict_row` type when querying the database, but we have progressed to now primarily returning Pydantic models. Using the `cursor` that comes from `get_connection` requires no maintenance overhead on us and will work out of the box regardless of the return type we want to use.

As an example, we've gone from this:
```python
# old `get_cursor` always used `dict_row` by default
async with db.get_cursor() as cur:
    now = dt.now(UTC)
    await cur.execute(
        """
        SELECT users.*, sessions.expires_at
        FROM sessions
        JOIN users ON sessions.user_id = users.id
        WHERE sessions.token_hash = %s AND sessions.expires_at > %s
        """,
        (token_hash, now),
    )
```

to this:
```python
    # Using the connection's `cursor` allows us to use `dict_row` or a Pydantic model. It is much more flexible
    async with conn.cursor(row_factory=dict_row) as cur:
        now = dt.now(UTC)
        await cur.execute(
            """
            SELECT users.*, sessions.expires_at
            FROM sessions
            JOIN users ON sessions.user_id = users.id
            WHERE sessions.token_hash = %s AND sessions.expires_at > %s
            """,
            (token_hash, now),
        )
```

A lot of our more structured queries will use `class_row` instead, like this:
```python
async def get_user_by_id_db(id: str, db: AsyncDatabaseConnection) -> DbUser:
    query = """
            SELECT id, username, email, jurisdiction_id
            FROM users
            WHERE id = %s
            """
    params = (id,)

        # Here we use `DbUser` instead of `dict_row`
        async with conn.cursor(row_factory=class_row(DbUser)) as cur:
            await cur.execute(query, params)
            row = await cur.fetchone()

    if not row:
        raise Exception(f"User with ID {id} not found.")

    return row
```

## 🧪 How to test

Automated tests should be passing and the app should function exactly as it did previously.


rogeruiz:
Looks pretty straight forward. But this is going to cause issues for my PR right? I'll need to refactor stuff ahead of time? #274

rogeruiz (approved):
lgtm 🌈 

Straight-forward changes. Thanks @jakewheeler

PR #324, opened 2025-09-04
WIP: configuration table updates and create pathway for refinement to be configuration driven

### LINKED ISSUE #285: [USER STORY] Refiner should accept my configuration as an input
## 👤💭 User Story

As a user, I want to be able to create my own configuration and use it with the Refiner. I expect the Refiner to use the code sets I've defined in my configuration during the refining process.

## 🏗️ Acceptance Criteria

- [ ] Refiner UI passes config ID to the server to use with refining
- [ ] Refiner function (sync and async) have been updated to take an input based on a user-defined configuration
- [ ] Refiner makes use of codes defined in the user's configuration

## 📜 Additional context

TBD

robertmitchellv:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #285 
- #285

### DIFF
--- a/refiner/migrations/000002_update_configurations.down.sql
+++ b/refiner/migrations/000002_update_configurations.down.sql
@@ -1,0 +1,21 @@
+-- if we need to rollback the migration changes; run the following
+
+-- restore original columns
+ALTER TABLE configurations
+    ADD COLUMN name TEXT,
+    ADD COLUMN description TEXT,
+    ADD COLUMN loinc_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN snomed_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN icd10_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN rxnorm_codes_additions JSONB DEFAULT '[]'::jsonb;
+
+-- remove new columns
+ALTER TABLE configurations
+    DROP CONSTRAINT IF EXISTS fk_primary_condition,
+    DROP COLUMN IF EXISTS condition_id,
+    DROP COLUMN IF EXISTS local_codes;
+
+-- restore original constraint
+ALTER TABLE configurations
+    ADD CONSTRAINT check_included_conditions
+        CHECK (jsonb_array_length(included_conditions) > 0);
--- a/refiner/migrations/000002_update_configurations.up.sql
+++ b/refiner/migrations/000002_update_configurations.up.sql
@@ -1,0 +1,38 @@
+-- add new columns
+-- **condition_id**
+-- initially we planned to use the `activations` table to track the explicit relationships between
+-- the RC SNOMED codes, RSG codes, and their parent CG. instead, we'll ensure that all configurations
+-- are _explicitly_ tied to a condition. this will ensure that we know immediately which condition any
+-- configuration is set up for and allow for `included_conditions` to simply track both the primary
+-- condition and its secondary conditions
+-- **local_codes**
+-- initially we planned to store the snomed, loinc, icd10, and rxnorm codes in:
+-- - loinc_codes_additions, snomed_codes_additions, icd10_codes_additions, and rxnorm_codes_additions
+-- instead we'll store the data as a FHIR Coding (code, display, system) in `custom_codes`
+-- in anticipation of needing the same functionality but for nonstandard code systems and local codes,
+-- we're adding `local_codes` and let it function in a similar way to `custom_codes`
+ALTER TABLE configurations
+    ADD COLUMN condition_id UUID REFERENCES conditions(id),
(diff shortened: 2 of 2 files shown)


refiner/app/db/configurations/db.py
```diff
@@ -6,81 +6,93 @@
...
 from ..pool import AsyncDatabaseConnection
-from .model import DbConfiguration, DbConfigurationCustomCode
+from .model import (
+    DbConfiguration,
+    DbConfigurationCustomCode,
+)
+
+# NOTE: the db table for configurations has a family_id field (for future versioning/families),
+# but this is intentionally omitted from the pydantic models and sql queries for now.
+# if/when versioning is fully implemented, we can decide exactly how we'll tackle it
```
jakewheeler, line 16: If we are unsure whether or not we'll make use of this, should we remove it for now and add it back in later if a need for it becomes obvious?

refiner/app/db/configurations/db.py
```diff
@@ -190,6 +203,8 @@ async def associate_condition_codeset_with_configuration_db(
     """
     Given a condition, associate its set of codes with the specified configuration. Prevents the addition of duplicate conditions.
 
+    NOTE: If the primary condition ever changes, you should also update the `name` field to match the new condition's display_name.
+
```
jakewheeler, line 207: Should we get rid of a `configuration`'s `name` and instead join a config on the `condition` by `condition_id` and use its `display_name`? Then we could do something like this (for example) and `name` is always automatically in sync:
```sql
SELECT
    c.id,
    co.display_name AS name,
    c.jurisdiction_id,
    c.condition_id,
    c.included_conditions,
    c.custom_codes,
    c.local_codes,
    c.sections_to_include,
    c.cloned_from_configuration_id,
    c.version
FROM configurations c
LEFT JOIN conditions co
  ON c.condition_id = co.id
WHERE c.jurisdiction_id = 'SDDH'
ORDER BY name ASC;
```

refiner/migrations/000002_update_configurations.down.sql
```diff
@@ -0,0 +1,19 @@
...
+    ADD COLUMN description TEXT,
+    ADD COLUMN loinc_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN snomed_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN icd10_codes_additions JSONB DEFAULT '[]'::jsonb,
+    ADD COLUMN rxnorm_codes_additions JSONB DEFAULT '[]'::jsonb;
+
+-- remove new columns
+ALTER TABLE configurations
+    DROP COLUMN IF EXISTS condition_id,
+    DROP COLUMN IF EXISTS local_codes;
```
jakewheeler, line 14: I don't know that it matters much either way - do we need `IF EXISTS` since we can safely assume migrations will run in order? 

One pro I can think of for leaving it off would be for us to easily know if something is wrong with the migrations since they'd break when running.

refiner/app/db/conditions/model.py
```diff
@@ -3,12 +3,36 @@
 from pydantic import BaseModel
 
 
+class DbCoding(BaseModel):
```
jakewheeler, line 6: Should we call this `DbConditionCode` to keep it in-line with some of our other naming? Like `DbConfigurationLocalCode`, `DbConfigurationCustomCode`, etc.

refiner/app/db/configurations/db.py
```diff
@@ -95,25 +107,24 @@ async def get_configurations_db(
...
             custom_codes,
+            local_codes,
             sections_to_include,
-            cloned_from_configuration_id
+            cloned_from_configuration_id,
+            version
         FROM configurations
         WHERE jurisdiction_id = %s
-        ORDER BY name asc;
+        ORDER BY id asc;
```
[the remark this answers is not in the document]
robertmitchellv: oh good catch! yeah, this was unintended

refiner/migrations/000002_update_configurations.up.sql
```diff
@@ -0,0 +1,33 @@
...
+
+-- drop unnecessary columns
+-- since we're not using these to store user driven "custom codes" there's
+-- no reason to keep them around
+ALTER TABLE configurations
+    DROP COLUMN IF EXISTS description,
+    DROP COLUMN IF EXISTS loinc_codes_additions,
+    DROP COLUMN IF EXISTS snomed_codes_additions,
+    DROP COLUMN IF EXISTS icd10_codes_additions,
+    DROP COLUMN IF EXISTS rxnorm_codes_additions;
```
jakewheeler, line 26: Should we remove the `IF EXISTS` from these as well? And line `33`?

robertmitchellv: i suppose you're right. i'm going to need to redo migrations then to test this out once the change is made

PR #337, opened 2025-09-09
feat: add username to nav bar + dropdown with logout button

### LINKED ISSUE #174: [USER STORY] Basic user information
## 👤💭 User Story

As a user, I want to see my log-in information, so that I know I'm logged in to the right account and organization

## 🏗️ Acceptance Criteria

- [ ] Username in header on every page

lina-roth:
# 🔀 PULL REQUEST

Screenshot 2025-09-09 at 2 09 28 PM

## 💡 Summary

- [ ] Add logged in user to navigation bar
- [ ] Add dropdown with logout button

## 🔗 Related Issue

Fixes #174 
- #174 

## ✅ Acceptance Criteria

- [ ]  Username in header on every page

## 🧪 How to test

1. Run `docker compose up`
4.  Login and see username in nav bar
5. Click username, click Logout button, see that you are logged out

### DIFF
--- a/client/src/components/Layout/index.tsx
+++ b/client/src/components/Layout/index.tsx
@@ -4,7 +4,7 @@ import CdcLogo from '../../assets/cdc-logo.svg';
 
 import NavigationBar from '../NavigationBar';
 import { Icon } from '@trussworks/react-uswds';
-import { Menu, MenuButton, MenuItem, MenuItems } from '@headlessui/react'
+import { Menu, MenuButton, MenuItem, MenuItems } from '@headlessui/react';
 
 interface LayoutProps {
   children: React.ReactNode;
@@ -17,7 +17,7 @@ export function Layout({ username, children }: LayoutProps) {
       <a className="usa-skipnav" href="#main-content">
         Skip to main content
       </a>
-      <Header username={username}/>
+      <Header username={username} />
       <main
         id="main-content"
         className="bg-primary-container flex grow flex-col"
@@ -50,19 +50,19 @@ export function Header({ username }: HeaderProps) {
 
         {/* User dropdown */}
         <Menu as="div" className="relative">
-          <MenuButton className="font-public-sans flex items-center gap-2 text-white px-3 py-2 rounded hover:bg-blue-cool-70 focus:outline-none">
+          <MenuButton className="font-public-sans hover:bg-blue-cool-70 flex items-center gap-2 rounded px-3 py-2 text-white focus:outline-none">
             <Icon.Person size={3} aria-hidden />
             {username}
           </MenuButton>
 
           <MenuItems
-              anchor="bottom"
-              className="absolute right-0 mt-2 w-40 origin-top-right rounded-md bg-white shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none"
+            anchor="bottom"
+            className="ring-opacity-5 absolute right-0 mt-2 w-40 origin-top-right rounded-md bg-white shadow-lg ring-1 ring-black focus:outline-none"
           >
             <MenuItem>
               <a
-                  href="/api/logout"
-                  className="font-public-sans block w-full px-4 py-2 text-sm text-gray-700 data-[focus]:bg-gray-100 data-[focus]:text-gray-900"
+                href="/api/logout"
+                className="font-public-sans block w-full px-4 py-2 text-sm text-gray-700 data-[focus]:bg-gray-100 data-[focus]:text-gray-900"
               >
                 Logout
               </a>
--- a/client/src/components/Layout/index.tsx
+++ b/client/src/components/Layout/index.tsx
@@ -1,8 +1,12 @@
+import { useEffect, useState, useRef } from 'react';
 import { Link } from 'react-router';
 import DibbsLogo from '../../assets/dibbs-logo.svg';
 import CdcLogo from '../../assets/cdc-logo.svg';
 
 import NavigationBar from '../NavigationBar';
+import { getUser } from '../../api/user/user.ts';
+import type { GetUser200 } from '../../api/schemas';
+import { Icon } from '@trussworks/react-uswds';
 
 interface LayoutProps {
   children: React.ReactNode;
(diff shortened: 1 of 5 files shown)


client/src/components/Layout/index.tsx
```diff
@@ -27,6 +31,33 @@ export function Layout({ children }: LayoutProps) {
...
+
+  useEffect(() => {
+    getUser()
+      .then((response) => {
+        setCurrentUser(response.data);
+      })
+      .catch((err) => {
+        console.error('Failed to load user', err);
+      });
+  }, []);
```
jakewheeler, line 46: When the [app loads we call `useLogin`](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/client/src/App.tsx#L14) which makes a request to the server to fetch the user, so we already user info by this point and don't need to make another request. We have two main options:

1. When we call `useLogin` in `App.tsx`, we pass the `user` to `Layout` and then to `Header`
2. When we call `useLogin` we store the returned `user` globally using a Context and then make use of it here

Since we don't need user data outside of the header, option 1 seems like the best way to go. Passing the `user` from `App > Layout > Header` is much lower overhead than a Context.

client/src/components/Layout/index.tsx
```diff
@@ -27,6 +31,33 @@ export function Layout({ children }: LayoutProps) {
 }
 
 export function Header() {
+  const [currentUser, setCurrentUser] = useState<GetUser200>(null);
+  const [menuOpen, setMenuOpen] = useState(false);
```
jakewheeler, line 35: Mostly an FYI since we aren't going to be making a request to the server here: we'll want to use the generated mutation/query hooks instead of calling `fetch` or `axios` directly (this is what `getUser` is doing).

[Here's an example of what that looks like.](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/client/src/pages/Configurations/index.tsx#L48) Let me know if you have any questions about this!

client/src/components/Layout/index.tsx
```diff
@@ -40,12 +71,29 @@ export function Header() {
...
+
+        {/* User dropdown */}
+        <div className="relative" ref={menuRef}>
+          <button
+            type="button"
+            onClick={() => setMenuOpen((prev) => !prev)}
+            className="font-public-sans hover:bg-blue-cool-70 flex items-center gap-2 rounded px-3 py-2 text-white focus:outline-none"
+          >
+            <Icon.Person size={3} aria-hidden />
+            {currentUser ? currentUser.username : 'Loading...'}
```
jakewheeler, line 83: This will no longer need to load with the other changes mentioned.
```suggestion
            {user.username}
```

client/src/components/Layout/index.tsx
```diff
@@ -40,12 +71,29 @@ export function Header() {
...
-          Logout
-        </a>
+
+        {/* User dropdown */}
+        <div className="relative" ref={menuRef}>
+          <button
+            type="button"
+            onClick={() => setMenuOpen((prev) => !prev)}
+            className="font-public-sans hover:bg-blue-cool-70 flex items-center gap-2 rounded px-3 py-2 text-white focus:outline-none"
+          >
```
jakewheeler, line 81: There is currently no way for a screen reader to know this button opens up a menu. We should probably add a few aria attributes here. I _think_ something like this would work.
```suggestion
          <button
            type="button"
            aria-haspopup="menu"
            aria-expanded={menuOpen}
            onClick={() => setMenuOpen((prev) => !prev)}
            className="font-public-sans hover:bg-blue-cool-70 flex items-center gap-2 rounded px-3 py-2 text-white focus:outline-none"
          >
```

client/src/components/Layout/index.tsx
```diff
@@ -40,12 +71,29 @@ export function Header() {
...
+          {menuOpen && (
+            <div className="ring-opacity-5 absolute right-0 mt-2 w-40 overflow-hidden rounded-md bg-white shadow-lg ring-1 ring-black">
+              <a
+                href="/api/logout"
+                className="font-public-sans block w-full px-4 py-2 text-sm text-gray-700 hover:bg-gray-100"
+              >
+                Logout
+              </a>
+            </div>
+          )}
```
jakewheeler, line 95: More a11y things - we should denote to screen readers that this is a menu opened by the button. We should also give the menu some structure using a list. 

I believe something like this works:
```suggestion
          {menuOpen && (
            <ul
              role="menu"
              className="ring-opacity-5 absolute right-0 mt-2 w-40 overflow-hidden rounded-md bg-white shadow-lg ring-1 ring-black"
            >
              <li role="none">
                <a
                  href="/api/logout"
                  role="menuitem"
                  className="font-public-sans block w-full px-4 py-2 text-sm text-gray-700 hover:bg-gray-100"
                >
                  Logout
                </a>
              </li>
            </ul>
          )}
```

client/src/components/Layout/index.tsx
```diff
@@ -27,6 +31,33 @@ export function Layout({ children }: LayoutProps) {
...
+      if (menuRef.current && !menuRef.current.contains(event.target as Node)) {
+        setMenuOpen(false);
+      }
+    }
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, []);
+
```
jakewheeler, line 60: There are a few other things to consider here:

1. When the menu opens the first item in the list should receive focus automatically
2. When the menu closes the focus should go back to the button that opened it automatically
3. We need to make this work generally with a keyboard (menu navigation, closing the menu, etc.)

It honestly may be worth looking into a [pre-built component like this](https://headlessui.com/react/menu) that we know has a11y considerations already baked into it and we can simply apply our styling to it. 

I would suggest going this route for now instead of us maintaining our own solution. There are likely many more a11y considerations beyond the ones I've listed that will be implemented in the component I linked and the likelihood of us covering them all is not very high IMO.

client/src/components/Layout/index.tsx
jakewheeler: Disregard this if we go with my other suggestion to use the headless UI component.

client/src/components/Layout/index.tsx
jakewheeler: Disregard this if we go with my other suggestion to use the headless UI component.

jakewheeler (changes requested):

client/src/components/Layout/index.tsx
lina-roth: good call

PR #373, opened 2025-09-16
fix: accept S3 upload bucket name as an environment variable

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the app to accept an `S3_UPLOADED_FILES_BUCKET_NAME`, which is the bucket where user uploaded files will be stored. It was previously hard-coded to `refiner-app` but now that we're deploying to a live environment this won't suffice anymore.

> [!NOTE]
> Ignore the generated files. It looks like the latest changes may not have made it into a previous PR and should have. This is not related to the environment variable change.

## 🧪 How to test

Refined file downloading should continue to work locally as it did previously.

**Make sure to stop your containers and bring them up again so the environment variable gets loaded in**.


lina-roth (approved):
✅ nice!

PR #376, opened 2025-09-17
wip: inline testing flow

jakewheeler:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #
- #


refiner/app/api/v1/demo.py
```diff
@@ -219,33 +86,29 @@ async def demo_upload(
...
                     refined_eicr=stripped_refined_eicr,
                     stats=[
                         f"eICR file size reduced by {
-                            _get_file_size_difference_percentage(
+                            get_file_size_reduction_percentage(
                                 original_xml_files.eicr, condition_refined_eicr
                             )
                         }%",
                     ],
-                    processing_info=ConditionProcessingInfo(
```
jakewheeler, line 244: I removed this since:
1. It is unused by the client
2. Made it simpler for me to define a `Condition` in the new route

refiner/tests/integration/test_demo.py
```diff
@@ -10,21 +10,22 @@
...
+from app.services.ecr.refine import get_file_size_reduction_percentage
+from app.services.file_io import create_refined_ecr_zip_in_memory
 
 api_route_base = "/api/v1/demo"
 
 client = TestClient(app=app)
 
 
 @pytest.mark.asyncio
+@pytest.mark.integration
```
jakewheeler, line 28: Am I using this correctly? Should this ONLY be used when interacting with the containerized environment?

I am wondering if I've been using it too loosely. For example, if I interact with files on the filesystem in my test, do I mark a test with this?

client/src/api/configurations/configurations.ts
```diff
@@ -674,4 +676,83 @@ export const useDeleteCustomCodeFromConfiguration = <TError = AxiosError<HTTPVal
...
+    upload_refined_files_to_s3 (Callable[[UUID, io.BytesIO, str, Logger], str]): service to upload a zip to S3
+    user (DbUser): Logged in user
+    db (AsyncDatabaseConnection): Database connection
+    sample_zip_path (Path): Path to example .zip eICR/RR pair
+    logger (Logger): Standard logger
+Returns:
+    ConfigurationTestResponse: response given to the client
+ * @summary Run Configuration Test
+ */
+export const runInlineConfigurationTest = (
```
fzhao99, line 695: is this function codegened at all, or is it just taking the generated types from Orval to facilitate the different testing pieces? Trying to grok how the different parts fit together

client/src/pages/Configurations/ConfigTest/index.test.tsx
```diff
@@ -25,4 +110,76 @@ describe('Config testing page', () => {
       screen.getByText('Turn on configuration', { selector: 'a' })
     ).toBeInTheDocument();
   });
+
+  it('should warn the user that the expected condition was not found during inline testing', async () => {
+    const user = userEvent.setup();
+
+    const warningMessage =
+      'Condition was not detected in the eCR file you uploaded.';
```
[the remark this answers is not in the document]
jakewheeler: The text that gets rendered in the component is dynamic and will come from the server in a real scenario. This specific text is truly only needed for the test case.

client/src/api/configurations/configurations.ts
fzhao99: yep saw that page so wanted to make sure that's what was happening here. thanks so much!

refiner/app/api/v1/demo.py
robertmitchellv: sounds good; if we need some kind of workflow style object in the future we can rethink something like this. i was wondering what it did earlier this week

refiner/tests/integration/test_demo.py
robertmitchellv: i think the idea is to mark as integration only those tests that require the environment started by testcontainers/docker compose and for things that are just pure utilities that don't necessarily need the environment that was spun up by testcontainers/docker compose to not mark them as integration

tl;dr is if you **want** everything to be ran **from** the container, than everything should use the `@pytest.mark.integration`  decorator

robertmitchellv (approved):
overall i think this makes the routes a lot cleaner and easier to extend for what i'm working on. i like the new services and the new api utilities as well. great job with this 🥇 

and everything worked as expected locally as well :shipit:

Issue #146, opened 2025-06-17
[USER STORY] Section-based Inclusion/Omission


kplaneaux:
## 🏗️ Acceptance Criteria

- [ ] ..

## 🎬 Scenarios

- ...

## 🎨 Design Notes

- ...

PR #378, opened 2025-09-19
feat: Add new Section-based Inclusion/Omission for Configurations

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch adds a new Sections row to show an eICR Section Review pane on the
right-hand side of the page. This is not tied to the refiner API yet. It's only
the UI components.

## 🔗 Related Issue

Related to: #146

## ✅ Acceptance Criteria

- [ ] User can indicate whether to retain, refine, or omit sections of the eCR
- [ ] Can ignore "N/A" as an option for now
- [ ] Section names should come from display names in refiner/assets/refiner_details.json
  - [ ] History of encounters - 46240-8
  - [ ] History of Present Illness - 10164-2
  - [ ] History of Immunizations - 11369-6
  - [ ] Medications Administered - 29549-3
  - [ ] Plan of Treatment - 18776-5
  - [ ] Problem List - 11450-4
  - [ ] Reason For Visit - 29299-5
  - [ ] Relevant diagnostic tests and/or laboratory data - 30954-2
  - [ ] Social History - 29762-2
- [ ] Should be clear in eICR output whether section was omitted or refined if empty

## 🧪 How to test

Run `just client test` and load the page and navigate to a configuration. Note
the new `Sections` under **More options** on the left-hand side. Click it and
see the new eICR Section Review component.

### DIFF
--- a/client/src/components/Button/RadioCell.tsx
+++ b/client/src/components/Button/RadioCell.tsx
@@ -5,7 +5,7 @@ export interface RadioCellProps {
   action: UpdateSectionProcessingEntryAction;
   checked: boolean;
   ariaLabel: string;
-  applyAction?: (
+  applyAction: (
     index: number,
     action: UpdateSectionProcessingEntryAction
   ) => void;
@@ -26,11 +26,11 @@ export function RadioCell({
     <td
       className="text-center !break-all !whitespace-normal focus:outline-none"
       tabIndex={0}
-      onClick={() => applyAction && applyAction(index, action)}
+      onClick={() => applyAction(index, action)}
       onKeyDown={(e) => {
         if (e.key === ' ' || e.key === 'Enter') {
           e.preventDefault();
-          applyAction && applyAction(index, action);
+          applyAction(index, action);
         }
       }}
     >
--- a/client/src/pages/Configurations/ConfigBuild/EicrSectionReview.test.tsx
+++ b/client/src/pages/Configurations/ConfigBuild/EicrSectionReview.test.tsx
@@ -3,7 +3,6 @@ import React from 'react';
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { QueryClient } from '@tanstack/react-query';
 import { TestQueryClientProvider } from '../../../test-utils';
 import EicrSectionReview from './EicrSectionReview';
 import { DbConfigurationSectionProcessing } from '../../../api/schemas/dbConfigurationSectionProcessing';


robertmitchellv:
do you think we need a way to add a horizontal scroll in the event that a user is quite zoomed in?

image

this is sort of my default zoom but if needed my workaround is to just move it to my larger screen

robertmitchellv:
not sure if this is a me thing or a firefox derived thing (with zen browser, which i use) but i'm getting this empty white space under the footer when looking at the sections part

image

rogeruiz:
>do you think we need a way to add a horizontal scroll in the event that a user is quite zoomed in?

We should be able to still scroll within the Table inside the small area down on the right-hand side. It's not great though.

===

Also I've updated the styles a bit more to hopefully have things match Figma a bit better.

rogeruiz:
Hmm not sure. I'll give Zen Browser a try and see if I can reproduce this! thanks!

client/src/pages/Configurations/ConfigBuild/EicrSectionReview.tsx
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * EicrSectionReview displays an overview or review of eICR sections and allows
+ * users to choose an action for each section (retain, refine, remove).
+ * Radio inputs are fully accessible and can be selected by clicking anywhere
+ * in the containing table cell (td), supporting keyboard navigation as well.
+ */
+const EicrSectionReview: React.FC<{
+  configurationId: string;
+  sectionProcessing: DbConfigurationSectionProcessing[];
+}> = ({ configurationId, sectionProcessing }) => {
```
jakewheeler, line 18: I think it'd be ideal if we didn't make use of `React.FC` for the sake of consistency. Our typical approach seems to be:
```suggestion
interface EicrSectionReviewProps {
  configurationId: string;
  sectionProcessing: DbConfigurationSectionProcessing[];
}

export function EicrSectionReview({
  configurationId,
  sectionProcessing,
}: EicrSectionReviewProps) {
```

refiner/app/db/configurations/db.py
```diff
@@ -79,7 +96,7 @@ async def insert_configuration_db(
         # local_codes
         EMPTY_JSONB,
         # section_processing
-        EMPTY_JSONB,
+        Jsonb(section_processing_defaults),
```
jakewheeler, line 99: One thing to consider here going forward is that this is not backwards compatible. Meaning if a config was created prior to this change it will not receive this necessary data. 

Here's what I see happen in that scenario:
image

I think this is perfectly fine for now to be clear - we are only running in the demo environment where data loss is expected, so we can refresh the env and everything is good. 

Mostly just wanted to mention it as a reminder that we'll eventually need to write migrations to update old data. That said, changes to configurations will require a LOT of consideration once activations are a thing 😅

refiner/app/db/configurations/db.py
```diff
@@ -546,6 +563,86 @@ async def edit_custom_code_from_configuration_db(
...
+        Updated DbConfiguration or None if the update fails
+    """
+
+    # Validate input actions
+    valid_actions = {"retain", "refine", "remove"}
+    for su in section_updates:
+        if "code" not in su or "action" not in su:
+            raise ValueError("Each section update must contain 'code' and 'action'.")
+        if su["action"] not in valid_actions:
+            raise ValueError(f"Invalid action '{su['action']}' for section update.")
```
jakewheeler, line 589: Do we want to change the type of `section_updates` to something more concrete (rather than `list[dict[str, str]]`) to reduce having to do this kind of checking?

client/src/pages/Configurations/ConfigBuild/EicrSectionReview.tsx
```diff
@@ -0,0 +1,228 @@
...
+          <b>Include entire section:</b> Includes everything from this section,
+          ignoring your configuration.
+        </li>
+        <li>
+          <b>Remove section:</b> Excludes this section from the eICR entirely.
+        </li>
+      </ul>
+      <Table
+        bordered
+        className="margin-top-2 !max-w-[740px] !border-separate border-spacing-0 rounded-full border-l-0"
```
jakewheeler, line 168: Is `margin-top-2` USWDS styling? Was this intended to be `mt-2`?

refiner/app/db/configurations/db.py
jakewheeler: No, I think you went about it just fine 👍 We'll need to make sure we reset the DB when this change gets deployed but I think we can get away with doing this type of thing until we get to a point where data loss is no longer acceptable.

PR #394, opened 2025-09-26
test: update integration test to use correct endpoint

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

- Update the route in the test to include `/v1/`
- Update mock AWS region to `us-east-1` (arbitrary choice) so that localstack doesn't display a warning about an incorrect region being used


rogeruiz (approved):
lgtm 🌈

PR #414, opened 2025-09-29
feat: implement configuration based independent testing workflow

### LINKED ISSUE #343: [TASK] (Server) Configuration-Based Independent Testing
## 🚧 What needs to be done

Update independent testing to use configurations instead of conditions table, completing the transition away from legacy data sources.

## 🤔 Why it needs to be done

- Configurations represent the user-defined business logic that should drive the Refine generally and the independent testing suite specifically
- This completes the modernization of the independent testing workflow from a back end POV
- Provides foundation for consistent testing approach across sync/async and independent/inline modes (cornerstone of future work)
- Aligns testing with actual user configuration workflows (rather than just the codes from the `condition` table)

## 🏗️ Acceptance Criteria

- [ ] Independent testing queries configurations table instead of conditions (there will need to be service level functions to help make this link more explicit)
- [ ] Configuration-to-test-data mapping implemented (meaning end-to-end it's working with our test file)
- [ ] All existing independent testing scenarios work with configurations (we may have more test data at this time or just the one bundle--either way whatever we have for testing should have existing seeding to enable out of the box functionality for local deployment)
- [ ] Test data seeding creates appropriate default configurations (JD that matches RR, configurations for conditions found in the RR, etc.)
- [ ] Documentation updated as needed

## ⚙️ Technical Notes

- Leverage `ProcessedConfiguration` class from #324
- Update test data fixtures to include configurations
- Ensure configuration validation catches test setup issues
- Consider caching strategies for configuration processing

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This pull request introduces a significant refactoring to implement a new configuration-based independent testing workflow. The primary goal is to make the eICR refinement process more modular, testable, and driven by explicit configurations.

The key changes include:

-   **Refactored `ecr/` Service**: The service layer under `ecr/` has been streamlined to focus exclusively on XML processing. All orchestration logic, database interactions, and asynchronous operations have been removed, resulting in pure functions that are easier to compose and test in different contexts (e.g., application APIs, Lambda functions).
-   **New `testing` Service**: A new `app/services/testing.py` service has been created to orchestrate the "independent testing" workflow. This service now handles the logic that was previously in the `demo` route, including fetching data, processing conditions, and calling the `ecr` service for refinement. It uses a trace object (`IndependentTestingTrace`) to track the state of each condition through the pipeline.
-   **Updated Demo Route**: The `/api/v1/demo` endpoint has been simplified to focus on handling web request/response logic. It calls the new `testing` service to perform the core workflow, making the route cleaner and more aligned with its responsibility as a controller.

## 🔗 Related Issue

Fixes #343
- #343

## ✅ Acceptance Criteria

- [x] Independent testing queries configurations table instead of conditions (there will need to be service level functions to help make this link more explicit)
- [x] Configuration-to-test-data mapping implemented (meaning end-to-end it's working with our test file)
  - this is for the integration tests only right now
- [x] All existing independent testing scenarios work with configurations (we may have more test data at this time or just the one bundle--either way whatever we have for testing should have existing seeding to enable out of the box functionality for local deployment)
  - still need more test data
- [x] Test data seeding creates appropriate default configurations (JD that matches RR, configurations for conditions found in the RR, etc.)
  - for the integration tests yes, in the dev app not right now
- [ ] Documentation updated as needed

### Unit & Integration Tests

Extensive work has been done on unit and integration tests to cover the new and refactored services. The integration test database is now seeded with two sample configurations (COVID and Influenza) to validate the configuration-based workflow.
Unit tests

5. Additionally you can check coverage with `pytest --cov=app tests` (but we're not officially tracking right now)

### Manual Application Testing

The following steps allow you to test the configuration-driven logic.

1.  Run `docker compose up`.
2.  Run `just db refresh`.
3.  Navigate to the application frontend at [http://localhost:8081](http://localhost:8081).
4.  Use the default sample zip file for testing.

**Testing Scenarios:**

-   **Without a configuration**: With a clean database, attempting to refine the eICR will result in the app not really knowing what to do (since we'll be connecting those dots later)
-   **With a COVID configuration**: Add a configuration for COVID. The application will now correctly identify the COVID condition in the RR and produce a refined eICR document for it (drop down will just show COVID-19)
-   **With COVID and Influenza configurations**: Add a second configuration for Influenza. The application will now discover and refine both conditions independently, producing two refined documents.

## ℹ️ Additional Information

-   The `configuration` route was updated minimally to ensure the inline testing continues to function. A more thorough implementation will be handled in a subsequent ticket.
-   The new `testing` service is designed to be extensible. The current implementation is for "independent testing," but it will be expanded to handle "inline testing" in an upcoming feature.


refiner/app/api/v1/configurations.py
```diff
@@ -1004,7 +1004,7 @@ async def run_configuration_test(
 
     # TODO: this will need to be replaced with a version of `refine` that accepts
     # and makes use of a configuration.
-    refined_results = await refine(
+    refined_results = await refine_eicr(
```
jakewheeler, line 1007: We'll need to update the args here for this to work.

refiner/app/api/v1/demo.py
```diff
@@ -129,10 +153,11 @@ async def demo_upload(
             logger,
         )
 
+        # STEP 7:
+        # construct and return the response model
         formatted_unrefined_eicr = format.strip_comments(
             format.normalize_xml(original_xml_files.eicr)
         )
-
         return RefinedTestingDocument(
```
jakewheeler, line 161: We'll want to update the return type to include the names of conditions in which we didn't find matching configs. `RefinedTestingDocument` might not make sense as a name anymore.

refiner/app/db/configurations/db.py
```diff
@@ -544,3 +544,53 @@ async def edit_custom_code_from_configuration_db(
         return None
 
     return DbConfiguration.from_db_row(row)
+
+
+async def get_configurations_by_condition_ids_and_jurisdiction(
```
jakewheeler, line 549:
```suggestion
async def get_configurations_by_condition_ids_and_jurisdiction_db(
```

refiner/app/services/testing.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # output only successful refinements
+    refined_documents = [
+        entry["refined_document"] for entry in trace if entry["refined_document"]
+    ]
+
+    return {"refined_documents": refined_documents}
+
+
+def extract_reportable_conditions_for_jurisdiction(
```
jakewheeler, line 144: Should this function and the others in this file that aren't `independent_testing` be marked "private"?
```suggestion
def _extract_reportable_conditions_for_jurisdiction(
```

refiner/tests/assets/mon-mothma-covid-lab-positive_RR.xml
```diff
@@ -432,7 +432,7 @@
                         <templateId extension="2017-04-01"
                           root="2.16.840.1.113883.10.20.15.2.4.1" />
                         <participantRole>
-                          <id extension="ht" root="2.16.840.1.113883.4.6" />
+                          <id extension="SDDH" root="2.16.840.1.113883.4.6" />
```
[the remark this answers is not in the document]
robertmitchellv: yeah this was sad to finally notice 😭

refiner/tests/integration/conftest.py
```diff
@@ -15,13 +15,14 @@
 TEST_SESSION_TOKEN_HASH = get_hashed_token(TEST_SESSION_TOKEN)
 
 # User info
-TEST_USERNAME = "test-user"
-TEST_USER_ID = "112569a2-315a-47b7-be57-f7f3b87734f8"
+TEST_USERNAME = "refiner"
+TEST_EMAIL = "refiner@refiner.com"
+TEST_USER_ID = "673da667-6f92-4a50-a40d-f44c5bc6a2d8"
```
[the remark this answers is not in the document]
robertmitchellv: when it wasn't working at one point i thought: _hmm maybe i should add the user we normally use and then it'll work?_ i think this was at midnight or so

refiner/app/api/v1/demo.py
robertmitchellv: so i mentioned in another thread but i updated these a bit and i think the overall output is cleaner and more type safe

refiner/app/services/testing.py
robertmitchellv: this is done now

refiner/app/api/v1/configurations.py
jakewheeler: Sorry 😭 I didn't realize the implications of this at the time of my comment.

refiner/app/api/v1/configurations.py
```diff
@@ -1004,51 +1005,58 @@ async def run_configuration_test(
...
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail=f"{associated_condition.display_name} was not detected as a reportable condition in the eCR file you uploaded.",
-        )
+    # 3. Call the new `refine_eicr` synchronously with the correct arguments.
+    refined_eicr_str = refine_eicr(
+        xml_files=original_xml_files,
+        processed_configuration=processed_config,
+    )
 
+    # TODO: will be worked on in a future ticket
```
jakewheeler, line 1023: 👍 I was confused when this route worked on a config I didn't expect it to but this makes sense to me.

PR #432, opened 2025-10-03
WIP: Configuration based inline testing 

### LINKED ISSUE #345: [TASK] (Server) Inline Testing Implementation
## 🚧 What needs to be done

Implement inline testing with configurations, allowing users to test configuration changes in real-time during the editing process.

## 🤔 Why it needs to be done

- Users need immediate feedback when building and modifying configurations
- Inline testing completes the configuration workflow user experience
- Provides validation before configurations are saved/activated/etc

## 🏗️ Acceptance Criteria

- [ ] Real-time testing of configuration changes without saving (need more info from PRD on this one. the question is: "what exactly are we testing inline? a draft? something not yet saved? something already saved?)
  - [ ] Clear distinction between draft/in-progress and saved configurations
- [ ] Integration with existing `ProcessedConfiguration` infrastructure
- [ ] User can test partial/incomplete configurations (this is also worth getting feedback from PRD on)

## ⚙️ Technical Notes

> [!NOTE]
> I'm still on the fence about whether or not this ticket or #344 should be completed first. would be good to get others' thoughts on this

robertmitchellv:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #345 
- #345


refiner/app/services/testing.py
```diff
@@ -249,3 +425,28 @@ async def _map_conditions_to_configurations(
...
+    Before we're able to detrmine reportability we'll need to get some data from the database to check
+    against the reportability_data results.
+
+    Returns a tuple of (configuration, condition).
+    """
+
+    configuration = await get_configuration_by_id_db(
+        id=configuration_id, jurisdiction_id=jurisdiction_id, db=db
+    )
+    assert configuration is not None, "Configuration not found, but was expected."
```
robertmitchellv, line 445: i'm not sure how acceptable this is as a way to deal with situations where mypy isn't sure that these things will exist, but where the user wouldn't be able to start this process (`inline_testing` workflow) without a `DbConfiguration` and `DbCondition`'s `conditions.id` <-> `configurations.condition_id` definitely existing and being accessible via their own pks.

refiner/app/services/testing.py
```diff
@@ -169,17 +212,141 @@ async def independent_testing(
...
+        "configuration_does_not_match_conditions": None,
+    }
 
 
 # NOTE:
 # PRIVATE FUNCTIONS
 # =============================================================================
 
 
+class ReportabilityData(TypedDict):
```
robertmitchellv, line 337: ended up having to just sketch this model out to solve mypy issues

jakewheeler (changes requested):
Looking good! I added some notes regarding the structure that I think will help with a few of the issues we're hitting. Let me know if you have any questions!

refiner/app/services/testing.py
robertmitchellv: yeah, this is probably the better way to approach this--i can't imagine how these things wouldn't be present if the app is working but it's probably better to just add a check anyway

robertmitchellv: db param  is now not being passed to `inline_testing`; i don't think this will work exactly in the same way for `independent_testing` but maybe we can resolve that later

PR #443, opened 2025-10-06
chore: introduce `react/jsx-curly-brace-presence` ESLint rule

jakewheeler:
# 🔀 PULL REQUEST

This PR adds an ESLint rule to enforce not using curly braces around string values in props. I noticed a few of these floating around and recalled that [we used this rule on Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/10db9269f5e5e04aceba5bd6e7297a7fffce09f1/containers/ecr-viewer/.eslintrc.json#L90) to help keep the code consistent.


rogeruiz (approved):
lgtm 🌈

PR #476, opened 2025-10-15
docs: update instructions on deploying to demo using GitHub action 

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Updates deployment instructions for using the Deploy to Demo action. Thanks to @alismx for doing all the hard work here!

For now, we need one manual step within Azure as documented until DevOps can do some more work to the global DIBBs infra. **As a result, the deploy action will result in demo displaying a 502 until we manually update the relevant IP address in Azure, so use this with caution.**

## 🔗 Related Issue

Fixes [#490](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/zh/490)
- [#490](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/zh/490) 

Made [a followup](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/477) to fully automate the deploy process once DevOps finishes the work they need.

## 🧪 How to test
- Run the new deploy job (can optionally also do the "plan" version that won't run the apply step
- If testing _in demo__ :grimace:, have a cut of main that's different than the version deployed in demo. Run the job, and notice that the action completes successfully and updates the latest cut of main.


refiner/README.md
```diff
@@ -41,6 +41,25 @@ To build the Docker image for the Message Refiner from source instead of downloa
...
 
+## Deploying to Demo
+
+We have a [GitHub action](https://github.com/CDCgov/dibbs-ecr-refiner/actions/workflows/trigger_demo_workflow.yaml) to deploy to our demo environment on demand whenever a new release is ready to go out. **Doing so will cause the demo site to temporarily go down / require a manual step to get it back up, so deploy using this method only when a few seconds of downtime is acceptable.**
+
+### Prerequiste permissions
+
+- Access/permissions to run actions in the Skylight infra repo.
+- Access to the Skylight Azure environment and the resources within the Refiner container instances / DIBBs app hub.
+- Ability to run actions in the Refiner repo.
```
rogeruiz: We should add here how to get this access. Something generic even would be helpful in case folks don't have the access that's required.

refiner/README.md
```diff
@@ -41,6 +41,25 @@ To build the Docker image for the Message Refiner from source instead of downloa
...
+
+### Instructions
```
jakewheeler: Do we want to add something about running migrations in here too? 

I don't think it needs to be detailed since I think there's some process improvement we need to do there so that it's simpler to run but I think a step saying `Run migrations to make sure the database schema is up to date` could save someone from being confused about why the site is down 😂

jakewheeler:
> It seems though that I weight the annoyance of changing things in UI more than others on the team, so happy to work on it if others would find it useful!

I'm good with this if the automation is right around the corner 😆 

> Ahh didn't realize this! Would the approach that SimpleReport took that spun up a separate container to run migrations against the DB allow us to sequence this into our deploy pipeline automatically? Or would we need to do something else?
>
> Think even if we just stick with on-demand deploys, having migrations folded in so that it's truly as close to "click this button" as we can would be valuable.

Yes, I think SR's approach would work well for us and it seemed like @alismx thought so as well. Not totally sure off the top of my head what the level of effort to get this working would be but I 1000% agree that "click this button" deployments would be immensely valuable! 

We would be able to fully automate deployments at this point and there would be no real reason in my mind why we couldn't go down the "deploy on merge to main" path.

PR #508, opened 2025-10-20
feat: Add XML -> HTML conversion

### LINKED ISSUE #397: [USER STORY] Generate Refined HTML file(s)
## 👤💭 User Story

As a user, I want to view the HTML output for my files, so that I can know whether it'll work with my downstream processes

## 🏗️ Acceptance Criteria

- [ ] File download includes an HTML version of the refined output XML(s)

## 🎨 Design Notes
**No additional designs needed**

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch is missing the correct stylesheet. Waiting on APHL to submit theirs to replace the file at `refiner/app/xslt/cda2html.xsl`.

## 🔗 Related Issue

- Closes #397

## ✅ Acceptance Criteria

- [x] File download includes an HTML version of the input and refined output XML(s)
- [x] Need XSLT 3.0 processor
  - XSLT 1.0 processor may be enough for our needs.
- [x] Host 3rd-party libraries and stylesheets statically from Refiner to avoid cross-origin dependency linking. Pull 3rd-party vendor files from public CDNs.
  - jQuery, jQuery-UI, and Bootstrap CSS & JS.

## 🧪 How to test

Go to the [http://localhost:8081/testing](http://localhost:8081/testing) in your
browser and click on **Download results** to download a ZIP containing the HTML
and XML files. Loading up the HTML locally should render it properly with UX
that allows scroll-to functionality for the different sections in the HTML
document. The template was copied from [HL7/CDA-phcaserpt-1.1.1][repo1] and
[HL7/CDA-phcaserpt][repo2]. We're only using the `1.1.1` version for now. But
both templates have been modified to leverage network requests for the 3rd-party
libraries that they depend on. For now, the `1.1.1` version is hard-coded with a
`TODO` for how we may update things in the future.

[repo1]: https://github.com/HL7/CDA-phcaserpt-1.1.1/
[repo2]: https://github.com/HL7/CDA-phcaserpt/

### DIFF
--- a/refiner/tests/test_xslt.py
+++ b/refiner/tests/test_xslt.py
@@ -1,7 +1,5 @@
 import logging
-import os
 import tempfile
-from pathlib import Path
 
 import pytest
 
@@ -48,29 +46,3 @@ def test_transform_xml_to_html_malformed_xslt() -> None:
             transform_xml_to_html(
                 VALID_CDA_XML, xslt_file.name, logging.getLogger("xslt-test")
             )
-
-
-def test_transform_xml_to_html_xslt_outside_trusted_directory() -> None:
-    """
-    Should raise XSLTTransformationError if xslt file is outside trusted directory.
-    """
-    VALID_CDA_XML = b"""
-    <ClinicalDocument xmlns="urn:hl7-org:v3">
-      <recordTarget></recordTarget>
-    </ClinicalDocument>
-    """
-    with tempfile.NamedTemporaryFile(
-        suffix=".xslt", dir=tempfile.gettempdir(), delete=False
-    ) as xslt_file:
-        xslt_file.write(MINIMAL_XSLT)
-        xslt_file.flush()
-        xslt_file_path = Path(xslt_file.name)
-    try:
-        # This file is definitely not under the assets/xslt trusted dir.
-        with pytest.raises(XSLTTransformationError) as exc_info:
-            transform_xml_to_html(
-                VALID_CDA_XML, xslt_file_path, logging.getLogger("xslt-test")
-            )
-        assert "outside trusted directory" in str(exc_info.value)
-    finally:
-        os.unlink(xslt_file_path)


refiner/app/api/v1/demo.py
```diff
@@ -137,6 +138,149 @@ async def demo_upload(
...
+                raise HTTPException(
+                    status_code=status.HTTP_502_BAD_GATEWAY,
+                    detail=["Failed to upload ZIP to S3. Please try again later."],
+                )
+        except Exception as e:
+            logger.error("Exception during S3 upload", extra={"error": str(e)})
+            raise HTTPException(
+                status_code=status.HTTP_502_BAD_GATEWAY,
+                detail=["Failed to upload ZIP to S3. Please try again later."],
+            )
```
jakewheeler: As mentioned, I don't think we should do this - but as a side note: when we throw an `HTTPException` in a handler FastAPI takes care of it and returns it to the client.

The way this is set up currently is that if `presigned_s3_url` is empty, we throw `502 Bad Gateway`, and then we catch it here and throw the same thing again. We don't need to do that since FastAPI will do the work for us.

refiner/app/api/v1/demo.py
```diff
@@ -137,6 +138,149 @@ async def demo_upload(
...
+    except FileProcessingError as e:
+        logger.error("FileProcessingError", extra={"error": str(e)})
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=FILE_PROCESSING_ERROR,
+        )
+    except (
+        HTTPException
+    ):  # Returns the raw HTTPException without modification (502 from S3 upload)
+        raise
```
jakewheeler: This is pretty different from [what's in `main`](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/api/v1/demo.py#L111) due to some recent changes. I would suggest taking a looking and going with what we have there since we made changes to this as recently as yesterday afternoon.

refiner/app/db/demo/test_model.py
```diff
@@ -0,0 +1,20 @@
+from .model import IndependentTestUploadResponse
```
jakewheeler: All `test_*` files need to live in the `refiner/tests` directory.

refiner/app/api/v1/demo.py
```diff
@@ -137,6 +138,149 @@ async def demo_upload(
...
+            condition_obj = refined_document.reportable_condition
+            condition_refined_eicr = format.normalize_xml(refined_document.refined_eicr)
+
+            condition_code = condition_obj.code
+            condition_name = condition_obj.display_name
+
+            filename_xml = file_io.create_split_condition_filename(
+                condition_name=condition_name, condition_code=condition_code
+            )
+            filename_html = filename_xml.replace(".xml", ".html")
```
jakewheeler: Nice 👍

rogeruiz:
Thanks @jakewheeler for the thorough feedback! I've updated things to match and
should have fixed the `demo.py` endpoint merge conflict and my additions that
weren't needed in the first place.

refiner/app/api/v1/demo.py
```diff
@@ -78,145 +79,193 @@ async def demo_upload(
 
     Any exceptions during file processing or workflow execution are caught and mapped to HTTP errors.
     """
+    try:
```
jakewheeler: I don't think we need to wrap this whole function in a try/except. We should be covered without this.

refiner/app/api/v1/demo.py
```diff
@@ -78,145 +79,193 @@ async def demo_upload(
...
         )
+    except HTTPException:
+        raise
     except Exception as e:
-        logger.error("Error in the independent testing flow", extra={"error": str(e)})
+        logger.error("Unhandled error in demo_upload", extra={"error": str(e)})
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail=GENERIC_SERVER_ERROR,
         )
```
jakewheeler: I believe this is not necessary.

refiner/app/services/xslt.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def transform_xml_to_html(xml_bytes: bytes, xslt_path: Path, logger: Logger) -> bytes:
+    """
+    Transforms CDA XML to HTML using the specified XSLT stylesheet.
+
+    Args:
+        xml_bytes (bytes): The raw XML document bytes.
+        xslt_path (Path): Path to the XSLT stylesheet file.
+        logger (Logger): Logger for logging errors and debug information.
```
jakewheeler: Thank you 🙏

refiner/app/db/demo/model.py
```diff
@@ -11,12 +11,15 @@ class Condition:
     display_name: str
     refined_eicr: str
     stats: list[str]
+    # Optionally: html_files for per-condition HTML
```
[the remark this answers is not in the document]
rogeruiz: oh dang. sorry! thought i got these

refiner/app/api/v1/demo.py
```diff
@@ -78,145 +79,193 @@ async def demo_upload(
...
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail="ZIP archive cannot be read. CDA_eICR.xml and CDA_RR.xml files must be present.",
+        # STEP 4:
+        # for each unique reportable condition code found in the RR (with a config),
+        # build a refined XML and collect metadata. The code used is from the RR.
+        conditions: list[Condition] = []
+        refined_files_to_zip = []
+        xslt_stylesheet_path = file_io.get_asset_path(
+            "xslt",
+            "CDA-phcaserpt-1.1.1-CDAR2_eCR_eICR.xsl",
```
jakewheeler: This is still not quite what I had in mind since we aren't passing in the path with `Depends`. To give a better example:

In `xslt.py` we may define this function:
```python
# Something along these lines
def get_path_to_xslt_stylesheet() -> Path:
    """Returns the path to the eICR XSLT stylesheet."""

    return file_io.get_asset_path("xslt", "CDA-phcaserpt-1.1.1-CDAR2_eCR_eICR.xsl")
```
This function would be imported into `demo.py` and `demo_upload`'s signature would be updated to dependency inject this path into the function:
```python
# up top
from ...services.xslt import get_path_to_xslt_stylesheet
...
...
..
async def demo_upload(
    uploaded_file: UploadFile | None = File(None),
    demo_zip_path: Path = Depends(get_sample_zip_path),
    xslt_stylesheet_path: Path = Depends(get_path_to_xslt_stylesheet), # <-- new line, path is ready to go
    ....
```
and this line I highlighted can be deleted since we've already acquired the path. Does that help?

PR #511, opened 2025-10-20
feat(server): hook in section processing data from configurations to power refine_eicr

### LINKED ISSUE #445: [TASK] hook the section processing data from configurations into the refine_eicr workflows for both independent and inline testing
## 🚧 What needs to be done

This task will involved modifying the two ways that refine_eicr has parameters feed to it; both inline and independent testing. The `refine_eicr` function will need to be updated to work with the section processing instructions. Currently 2/3 sort of de-facto exist but the "remove"; or, force a minimal section will need to be created and hooked into the `process_eicr` script.

## 🤔 Why it needs to be done

Currently the `sections_to_include` argument to `refine_eicr` hasn't been truly connected to the refiner application. This will work will actually hook the data that is now being collected in the configuration and use it in the refining process.

## 🏗️ Acceptance Criteria

- [ ] Both workflows in `testing.py` are modified to unpack and package the section processing instructions for `refine_eicr`
- [ ] `refine_eicr` is updated to work with either a single or multiple arguments for the different instructions
- [ ] a `_force_minimal_section` function will need to be created
- [ ] the `_process_section` function will need some updated control flow to work with this new function
- [ ] test that, end to end, the results produce what is expected of refine, retain, remove

## ⚙️ Technical Notes

- ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This takes the `configurations.section_processing` data from that column of the `configurations` table to power a new model `RefinementPlan` that becomes the final set of instructions for `refine_eicr`.

In addition to providing processing instructions for the sections that are required by the eICR 1.1 spec, we're also scanning the eICR for additional sections that, although not strictly required by the eICR spec may be present in the document anyway and giving it a default section processing rule of "refine". This choice was made based on the default for everything being "refine" unless the user overrides this. In the future we can potentially make this configurable, but it will keep large sections that aren't configured with instructions to be refined for relevant data or filtered down to a minimal section.

## 🔗 Related Issue

Fixes #445 
- #445 

## ✅ Acceptance Criteria

- [x] Both workflows in `testing.py` are modified to unpack and package the section processing instructions for `refine_eicr`
- [x] `refine_eicr` is updated to work with either a single or multiple arguments for the different instructions
- [x] a `_force_minimal_section` function will need to be created
- [x] the `_process_section` function will need some updated control flow to work with this new function
- [x] test that, end to end, the results produce what is expected of refine, retain, remove

## 🧪 How to test

Testing the application:

1. Run `docker compose up`
2. Run `just db refresh`
3. Check that all services are up and running with `docker ps`
4. In your browser, navigate to [http://localhost:8081](http://localhost:8081) to test the application
5. You should be able to create your own section processing rules--i would say pick "retain" for all sections of one condition and "remove" for all sections of the other condition. The diff will be clear. You can also mix and match and check it to see.

## ℹ️ Additional Information

Another thing that was added in this PR was allowing for rootless docker to work with our docker compose set up. small changes to env vars and how they're set in auth and s3 were also added.

### DIFF
--- a/refiner/app/services/aws/s3.py
+++ b/refiner/app/services/aws/s3.py
@@ -14,6 +14,9 @@ uploaded_artifact_bucket_name = ENVIRONMENT["S3_UPLOADED_FILES_BUCKET_NAME"]
 
 config = Config(signature_version="s3v4")
 
+# S3_ENDPOINT_URL is **only** used in local dev; so we grab that
+# via os.getenv rather than the ENVIRONMENT we import from core.config
+# this is so we can connect to localstack
 s3_client = boto3.client(
     "s3",
     region_name=ENVIRONMENT["AWS_REGION"],


refiner/app/api/auth/config.py
```diff
@@ -9,7 +9,16 @@
...
+    # FOR THE BROWSER:
+    # this is the url the user is redirected to for logging in
+    # It **must** be the public url
+    authorization_endpoint=f"{ENVIRONMENT['AUTH_ISSUER']}/protocol/openid-connect/auth",
+    # FOR THE BACKEND:
+    # these are for server-to-server communication
+    # they **must** be the internal urls
+    token_endpoint=f"{ENVIRONMENT['AUTH_ISSUER_INTERNAL']}/protocol/openid-connect/token",
+    userinfo_endpoint=f"{ENVIRONMENT['AUTH_ISSUER_INTERNAL']}/protocol/openid-connect/userinfo",
+    jwks_uri=f"{ENVIRONMENT['AUTH_ISSUER_INTERNAL']}/protocol/openid-connect/certs",
```
jakewheeler: `AUTH_ISSUER_INTERNAL` won't be available in a live environment. Would this cause auth to fail?

robertmitchellv: broke this out into a, dev do this, else do this based on the convo we had earlier and this should fix the concern

refiner/app/core/config.py
```diff
@@ -44,9 +44,11 @@ def _get_env_variable(name: str) -> str:
     "AUTH_CLIENT_ID": _get_env_variable("AUTH_CLIENT_ID"),
     "AUTH_CLIENT_SECRET": _get_env_variable("AUTH_CLIENT_SECRET"),
     "AUTH_ISSUER": _get_env_variable("AUTH_ISSUER"),
+    "AUTH_ISSUER_INTERNAL": _get_env_variable("AUTH_ISSUER_INTERNAL"),
```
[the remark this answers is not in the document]
robertmitchellv: removed the ones that are local only from `core/config.py` and am using `os.getenv` for them based on conditional logic in the two places they're needed:
- auth
- s3

refiner/app/services/terminology.py
```diff
@@ -5,110 +5,15 @@
...
 # =============================================================================
 
 
-# NOTE:
-# CONDITION PROCESSING
-# =============================================================================
-
-
-@dataclass(frozen=True)
-class ConditionPayload:
```
robertmitchellv: these aren't being used anywhere for any reason so i pulled them out rather than let them sort of atrophy in our repo

refiner/app/services/ecr/process_eicr.py
```diff
@@ -33,10 +38,88 @@
...
+        XMLParsingError: If XPath evaluation fails
+    """
+
+    try:
+        xpath_query = f'.//hl7:section[hl7:code[@code="{loinc_code}"]]'
+        xpath_result = structured_body.xpath(xpath_query, namespaces=namespaces)
+
+        if isinstance(xpath_result, list) and len(xpath_result) == 1:
+            section = cast(list[_Element], xpath_result)
+            return section[0]
```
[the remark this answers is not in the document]
robertmitchellv: i know that validation happens on the AIMS platform, i'm assuming before RCKMS, but it will definitely happen before the refiner works on either documents. so we can safely assume that the main required sections will be present (otherwise validation would fail). at the very least we expect the sections that we provide processing instructions for in the configuration. i expect that we'll generally have all required plus one or two other ones on average. i did an analysis of this on sample data i can share on slack too

refiner/app/api/auth/config.py
```diff
@@ -1,17 +1,38 @@
...
+        client_kwargs={"scope": "openid email profile"},
+    )
+else:
+    _oauth.register(
+        name=ENVIRONMENT["AUTH_PROVIDER"],
+        client_id=ENVIRONMENT["AUTH_CLIENT_ID"],
+        client_secret=ENVIRONMENT["AUTH_CLIENT_SECRET"],
+        server_metadata_url=f"{ENVIRONMENT['AUTH_ISSUER']}/.well-known/openid-configuration",
+        client_kwargs={"scope": "openid email profile"},
+    )
```
jakewheeler: I don't anticipate anything breaking here but we'll want to test this in the demo environment as soon as we can once this is merged.

refiner/app/services/aws/s3.py
```diff
@@ -54,6 +59,12 @@ def upload_refined_ecr(
             ExpiresIn=expires,
         )
 
+        # for local dev, boto3 creates a URL with the internal hostname ('localstack')
+        # We must replace it with the public hostname ('localhost') before sending it
+        # to the browser
+        if ENVIRONMENT["ENV"] == "local":
+            presigned_url = presigned_url.replace("localstack:4566", "localhost:4566")
+
```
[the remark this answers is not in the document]
robertmitchellv: just so everyone else is aware, Jake and i talked offline about my initial proposed solution, which created an internal and external s3 client. we agreed that instead we'd only have a single `S3_ENDPOINT_URL` that uses the `localstack:port` format and that if we're in a "local" environment that we'll replace "localstack" with "localhost" rather than maintain the two s3 clients. we can always revisit this later as well when we know more about how deploys will work on AWSs via AIMS

PR #513, opened 2025-10-21
feat: add other system option to custom code

### LINKED ISSUE #503: [TASK] Add "Other" to custom code modal "Code System" drop-down
We should have an "other" option for code system in order to not limit JDs in what custom codes they can add.

lina-roth:
# 🔀 PULL REQUEST

Screenshot 2025-10-21 at 2 12 05 PM
Screenshot 2025-10-21 at 2 21 07 PM
Screenshot 2025-10-21 at 2 21 33 PM

## 💡 Summary

- Adding 'other' option to custom code system select 
- Adding 'Select system' as default option 
- Adding check to ensure cannot be created with default option 'Select System'

## 🔗 Related Issue

Fixes #503 
- #503 

## ✅ Acceptance Criteria

- [ ] We should have an "other" option for code system in order to not limit JDs in what custom codes they can add.

## 🧪 How to test

1. Run `docker compose up`
4. You should be able to create a configuration
5. When adding a custom code you should see 'Select system' as the default option
6. You should be able to add, edit, delete a custom code with 'OTHER' selected as the system


client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -733,17 +733,19 @@ export function CustomCodeModal({
   // TODO: this should come from the server.
   // Maybe get this info as part of the seed script?
   const systemValues = [
+    { name: 'Select system', value: '' },
     { name: 'ICD-10', value: 'icd-10' },
     { name: 'SNOMED', value: 'snomed' },
     { name: 'RxNorm', value: 'rxnorm' },
     { name: 'LOINC', value: 'loinc' },
+    { name: 'OTHER', value: 'other' },
```
jakewheeler: Let's change this so it's consistent with other places we display it.
```suggestion
    { name: 'Other', value: 'other' },
```

For example
image

PR #516, opened 2025-10-22
fix: align  inline and independent testing reduction percentages

### LINKED ISSUE #505: [BUG] Size Reduction metric differs between in-line and independent testing for the same test file
For the same test file and the same configuration, the size reduction metric output is different between in-line and independent testing flows

To reproduce:
- Create default COVID configuration if not already created
- Run in-line testing with test file (36%)
- Run independent testing with test file (44%)

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Tweaks the way we're processing files fed into the reduction percentage calculation in order to get independent and inline reduction percentages to match. Adds some tests to make sure things match with the test file.

## 🔗 Related Issue

Fixes #505 and #521
- #505 
- #521

## ✅ Acceptance Criteria
- [ ] Percentages between the inline and independent testing reduction outputs should match 

## 🧪 How to test
- Run the test file between the two flows of the app
- Check that the outputted percentages match

### DIFF
--- a/refiner/app/core/config.py
+++ b/refiner/app/core/config.py
@@ -47,6 +47,5 @@ ENVIRONMENT: dict[str, str] = {
     "AWS_ACCESS_KEY_ID": _get_env_variable("AWS_ACCESS_KEY_ID"),
     "AWS_SECRET_ACCESS_KEY": _get_env_variable("AWS_SECRET_ACCESS_KEY"),
     "AWS_REGION": _get_env_variable("AWS_REGION"),
-    "S3_ENDPOINT_URL": _get_env_variable("S3_ENDPOINT_URL"),
     "S3_UPLOADED_FILES_BUCKET_NAME": _get_env_variable("S3_UPLOADED_FILES_BUCKET_NAME"),
 }
--- a/refiner/app/services/aws/s3.py
+++ b/refiner/app/services/aws/s3.py
@@ -19,7 +19,9 @@ s3_client = boto3.client(
     region_name=ENVIRONMENT["AWS_REGION"],
     aws_access_key_id=ENVIRONMENT["AWS_ACCESS_KEY_ID"],
     aws_secret_access_key=ENVIRONMENT["AWS_SECRET_ACCESS_KEY"],
-    endpoint_url=os.getenv("S3_ENDPOINT_URL"),
+    endpoint_url=os.getenv(
+        "S3_ENDPOINT_URL"
+    ),  # Only needed for local dev to connect to localstack
     config=config,
 )
 


client/src/pages/Testing/RunTest.tsx
```diff
@@ -63,6 +63,7 @@ export function RunTest({
         <span>
           To download test files for some conditions you can
           <a
+            target="_blank"
```
fzhao99: decided to add this one liner to [fix this](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/521)

refiner/app/api/v1/configurations.py
```diff
@@ -1151,21 +1151,26 @@ async def run_configuration_test(
 
     # STEP 5:
     # construct and return the final response
-    original_unrefined_eicr = strip_comments(normalize_xml(original_xml_files.eicr))
-    matched_condition_refined_eicr = strip_comments(normalize_xml(refined_eicr_str))
+    stripped_unrefined_eicr = strip_comments(original_xml_files.eicr)
+    normalized_stripped_original_ecr = normalize_xml(stripped_unrefined_eicr)
```
jakewheeler: I'm finding this naming a little confusing since we go from calling it `unrefined` to `original` (and also `ecr` but I think that's probably a typo ). We should pick one and stick with it.

Also, any reason not to do the stripping and normalizing in one line? I think it was probably that way before since we don't actually make use of the stripped output on its own anywhere. If we want to keep it as is that's also fine. Just two things to name instead of one 😂

refiner/app/api/v1/demo.py
```diff
@@ -129,21 +129,11 @@ async def demo_upload(
...
-            xml_files=original_xml_files,
-            jurisdiction_id=jd,
-        )
-    except Exception as e:
-        logger.error("Error in the independent testing flow", extra={"error": str(e)})
-        raise HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail=GENERIC_SERVER_ERROR,
-        )
-
```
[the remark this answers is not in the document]
jakewheeler: 🤣 no problem!

refiner/app/api/v1/configurations.py
```diff
@@ -1151,21 +1150,23 @@ async def run_configuration_test(
 
     # STEP 5:
     # construct and return the final response
-    original_unrefined_eicr = strip_comments(normalize_xml(original_xml_files.eicr))
-    matched_condition_refined_eicr = strip_comments(normalize_xml(refined_eicr_str))
+    formatted_unrefined_eicr = strip_comments(normalize_xml(original_xml_files.eicr))
+    formatted_refined_eicr = strip_comments(
+        normalize_xml(refined_document.refined_eicr)
```
jakewheeler: 🎯

refiner/app/services/testing.py
```diff
@@ -193,7 +193,7 @@ async def independent_testing(
         sections_to_include = None
 
         refined_eicr_str = refine_eicr(
-            xml_files=XMLFiles(xml_files.eicr, xml_files.rr),
+            xml_files=xml_files,
```
jakewheeler: 🧠

PR #523, opened 2025-10-23
fix: do not require `S3_ENDPOINT_URL` environment variable to be present

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes an issue where the app was requiring `S3_ENDPOINT_URL` to be present in the environment in a non-local dev scenario. There is no need for it to be present in live environments.

## 🧪 How to test

- Shutting down and restarting your docker compose setup locally should function exactly the same
- S3 downloads in the demo environment should be unchanged (need to deploy the change to test this)


rogeruiz (approved):
lgtm 🌈

Issue #422, opened 2025-10-02
[USER STORY] Activity Log - Creation


kplaneaux:
## 👤💭 User Story

As a user, I want to be able to review user activity, so that I can understand what changes were made to the application, by who, and when

## 🏗️ Acceptance Criteria

- [ ] Table displays a list of all "Created configuration" events
- [ ] Table rows include desired info
  - [ ] User name
  - [ ] Condition name
  - [ ] Event action description
  - [ ] Date (datetime log was created)
- [ ] Table is ordered newest to oldest by date

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=3324-10985&t=JQksXunTOIBhdN1O-1
<img width="1125" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNk8yQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a1488d15b6ef89751e948e1aedb35a5e2a3dac82/image.png" alt="image.png" />

PR #556, opened 2025-10-28
feat(ux): add activity log frontend

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Frontend part of the activity log that connects the basic page skeleton up to the backend via the events hook. 

## 🔗 Related Issue
Part of #422 
- #422 that will hook into @jakewheeler's work on the backend. Stubs out an assumed data array for now

## ✅ Acceptance Criteria
- [ ] Table displays a list of all "Created configuration" events
- [ ] Table rows include desired info
  - [ ] User name
  - [ ] Condition name
  - [ ] Event action description
  - [ ] Date (datetime log was created)
- [ ] Table is ordered newest to oldest by date

## 🧪 How to test
- Pull up the app locally
- Navigate to the activity log page
- Notice that the table renders with stubbed data

### DIFF
--- a/refiner/app/db/configurations/db.py
+++ b/refiner/app/db/configurations/db.py
@@ -115,7 +115,7 @@ async def insert_configuration_db(
                     user_id=user_id,
                     configuration_id=config.id,
                     event_type="create_configuration",
-                    action_text="Created configuration.",
+                    action_text="Created configuration",
                 ),
                 cursor=cur,
             )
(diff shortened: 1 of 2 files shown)


client/src/pages/ActivityLog/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    hour: 'numeric',
+    minute: 'numeric',
+    hour12: true,
+  });
+
+  if (isPending) return <Spinner variant="centered" />;
+  if (isError) return <ErrorFallback error={error} />;
+
+  return (
+    <section className="mx-auto">
```
jakewheeler: This looks a bit tight at some (even larger) screen widths. Here's what I'm seeing at 1200px width:
image

Looks like on other pages we add padding but I'm noticing here this causes its own issue. For example, adding `p-3` to the `section` makes the table look like this (775px width):
image

 Any ideas? Are we able to make the table more compact when the screen shrinks?

PR #562, opened 2025-10-30
docs(build): Introduce RFC/Decisions with Just

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch adds new automation around authoring decision records for the
project. Using for example, `just rfc new "Title of decision record"`, you can
create new decision records that can be displayed in either Structurizr Lite or
GitHub.

## 🔗 Related Issue

Related to the work done in #571 to be able to share decisions with the team.

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

You can test this out by running `just decisions new` with a title as an
argument. This should create a new decision record in the `docs/decision/`
directory with the contents from the `.template` found in that directory.

### DIFF
--- a/docs/decisions/.template
+++ b/docs/decisions/.template
@@ -4,7 +4,7 @@ Date: ${today}
 
 ## Status
 
-Proposed
+Proposed/Accepted/Superseded
 
 ## Context and Problem Statement
 


jakewheeler:
Am I missing a step? Is there supposed to be a `.template` file present already?

This is my output:
```
📝 Creating a new decision record on 2025-11-04
🔍 Found decision record title: test
🔦 Checking for existing decision records in /dibbs-ecr-refiner/docs/decisions
node:fs:561
  return binding.open(
                 ^

Error: ENOENT: no such file or directory, open '/dibbs-ecr-refiner/docs/decisions/.template'
    at Object.openSync (node:fs:561:18)
    at Object.readFileSync (node:fs:445:35)
    at Object.<anonymous> (/private/var/folders/b6/k0rrsktd7bggph42z089sj6w0000gn/T/just-JBkKuc/_new:123:21)
```

justfile
```diff
@@ -75,3 +82,71 @@ test:
...
+      .filter(isFile);
+
+    console.info(`🔍 Found ${(files.length + "").padStart(4, '0')} {{ type }}(s)`);
+
+    const nextNumber = files.length + 1;
+    const nextNumberString = (nextNumber + "").padStart(4, '0');
+
+    console.info(`🖊️ Setting your new {{ type }} to #${nextNumberString}`);
+
+    const nextFilePath = path.join(
```
jakewheeler: Any way we could include the date in the path for convenience?

No strong opinions on formatting. Something like this could work?
`0003_2025-11-04_name.md`

docs/decisions/.template
```diff
@@ -0,0 +1,27 @@
+# ${nextNumber}. ${title}
+
+Date: ${today}
+
+## Status
+
+Proposed
```
jakewheeler: What are the expected statuses we'd add here?

rogeruiz: Ah yeah good question. I think either `Proposed`, `Accepted`, and `Superseded` would be enough for now. I'm open to using other statuses that may make more sense. But I believe these three are good defaults.

justfile
rogeruiz: Sure! I'll add this shortly.

rogeruiz: Added in 6f65ef6.

PR #615, opened 2025-11-10
fix: Bring in changes to testing upgrades

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

I'm not entirely sure right now how this happened. But I'll look into it. The
two errors I was encountering are over here: https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/19241807464/job/55006085326?pr=604#step:4:538

**Error 1**

```
------
6.114 playwright.config.ts(10,17): error TS2591: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
6.114 playwright.config.ts(11,12): error TS2591: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node` and then add 'node' to the types field in your tsconfig.
```

**Error 2**

```
6.843 vite.config.ts(14,3): error TS2769: No overload matches this call.
6.843   The last overload gave the following error.
6.843     Object literal may only specify known properties, and 'test' does not exist in type 'UserConfigExport'.
------
```

I may cherry-pick these into a new PR so it doesn't force this PR to get merged
in just to fix `app-build`.

## 🔗 Related Issue

- Found in #604

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

GitHub Actions should be all passing.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -34,6 +34,7 @@
         "@testing-library/jest-dom": "^6.9.1",
         "@testing-library/react": "^16.3.0",
         "@testing-library/user-event": "^14.6.1",
+        "@types/node": "^24.10.0",
         "@types/react": "^19.2.2",
         "@types/react-dom": "^19.2.2",
         "@typescript-eslint/parser": "^8.43.0",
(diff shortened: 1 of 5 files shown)


jakewheeler (changes requested):
As mentioned I think we'll want to close this in favor of #616.

PR #633, opened 2025-11-20
swap activity log condition id with canonical url

### LINKED ISSUE #630: Replace condition ID references to conditions with canonical URL condition groupers
As part of the activation work, we're beginning to refer to conditions via their canonical URL that references the configuration grouper rather than the table ID's that are different for each condition version. 

We need to update all the places we're doing this check with the new values in order to have the places where the application only expects one condition to function correctly.

Preemptively breaking this out into another ticket for the last AC in [this ticket](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/308), but feel free to simul-close if the work is small enough

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Fast follow to the activations work that dedupes options on the activity log page by indexing by canonical URL rather than condition ID

## 🔗 Related Issue

Fixes #630 
- #630 

## ✅ Acceptance Criteria

- [ ] Activity log page filter options remain de-duped
## 🧪 How to test

- Check the activity log page
- Ensure the dropdown options work as before


refiner/app/api/v1/events.py
```diff
@@ -42,15 +40,23 @@ async def get_events(
...
         jurisdiction_id=user.jurisdiction_id, db=db
     )
-    configuration_options = [
-        ConfigurationTrace(name=c.name, id=c.id) for c in jd_configurations
-    ]
+
+    seen_urls = set()
+    configuration_options = []
+
+    for c in jd_configurations:
```
fzhao99: I did this using list comprehension initially but mypy was complaining about some return type issues, so did this the generic loop way

jakewheeler (approved):
Seems to work exactly as I'd expect! Thank you 👍

PR #645, opened 2025-11-25
feat: activate configuration

### LINKED ISSUE #308: [USER STORY] Activate Configuration
## 👤💭 User Story

As a user, I want to activate my configuration, so that eCRs for that condition can start being refined in production.

## 🏗️ Acceptance Criteria

- [ ] Activation page w/ "Turn on configuration" or "Turn off configuration" button (Turn on if no active configuration, turn off if there is an active configuration)
- [ ] Turning on a configuration gives it an "active" status on the back-end (recognized by independent testing flow)

## Technical Implementation
- [ ] Configurations with status `active` and `inactive` should not be able to be updated any longer as a part of this work (check the config's status in API routes that update it). They will still appear as editable in the UI
- [x] We will be able to have multiple configs of the same type once users can activate configurations (a COVID-19 config in `draft` and one in `active`, for example).  This may cause issues where we currently create links using `condition_id` instead of `canonical_url`. Please make this change where applicable. One known instance of this issue can be seen with the activity log filter. If two COVID-19 configs exist, we'll have two separate entries for COVID-19 filters when we only want to show the user one
- [x] Two new routes will need to be added to `/api/v1/configurations.py`
  - [x] `POST /api/v1/configurations/:id/activate`, no body needed
  - [x] `POST /api/v1/configurations/:id/deactivate`, no body needed
  - [x] Both routes handle errors
  - [x] Both routes only allow status to change when conditions are met
- [x] Configuration's `status` should be used to decide which components display in the UI 
- [x] Client makes use of `useActivateConfiguration()` and `useDeactivateConfiguration()` mutations, client determines which is appropriate to use based off config properties

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=2901-9377&t=zA8rVCNKp1danm9l-1

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Frontend for activations

https://github.com/user-attachments/assets/cde46143-6146-4447-9da1-df1f07b38925

Also to try and make or e2e's more reliable, set up some login fixtures at the worker level in line with the guidance from Playwright. Hoping that even if it doesn't solve all the flakiness issues that the e2e's are more reliable than before.

## 🔗 Related Issue

Fixes #308 
- #308

## ✅ Acceptance Criteria
- [ ] Activation page w/ "Turn on configuration" or "Turn off configuration" button (Turn on if no active configuration, turn off if there is an active configuration)

## 🧪 How to test
- Create a configuration locally
- Navigate to the activation tab
- Notice that the activation options are present / modal renders correctly
- Navigate back to the build tab and create another version / draft of that configuration
- Navigate back to the activation tab
- Notice that the activation options have updated according to the new options

### DIFF
--- a/client/e2e/utils.ts
+++ b/client/e2e/utils.ts
@@ -21,8 +21,7 @@ export function refreshDatabase(): string {
 }
 
 export async function login(page: Page, baseUrl = '/') {
-  await page.goto(baseUrl);
-
+  await page.goto(baseUrl, { waitUntil: 'networkidle', timeout: 60000 });
   await page.getByText('Log in').click();
   await page
     .getByRole('textbox', { name: 'Username or email' })


client/e2e/add-condition.spec.ts
```diff
@@ -1,39 +1,12 @@
...
-    await page.getByTestId('combo-box-input').press('Tab');
-    await page.getByRole('option', { name: 'Acanthamoeba' }).press('Enter');
-    await page.getByTestId('combo-box-input').press('Tab');
-    await page.getByTestId('combo-box-clear-button').press('Tab');
-    await page.getByTestId('modalFooter').getByTestId('button').click();
-    await expect(
-      page.getByRole('heading', { name: 'New configuration created' })
-    ).toBeVisible();
-
+    await createNewConfiguration('Acanthamoeba', page);
```
fzhao99: after this gets in, going to wrap this into some sort of fixture that does the setup automatically, potentially at the worker level, so that the tests are more parallellizable. That way each of the tests can have a more ergonomic way of getting a clean config to be working off of

client/playwright.config.ts
```diff
@@ -4,11 +4,10 @@ export default defineConfig({
   testDir: './e2e',
   timeout: 30 * 1000,
   expect: {
-    timeout: 5000,
+    timeout: 10 * 1000,
   },
-  fullyParallel: true,
   forbidOnly: !!process.env.CI,
-  retries: process.env.CI ? 2 : 0,
+  workers: 1,
```
fzhao99: couldn't get the parallized workers setup working which I _think_ is a function of the fact that we're doing logins at the worker level. We should be able to bump this up as we get more worker accounts.

client/src/pages/Configurations/ConfigActivate/index.tsx
```diff
@@ -1,55 +1,189 @@
...
+            </ul>
+          </div>
+          <hr className="text-gray-cool-20!" />
+          <ActivationButtons
+            configurationData={configuration.data}
+            isSuccess={isSuccess}
+            handleActivation={handleActivation}
+            handleDeactivation={handleDeactivation}
+            curVersion={configuration.data.version}
+            activeVersion={configuration.data.active_version}
```
jakewheeler: I think it'd make the most sense to continue passing in either `configurationData` itself (the whole object) or only pass in the properties that we plan to use within `ActivationButtons` instead of the full object + a couple individual properties (`curVersion` and `activeVersion`).

client/src/pages/Configurations/ConfigActivate/index.tsx
```diff
@@ -1,55 +1,189 @@
...
+      });
+      return;
+    }
+    activate(
+      {
+        configurationId: configuration.data.id,
+      },
+      {
+        onSuccess: async () => {
+          await refetchConfigurations(configuration.data, queryClient);
```
jakewheeler: When you invalidate the query it should auto-refresh the data and make the page appear up to date. Did you run into any issues with this?
```suggestion
          await queryClient.invalidateQueries({
            queryKey: getGetConfigurationQueryKey(configuration.data.id),
          });
```

client/src/pages/Configurations/ConfigActivate/index.tsx
```diff
@@ -1,55 +1,189 @@
...
+    // from all_versions
+    const idToDeactivate = configuration?.data?.all_versions
+      .filter((v) => v.status === 'active')
+      .map((c) => c.id);
+
+    deactivate(
+      { configurationId: idToDeactivate[0] },
+      {
+        onSuccess: async () => {
+          await refetchConfigurations(configuration.data, queryClient);
```
jakewheeler: Similar comment here - I believe an invalidation alone will do what we're after.

client/src/pages/Configurations/ConfigActivate/index.tsx
```diff
@@ -1,55 +1,189 @@
...
+  // split out the last trailing comma
+  codeSetDisplay = `${codeSetDisplay.slice(0, codeSetDisplay.length)} code set(s)`;
+
+  const numCustomCodes = configurationData.custom_codes.length;
+  if (configurationData.custom_codes.length > 0) {
+    codeSetDisplay += `, ${numCustomCodes} custom code(s)`;
+  }
+
+  return codeSetDisplay;
+}
```
jakewheeler: 👍  Looks good. My only suggestion here is just some super minor clean up. Double check me that this still works as it should!

```suggestion
function buildRetainedDataDisplay(configurationData: GetConfigurationResponse) {
  const codeSetNames = configurationData.code_sets
    .map((c) => c.display_name)
    .join(', ');
  const customCodeCount = configurationData.custom_codes.length;

  let result = `${codeSetNames} code set(s)`;

  if (customCodeCount > 0) {
    result += `, ${customCodeCount} custom code(s)`;
  }

  return result;
}
```

client/src/pages/Configurations/ConfigActivate/index.tsx
```diff
@@ -1,55 +1,189 @@
...
+
+async function refetchConfigurations(
```
jakewheeler: We'll need to check that the invalidation works as it should but we may be able to get rid of this function entirely.

client/src/pages/Configurations/ConfigurationTitleBar.tsx
```diff
@@ -84,8 +95,25 @@ export function ConfigurationTitleBar({ step }: ConfigurationTitleBarProps) {
             </div>
           </div>
         </div>
-        <p> {CONFIGURATION_TITLE_CONTENTS[step].subtitle}</p>
+        <p>
+          {boldWord(
+            CONFIGURATION_TITLE_CONTENTS[step].subtitle,
+            CONFIGURATION_TITLE_CONTENTS[step]?.boldedWords ?? undefined
+          )}
```
jakewheeler: I think we can make this a bit simpler if we change `subtitle` to type `React.ReactNode` because then we could do something like this:
```ts
    activate: {
      title: 'Turn on configuration',
      subtitle: (
        <span>
          Refiner will <span className="font-bold">immediately</span> start to
          refine eCRs with {condition} as a reportable condition, in accordance
          with this configuration.
        </span>
      ),
```
And this line would become:
```suggestion
{CONFIGURATION_TITLE_CONTENTS[step].subtitle}
```

No logic needed 👍

client/src/pages/Configurations/ConfigurationTitleBar.tsx
```diff
@@ -84,8 +95,25 @@ export function ConfigurationTitleBar({ step }: ConfigurationTitleBarProps) {
...
+        </p>
       </div>
     </div>
   );
 }
+
+function boldWord(text: string, word: string | undefined) {
```
jakewheeler: We'll be able to delete this if we go with my suggestion above!

client/src/pages/Configurations/ConfigurationTitleBar.tsx
```diff
@@ -7,27 +7,38 @@ type ConfigurationSteps = 'build' | 'test' | 'activate';
...
-  },
-  activate: {
-    title: 'Turn on configuration',
-    subtitle:
-      'Refiner will immediately start to refine eCRs with gonorrhea as a reportable condition, in accordance with this configuration.',
-  },
+type ConfigurationTitleContent = {
+  title: string;
+  subtitle: string;
+  boldedWords?: string;
```
jakewheeler: Mentioned in my other comment that this might be a good way to go!
```suggestion
  subtitle: React.ReactNode;
```

client/src/pages/Configurations/ConfigActivate/modals/SwitchActivationModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+      title={`Switch to Version ${curVersion}`}
+      body={
+        <div>
+          <p id="activation-confirmation-modal-text" className="my-6">
+            You're about to stop Version {activeVersion} and start Version{' '}
+            {curVersion}
+          </p>
+          <p>
+            The eCR pipeline will begin using Version {curVersion}{' '}
+            <b>immediately</b>
```
jakewheeler: Extremely minor but I'd probably lean towards this instead of using a `<b>` tag here since [that tag has a semantic meaning](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/b) that doesn't really apply in this case.
```suggestion
            <span className="font-bold">immediately</span>
```

client/src/pages/Configurations/ConfigActivate/ActivationButtons.tsx
```diff
@@ -0,0 +1,65 @@
...
+      )}
+      {buttonState === StatusUpdateStates.TURN_ON && (
+        <TurnOnButtons handleActivation={handleActivation}></TurnOnButtons>
+      )}
+
+      {buttonState === StatusUpdateStates.TURN_OFF && (
+        <TurnOffButtons handleDeactivation={handleDeactivation} />
+      )}
+    </div>
+  );
```
jakewheeler: Would something like this work instead? Whenever the state gets updated this component should be re-rendered so there shouldn't be a need for us to use the `useEffect`. Let me know if I'm missing a detail!

And as usual please double check my logic 😆 
```suggestion
  if (
    configurationData.status === 'active' &&
    configurationData.active_version === configurationData.version
  ) {
    return <TurnOffButtons handleDeactivation={handleDeactivation} />;
  }

  if (!configurationData.active_version) {
    return <TurnOnButtons handleActivation={handleActivation} />;
  }

  return (
    <SwitchFromPrevious
      handleActivation={handleActivation}
      handleDeactivation={handleDeactivation}
      curVersion={configurationData.version}
      activeVersion={configurationData.active_version}
    />
  );
```

client/e2e/utils.ts
```diff
@@ -22,24 +39,26 @@ export async function logout(page: Page) {
...
-        throw new Error(String(err.stderr));
-      } else if (err.message) {
-        throw new Error(String(err.message));
-      } else {
-        throw new Error(JSON.stringify(err));
-      }
-    }
-    throw new Error(String(error));
-  }
+export async function createNewConfiguration(
```
jakewheeler: Nice 👍

client/src/pages/Configurations/ConfigActivate/ActivationButtons.tsx
fzhao99: this is much better thanks! had an earlier stateful version of the implementation and didn't think to clean up the useEffect once I simplified things

client/playwright.config.ts
fzhao99: Yeah I thought it'd be fine to use multiple workers because of this fact as well, but in practice, think something else is going wrong. It does seem that just using the one worker is making the tests pass every time though, so I opted to just set the one worker up for now until I have the chance to look further into it in a followup.

Issue #423, opened 2025-10-02
[USER STORY] Restrict concurrent configuration editing


kplaneaux:
## 👤💭 User Story

As a user, I want to be sure that I'm the only one making changes to a configuration, so that I can ensure changes aren't being made concurrently by other users without my knowledge

## 🏗️ Acceptance Criteria

- [ ] If a user is already editing a configuration, all other users are restricted from editing until that user has navigated away from the draft "build" page
- [ ] Other users have a notification banner stating "View Only: [user] is currently editing this configuration.
- [ ] Lock is triggered when any user opens the configuration for a condition and unlocked when either they navigate out of the condition detail page or their session ends

## 🎨 Designs

When the user is on a page/tab in Build or Activate, disable the CTAs.

https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=3337-38211&t=utx951MW7Zm5DmKS-1

PR #646, opened 2025-11-25
feat: Restrict concurrent editing of configurations

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Implements the feature described in `docs/decisions/0001_2025-10-31_restrict-concurrent-configuration-editing.md` which add the restriction around concurrent configuration editing.

## 🔗 Related Issue

- #423

## ✅ Acceptance Criteria

- [ ] If a user is already editing a configuration, all other users are restricted from editing until that user has navigated away from the draft "build" page
- [ ] Other users have a notification banner stating "View Only: [user] is currently editing this configuration.
- [ ] Lock is triggered when any user opens the configuration for a condition and unlocked when either they navigate out of the condition detail page or their session ends

## 🧪 How to test

> [!NOTE]
> Because this PR includes some changes to the database and to our local
> Keycloak service, you'll need to ensure that you've deleted all containers to
> avoid caching issues with the `refiner-realm.json` file.

1. Stop all containers and delete them.
1. Start application using `just dev up`.
1. Run `just database refresh` to clean, migrate, and seed the database.
1. Create a private incognito window or a different browser.
1. Log into the app with `refiner/refiner` in one browser.
1. Log into the app with `refiner2/refiner2` in the other browser.
1. Navigate to the same configuration after creating one in one of the browsers.
1. Note the `View only` banner.
1. Navigate away from the configuration in the browser that doesn't have the
   banner.
1. Note the `View only` banner is no longer present after a refresh.

### DIFF
--- a/refiner/tests/integration/conftest.py
+++ b/refiner/tests/integration/conftest.py
@@ -6,7 +6,6 @@ import pytest
 import pytest_asyncio
 from httpx import AsyncClient
 from lxml import etree
-from psycopg.rows import dict_row
 from saxonche import PySaxonProcessor
 from testcontainers.compose import DockerCompose
 
@@ -51,7 +50,7 @@ def test_username():
 @pytest_asyncio.fixture(scope="session")
 async def db_conn():
     conn = await psycopg.AsyncConnection.connect(
-        "postgresql://postgres:refiner@localhost:5432/refiner", row_factory=dict_row
+        "postgresql://postgres:refiner@localhost:5432/refiner"
     )
     try:
         yield conn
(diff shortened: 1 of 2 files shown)


client/e2e/homepage.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { test, expect } from './fixtures/fixtures';
+import { login, logout } from './utils';
```
fzhao99: We re-worked this test [in a recent PR](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/client/e2e/04%20-%20homepage.spec.ts), so think we can just get rid of this file

client/src/pages/Configurations/ConfigActivate/buttonDisplays/TurnOffButtons.tsx
```diff
@@ -16,6 +20,7 @@ export function TurnOffButtons({ handleDeactivation }: TurnOffButtonsProps) {
         modalRef={deactivateRef}
         opener
         className={classNames('self-start', SECONDARY_BUTTON_STYLES)}
```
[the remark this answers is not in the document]
fzhao99: Ditto for the similar components in test

Image

refiner/app/core/config.py
```diff
@@ -48,4 +48,5 @@ def _get_env_variable(name: str) -> str:
     "AWS_SECRET_ACCESS_KEY": _get_env_variable("AWS_SECRET_ACCESS_KEY"),
     "AWS_REGION": _get_env_variable("AWS_REGION"),
     "S3_UPLOADED_FILES_BUCKET_NAME": _get_env_variable("S3_UPLOADED_FILES_BUCKET_NAME"),
+    "CONFIG_LOCK_TIMEOUT_MINUTES": os.getenv("CONFIG_LOCK_TIMEOUT_MINUTES", "30"),
```
fzhao99: is this value being used anywhere? it seems we're setting a constant value in `configuration_locks.py` and that's the value being used for the session timeout

refiner/app/services/configuration_locks.py
```diff
@@ -0,0 +1,154 @@
...
+    async def renew_lock(
+        configuration_id: UUID,
+        user_id: UUID,
+        jurisdiction_id: str,
+        db: AsyncDatabaseConnection = db,
+    ) -> bool:
+        """
+        Renew the lock for a configuration, extending its expiration.
+        """
+        now = datetime.utcnow()
```
fzhao99: think we need to replace these with `datetime.now(timezone.utc)`

https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

refiner/app/core/config.py
rogeruiz: ooh good catch, I missed this one. I was using an environment variable at first. I'll remove it.

PR #677, opened 2025-12-09
feat: add details to configuration version selection menu

### LINKED ISSUE #600: [USER STORY] Version Menu Details
## 👤💭 User Story

As a user, I want to quickly see the details of past versions, so that I can select the one I want to view

## 🏗️ Acceptance Criteria

- [ ] Active/Inactive/Draft status added to version drop-down
- [ ] Drafted date/time or last active date/time added to version drop-down

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds more detail to the configuration version selection menu. Depending on the status of the configuration you will see something like this:
1. Draft configs -> `Draft created {date}, {time} by {user's name that created it}`
2. Active/inactive configs -> `Last activated {date}, {time} by {user's name that activated it}`

### Technical changes
- New migration script to add the required `created_by` field to the `configurations` table
- New hook to handle datetime formatting
- Couple of new components to use for the version selection menu
- Updated the activation and deactivation handlers to no longer require the client to provide a body
  - The server has enough info to determine what `condition_canonical_url` is required 
- Updated `_activate_configuration_db` and `_deactivate_configuration_db` to return `DbConfiguration` objects rather than `GetConfigurationResponseVersion` (I needed to update this type and it was causing conflicts with the primary use case for this type)
- Updated the activation and deactivation handlers  to throw exceptions when their previous status is invalid to transition to the newly desired status
- Updated configuration activation to set the `last_activated_by` user ID value (I needed this info for this feature)
- Tests were updated to account for the new configuration property (`created_by`) and changes to the API

## 📺  Demo
Kapture 2025-12-10 at 12 28 57

## 🔗 Related Issue

Fixes #600 

## 🧪 How to test

1. Create a draft config for any condition
2. Activate the draft (`PATCH http://localhost:8081/api/v1/configurations/<config ID>/activate`)
3. Create a new draft config for that same condition using the button on the config page
4. Take a look at the version menu to ensure it's displaying the expected information


client/src/pages/Configurations/ConfigBuild/VersionMenu.tsx
```diff
@@ -62,3 +75,53 @@ export function VersionMenu({
...
+function MenuItemDetail({
+  created_at,
+  created_by,
+  last_activated_at,
+  last_activated_by,
+  className,
+}: MenuItemDetailProps) {
+  const formatDatetime = useDatetimeFormatter();
+
+  if (last_activated_at && last_activated_by) {
```
jakewheeler: There should never be a case where one of these is `null` and the other is not.

client/src/pages/Configurations/ConfigBuild/VersionMenu.tsx
```diff
@@ -34,20 +35,32 @@ export function VersionMenu({
...
+                      isActive={config.status === 'active'}
+                    />
+                    <MenuItemDetail
+                      className="text-gray-60"
+                      created_at={config.created_at}
+                      created_by={config.created_by}
+                      last_activated_at={config.last_activated_at}
+                      last_activated_by={config.last_activated_by}
+                    />
+                  </div>
```
jakewheeler: This is the main change to look at 👍

refiner/app/api/v1/configurations.py
```diff
@@ -1504,17 +1491,24 @@ async def activate_configuration(
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Configuration is already active.",
         )
 
     active_config = await activate_configuration_db(
-        config_to_activate=config_to_activate,
-        canonical_url=body.condition_canonical_url,
+        configuration_id=config_to_activate.id,
+        activated_by_user_id=user.id,
+        canonical_url=config_to_activate.condition_canonical_url,
```
jakewheeler: We can grab the canonical URL from the config we're wanting to activate so no need to make the client provide it.

refiner/app/db/configurations/activations/db.py
```diff
@@ -1,67 +1,110 @@
...
+
 
 async def _activate_configuration_db(
     configuration_id: UUID,
+    activated_by_user_id: UUID,
     *,
-    cur: AsyncCursor[GetConfigurationResponseVersion],
-) -> GetConfigurationResponseVersion | None:
+    cur: AsyncCursor[CursorType],
+) -> DbConfiguration | None:
```
jakewheeler: I needed to modify `GetConfigurationResponseVersion` for my own purposes in this PR which caused some conflicts with what we were using it for here. I took this opportunity to swap these functions over to returning `DbConfiguration | None` which I think is a better fit and a pattern we tend to follow for many of the other config related DB functions.

fzhao99 (approved):

PR #694, opened 2025-12-15
fix structurizr warnings

fzhao99:
# 🔀 PULL REQUEST

adding the things in structurizr needed to render the container-level diagrams

Running `just structurizr lite` should now render out the component-level changes without errors.

https://github.com/user-attachments/assets/dc5509b6-075a-4773-be62-57a7c1f480a9

## Additional Notes

Going to need to rework the publishing workflow due to some runner token permissions, but wanted to do that separate to this since we're needing to generate the items ahead of a deadline


.github/workflows/save-rendered-diagrams.yaml
```diff
@@ -6,24 +6,35 @@ permissions:
...
       # |_______||_____||____|_____|   __|
       #                            |__|
-      #
+
+      # generate a new app token so that the followup build / test workflow
+      # steps will still run
+      # https://docs.github.com/en/enterprise-server@2.22/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow:~:text=When%20you%20use,push%20events%20occur.
+
+      # todo: add these back in when we have the creds set up in the repo
+      # - name: Generate token
```
fzhao99: think since we're committing files in this workflow, the workflow will need elevanted app token permissions since [GitHub disables workflow runs for GitHub token-scoped commits](https://docs.github.com/en/enterprise-server@2.22/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow:~:text=When%20you%20use,push%20events%20occur.). Without this piece, whenever this workflow completes, the follow-up build and test workflows don't run and the PR gets stuck in limbo without the required workflow steps completing.

PR #714, opened 2025-12-22
chore: update e2e tests with configuration fixtures

### LINKED ISSUE #752: [TASK] Make e2e's fixtures-first and repeatable
Our e2e's right now don't clean up after themselves and thus aren't rerunnable locally. This causes issues with local development and flakiness in CI.

This ticket captures the work to make them so: tests should be rerunnable and more ergonomic to work with

### Acceptance Criteria
- [x] Tests are re-runnable locally and clean up after themselves
- [x] Tests use the Playwright fixture feature to make tests more extendable

fzhao99:
# 🔀 PULL REQUEST
Updating the e2e's to be more extendable by adding a fixture for configurations for use in subsequent tests

## 💡 Summary
- Adds a `configurationsPage` fixture for tests that use a configuration. This will update based off worker values to allow for re-runnability in local dev
- Add a Just hook to cascade delete a given configuration so that e2e's can clean themselves up. Invokes this in the fixture cleanup so that tests can run indepently
    - Updates some tests to be independent in light of this 
- Offers a getter to grab the configuration name from the created fixture
- Refactors the tests to use the created fixture name 

## Additional Info
Because test 3 is hard-coded to create the COVID and Influenza configurations, those aren't refactored using this fixture. Those are set up to use the corresponding utility functions instead

## 🔗 Related Issue
- Fixes #752 
  - #752

## 🧪 How to test

End to end tests
1. Spin up the e2e's in local UI mode
2. Run them multiple times and observe the cleanup output in local dev

### DIFF
--- a/refiner/migrations/000007_add_s3_url_to_config.down.sql
+++ b/refiner/migrations/000007_add_s3_url_to_config.down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE configurations
+DROP COLUMN IF EXISTS s3_url;
--- a/refiner/migrations/000007_add_s3_url_to_config.up.sql
+++ b/refiner/migrations/000007_add_s3_url_to_config.up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE configurations
+ADD COLUMN s3_url TEXT;


.justscripts/just/database.just
```diff
@@ -101,3 +101,7 @@ check-code CONDITION VERSION SYSTEM CODE:
             AND c.version = '{{VERSION}}' \
             AND elem->>'code' = '{{CODE}}'; \
     "
+[group('db')]
+[doc('Cascade delete all the related configuration artifacts. For use after an E2E test run')]
+delete-configuration CONFIGURATION_NAME:
```
fzhao99: This was the best way I could think of to set up the e2e cleanup, but obviously want to make sure we have sufficient protections to have this not accidentally run against a production data source. Lmk if folks can think of anything that needs adding to check the data we're attempting to delete is fine to delete

client/e2e/02 - add-condition.spec.ts
```diff
@@ -18,32 +24,28 @@ test.describe('Adding/modifying configurations by initial condition', () => {
...
 
     // Add a few code sets and check that remove button shows up properly
-    await page
-      .getByRole('listitem')
-      .filter({ hasText: 'Acute Flaccid Myelitis (AFM)' })
-      .click();
+    await page.getByRole('listitem').filter({ hasText: 'Diphtheria' }).click();
     await expect(
-      page.getByRole('button', { name: 'Remove Acute Flaccid Myelitis (AFM)' })
+      page.getByRole('button', { name: 'Remove Diphtheria' })
```
fzhao99: changing these to codesets further down the alphabetized list since sometimes in local dev, re-running the tests can result in name conflicts for the auto-selected condition

client/e2e/fixtures/setup.ts
```diff
@@ -32,10 +76,29 @@ export const test = baseTest.extend<object, { workerStorageState: string }>({
...
 });
+
+type RefinerFixtures = {
+  configurationPage: ConfigurationPage;
+};
+export const test = defaultFixturesTest.extend<RefinerFixtures>({
+  configurationPage: async ({ page }, use, workerInfo: WorkerInfo) => {
+    const configurationPage = new ConfigurationPage(
+      page,
+      workerInfo.workerIndex
```
jakewheeler: Why does `workerIndex` get passed in as the `conditionIndex`?

.justscripts/just/database.just
jakewheeler: This seems fine to me. It's still murky to me at this point how exactly we'll connect to and manage a prod DB but we'll certainly do whatever needs to happen to make sure only expected actions can be run against the database. I wouldn't worry about it too much for now but it is definitely a good thing to think about.

jakewheeler (approved):
Thanks Bob 🎉

PR #729, opened 2026-01-07
docs(rfc): S3 file structure for activations

### LINKED ISSUE #696: [SPIKE] S3 Bucket File Structure
separating jurisdiction configurations from one another, what does activation look like and how are the pieces moving, same with de-activate

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a document describing a solution to how we will be able to handle managing serialized configurations in an S3 bucket.

Once we iterate and agree upon the contents of this document, I'll work to update or create tickets that match the outlined plan.

## 🔗 Related Issue

Fixes #696

### DIFF
--- a/docs/decisions/0003_2026-01-06_activation-s3-file-structure.md
+++ b/docs/decisions/0003_2026-01-06_activation-s3-file-structure.md
@@ -368,7 +368,7 @@ The initial format of this file will look as such:
 
 Epic: [[EPIC] Production Activation (Lambda)](https://github.com/CDCgov/dibbs-ecr-refiner/issues/401)
 
-Tickets will be created or updated as needed to be relevant to the plan described in this document.
+- [#767](https://github.com/CDCgov/dibbs-ecr-refiner/issues/767)
 
 ### Real life scenario example
 


docs/decisions/0003_2026-01-06_activation-s3-file-structure.md
```diff
@@ -0,0 +1,789 @@
...
+
+```sh
+.
+└── {jurisdiction ID}/
+    └── {condition code}/
+        ├── current.json
+        └── {version}/
+            ├── active.json
+            └── metadata.json
+```
```
fzhao99: Not necessarily suggesting this world is better than what you have here (bc I haven't thought through it too deeply), but is there a world where we just store the active configuration per condition code and not store subsequent / previous versions?

Thinking is that we store in the S3 bucket only the information needed to process the lambda (ie the current active config) and have the db be the source of truth for the version history. On activation, the web app then deletes the active version and swaps it with the new information. Here's the list of pros/cons I can think of compared to what we have here:

Pros:
- Store less data / only the data that is relevant for the lamda
- Lambda data retrieval is faster, since it only needs to check for existence of one file in one place (and assumes the absence of that information means there's no active config)
- We store less version data across fewer places, which might guard against version drift issues
- Conceptually the storage model is a bit simpler 

Cons:
- Guessing the main reason you opted to add in the second `current.json` file is this bit (ie to protect against race condition issues where the lambda is trying to grab a file in the process of being deleted):

```
The main goal in using this file is to handle failures gracefully and without causing interruption to Lambda. We are able to enable/disable the Lambda's processing and/or point it to use a different configuration version without potentially deleting files it may be trying to use.
```
- Activation process is more heavily read/write and potentially less failsafe
- There's more communication between the web app and S3 more often
- Other things I'm missing?

In the event that you thought through this option and opted to go the two-file route, would love to understand it further!

PR #730, opened 2026-01-08
fix: use correct python timestamp comparisons

fzhao99:
The most recent PR had bad timestamp comparison functions that broke some app functionality. This uses [the recommend fix](https://stackoverflow.com/questions/60003764/typeerror-cant-compare-offset-naive-and-offset-aware-datetimes) to do correct datetime comparisons


jakewheeler (approved):
Thank you!

rogeruiz (approved):
lgtm 🌈

PR #733, opened 2026-01-08
fix: useLogin should only be called within App

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR addresses an issue where the `useLogin` hook was being called from various components. This is undesirable because every time `useLogin` is invoked it sends a network request to refetch user info.

This PR removes all usages of `useLogin` outside of the top-level `App` component. The client-side functionality that was depending on the `user` coming from `useLogin` has been moved server-side.

Additionally, this seems to have fixed an issue where the locking banner was always quickly being flashed to the user on page load because the locking state was being calculated by the client.

## 🧪 How to test

All automated tests should pass and the functionality should remain unchanged.


rogeruiz (approved):
lgtm 🌈

PR #751, opened 2026-01-13
Update Refiner to use 4.0.0 Condition Groupers

### LINKED ISSUE #705: [TASK] Update TES to 4.0.0 CGs
## 🤔 Why it needs to be done

APHL has just released the 4.0.0 versions of the condition groupers. We need to update them in the repo and make some changes in a couple of places. If easily done; add the additional context groupers too.

## 🏗️ Acceptance Criteria

- [ ] ...

## ⚙️ Technical Notes

- ...

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Adding the 4.0.0 Condition Grouper files as well as the 4.0.0 Additional Context Grouper files. The 4.0.0 release did not include a new Reporting Specification Grouper file--the 4.0.0 Condition Groupers reference the Reporting Specification Grouper file that came with the 3.0.0 release.

For the time being we're not grabbing the codes from the 4.0.0 Condition Grouper's `expansion.contains` element in the FHIR bundle. We're using the existing methodology, which is still working. This change will be made in the next iteration when we add the additional context grouper codes to our `conditions` table. That will be a good point where we can evaluate the delta between how we're sourcing codes currently and what the difference is when pulling them directly from `expansion.contains`.

## 🔗 Related Issue

Fixes #705 
- #705 

## 🧪 How to test

Unit tests

5. You can also run the maintenance scripts in the `scripts/` directory

or

Testing the application:

1. Run `docker compose up`
4. You should be able to see that everything is working as expected.

### DIFF
--- a/client/src/components/Button/ModalToggleButton.tsx
+++ b/client/src/components/Button/ModalToggleButton.tsx
@@ -1,0 +1,63 @@
+import {
+  ModalToggleButtonProps as UswdsModalButtonProps,
+  ModalToggleButton as UswdsModalToggleButton,
+} from '@trussworks/react-uswds';
+import {
+  ButtonVariant,
+  DISABLED_BUTTON_STYLES,
+  PRIMARY_BUTTON_STYLES,
+  SECONDARY_BUTTON_STYLES,
+  SELECTED_BUTTON_STYLES,
+  TERTIARY_BUTTON_STYLES,
+} from '.';
+import classNames from 'classnames';
+import { JSX } from 'react';
+
+interface ModalButtonProps extends UswdsModalButtonProps {
+  variant?: ButtonVariant;
+  className?: string;
+}
+export function ModalToggleButton({
+  children,
+  modalRef,
+  opener,
+  closer,
+  variant,
+  className,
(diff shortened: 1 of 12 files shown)


refiner/app/db/conditions/db.py
```diff
@@ -8,8 +8,8 @@
 from ..pool import AsyncDatabaseConnection
 from .model import DbCondition, DbConditionBase
 
-# TES and refiner are currently using version 3.0.0 for CGs and its child RSGs
-CURRENT_VERSION = "3.0.0"
+# TES and refiner are currently using version 4.0.0 for CGs and its child RSGs
+CURRENT_VERSION = "4.0.0"
```
jakewheeler: This is not really related to your work - I just wanted to call out that we really should figure out a way to get rid of this soon 😬

robertmitchellv: we also have to change this in the integration test's conftest.py file too

jakewheeler: I added a note to the eng sync doc about this!

PR #799, opened 2026-01-27
test: Lambda dev setup and integration tests

### LINKED ISSUE #770: [TASK] Add Lambda integration tests
## 🚧 What needs to be done

With the completion of #767, we should now have a functional Lambda container image that can read configuration files from S3 and use them to produce refined outputs.

We need a way to be able to, to the best of our ability, test that this functionality works prior to shipping the image to a real AWS environment. One way that we can do this is by making the Lambda part of Refiner's `docker-compose.yml` setup and writing integration tests to run against that container locally and in the GitHub Actions pipeline.

These tests will need to:
- Connect to the container
- Make the web app write an activation file to localstack
- Trigger the Lambda via a simulated event
- Make use of a seeded eICR and RR
- Have Lambda find and use the activation file

## 🏗️ Acceptance Criteria

- [ ] Local Lambda container testing setup has been configured
- [ ] Basic tests have been added to the integration testing suite
  - [ ] Finding and using an activation file
  - [ ] Using a `current.json` file with a `null` version property
  - [ ] Missing `current.json` file

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds integration tests that run against the newly created Lambda container (defined by `Dockerfile.lambda`). The Lambda container has been configured as a part of the project's Docker Compose and will be available to run requests against when the project containers boot up.

- Lambda container is available on port `9000` when project is running
  - URL: `http://localhost:9000/2015-03-31/functions/function/invocations`
- Adds integration tests (`tests/integration/test_lambda.py`) that run requests against the Lambda container
  - Make uses of a pre-defined COVID-19 activation file stored in `tests/fixtures/lambda`
- Updated `lambda/README.md` documentation
- Added `S3_ENDPOINT_URL` env variable to `lambda_function.py`, required for testing purposes
- Updated configuration files to live under the parent `configurations/` directory instead of JD IDs being at the top level of the S3 bucket (`services/aws/s3_keys.py`)
- Created functions to centralize the retrieval of S3 file keys to avoid typos and repeated strings all over the code base
  - Updated various files to use these functions

## 🔗 Related Issue

Fixes #770 

## 🧪 How to test

1. Shut down your docker compose environment if it's running and then start it up
2. You should see the `lambda` service build and begin running
3. Run the integration tests, `pytest refiner/tests -m integration`
4. You should see all of the new `tests/integration/test_lambda.py::TestLambda` tests passing


lina-roth (approved):
Tests work for me! All looks good. Nice documentation!

PR #812, opened 2026-02-02
chore: optimize seeding script

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR decreases the time it takes to run the seeding script by switching from `ValueSet` models to `dict`. The two main optimizations are:

1. No longer using `ValueSet.model_validate()`
2. Only checking for required data

### Comparison with `main`

Running `just db refresh` on `main` takes my machine **~18 seconds** to run, not counting dependency installations.

Running `just db refresh` on this branch takes my machine **~5 seconds** to run, not counting dependency installations.

Note that not needing to install `fhir.resources` also significantly reduces the container startup time.

## 🧪 How to test

- Integration tests pass
- `scripts/maintenance/check_seed_db.py` output should be identical to `main`'s

## ℹ️ Additional Information

> [!IMPORTANT]
> My main question: do we absolutely need this intensive validation or is this PR's solution acceptable? If we do need it, I'd advocate for making the validation occur during the TES data downloading process so that validation doesn't occur in the data import pipeline.


robertmitchellv (approved):
i suppose i'm ok with this but i do think we should use the `fhir` library in the future--especially when updating data in prod. it's probably slower because it's using pydantic under the hood for validation so it's making sure that the ValueSet resource is correct before using it. i think that would be valuable for the future

PR #821, opened 2026-02-04
chore: optional image push destinations when building manually

### LINKED ISSUE #822: [TASK] Add optional GHCR/ECR push to Docker build job
## 🚧 What needs to be done

When running a build/push, we do not currently have a way to select where the built image ends up. Right now it will always attempt to push to both GHCR and ECR. This is not ideal because there are situations, such as testing, where we don't need the image to be stored in both repositories. 

We also are unable to run this job against the same branch twice when not providing a version tag because ECR is setup to require immutable tags, meaning if the same branch name tries to be pushed to ECR again, the job will fail.

In this ticket, we want to:
- [x] Make each repository (GHCR and ECR) optional when running the workflow manually
  - [x] Set the options to `true` as the default
- [x] Push to both repositories when merging to `main`

## ⚙️ Technical Notes

N/A

## 📜 Additional context

N/A

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the build/push job so that we can choose not to push to GHCR and/or ECR when building and storing images. By default it will push to both image registries, however, there are now checkboxes when running the job that can be toggled. 

If both options are toggled off the job will fail since no tags are present.

image

Merging into `main` remains unchanged - `main` and `latest` images will be pushed to both GHCR and ECR automatically.

## 🔗 Related Issue

Fixes #822 

## 🧪 How to test

Take a look at a few of the [most recent jobs I kicked off manually](https://github.com/CDCgov/dibbs-ecr-refiner/actions/workflows/docker-image-push.yml). 

The output of the `Debug Docker tags` step is the easiest place to check that the tags look correct.

### DIFF
--- a/ops/entrypoint.sh
+++ b/ops/entrypoint.sh
@@ -38,7 +38,7 @@ case "$COMMAND" in
         ;;
     import)
         echo "Importing condition data"
-        exec python3 ./scripts/seeding/seed_db.py
+        exec python3 ./scripts/seeding/load_tes_data.py
         ;;
     python|python3)
         echo "Running Python script: $*"
@@ -47,7 +47,7 @@ case "$COMMAND" in
     prepare-db)
         echo "Running migration scripts and updating condition data"
         migrate -path ./migrations -database "$DATABASE_URL" up
-        exec python3 ./scripts/seeding/seed_db.py
+        exec python3 ./scripts/seeding/load_tes_data.py
         ;;
     *)
         echo "Running custom command: $COMMAND $*"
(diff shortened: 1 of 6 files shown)


.github/workflows/docker-image-push.yml
```diff
@@ -71,26 +82,41 @@ jobs:
...
+            if [ "$REF_NAME" = "main" ]; then
+              [ "$PUSH_GHCR" = "true" ] && TAGS+=("ghcr.io/${{ steps.repo.outputs.owner }}/dibbs-ecr-refiner/${BASES[$IMAGE]}:latest")
+              [ "$PUSH_ECR" = "true" ] && TAGS+=("$ECR_REGISTRY/dibbs-dev-refiner/${BASES[$IMAGE]}:latest")
+            fi
+
+            # Output as comma-separated string
+            echo "${IMAGE,,}_tags=$(IFS=, ; echo "${TAGS[*]}")" >> $GITHUB_OUTPUT
+          done
+
+      - name: Debug Docker tags
```
jakewheeler: Going to leave this in since it's a quick step and could be useful for tag debugging.

PR #824, opened 2026-02-04
refactor: remove CURRENT_VERSION constant from DB queries

### LINKED ISSUE #771: [TASK] Remove usage of `CURRENT_VERSION` constant
## 🚧 What needs to be done

We need to modify `conditions/db.py` to no longer rely on [`CURRENT_VERSION`](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/db/conditions/db.py#L12).

Condition and configuration entries in the database should be able to keep track of their own TES version for themselves. Relying on this variable is unsafe because it could incorrectly associate data between different TES versions.

## 🏗️ Acceptance Criteria

- [ ] Configurations always know to use conditions for the corresponding TES version
- [ ] `CURRENT_VERSION` is no longer used, everything is now dynamic

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This pull request refactors the database query layer to remove the hardcoded `CURRENT_VERSION` constant. Previously, many database functions were locked to a specific version of condition value sets (e.g., "4.0.0"), which prevented the system from handling multiple versions of conditions dynamically.

The primary goals of this change are:
1.  **Decouple from a single hardcoded version:** Allow the system to be aware of and process multiple versions of conditions (e.g., COVID-19 v1.0.0, v2.0.0, etc) that exist simultaneously in the database.
2.  **Enable version-aware logic:** Update services that consume condition data, particularly the "Independent Testing" feature, to correctly match reportable conditions found in an RR with the *specific version* of that condition a jurisdiction has configured. As we move to update new data in prod, we will want to be careful about how we allow users to pull in new "included conditions" depending on the version of the primary condition linked to the configuration.

**Key Changes:**
*   Removed `WHERE version = %s` clauses that used `CURRENT_VERSION` from database queries in `refiner/app/db/conditions/db.py`.
*   Introduced a new function `get_latest_conditions_db` to intelligently find the most recent semantic version for each condition, used for populating UI dropdowns.
*   Significantly refactored the `independent_testing` service in `refiner/app/services/testing.py` to:
    *   Fetch all possible condition versions matching a reportable code.
    *   Reconcile these possibilities against the jurisdiction's actual configurations.
    *   Correctly identify a condition as "configured" if *any* of its found versions match a configuration, resolving a bug where conditions were reported as both configured and not configured.
  
## 🔗 Related Issue

Fixes #771 
- #771 

## ✅ Acceptance Criteria

- [x] Configurations always know to use conditions for the corresponding TES version
- [x] `CURRENT_VERSION` is no longer used, everything is now dynamic

## 🧪 How to test

Unit tests

or

Testing the application:

1. Run `docker compose up`
4. You should be able to run independent testing with no configs, then one (COVID-19), then with both COVID-19 and Influenza to see that the message before refining reflects what we expect to see.

## ℹ️ Additional Information

One of the main challenges during this refactor was updating the `independent_testing` service. The initial logic was not prepared to handle a one-to-many relationship between a reportable SNOMED code and multiple condition versions in the database.

The solution involved a multi-step process:
1.  Gather all potential condition versions from the database.
2.  Fetch all active configurations for the jurisdiction to create a "ground truth" lookup set.
3.  Group the potential conditions by their conceptual `canonical_url` to treat all versions of "Influenza" as one entity.
4.  Iterate through these groups to make a single, clear decision on whether the condition was configured or not.

This approach fixed the original bug and also resolved an edge-case `IndexError` where the function would fail if no reportable conditions were found in an uploaded file.

### DIFF
--- a/refiner/app/db/conditions/db.py
+++ b/refiner/app/db/conditions/db.py
@@ -80,10 +80,10 @@ async def get_latest_conditions_db(
     # STEP 3: HIGHLIGHT LATEST VERSION FOR EACH GROUP
     # iterate through the dictionary and use `max()` with `packaging.version.parse`
     # as the key to correctly identify the latest version for each condition
-    latest_conditions: list[DbConditionBase] = []
-    for url, cond_group in grouped_conditions.items():
-        latest_in_group = max(cond_group, key=lambda c: parse(c.version))
-        latest_conditions.append(latest_in_group)
+    latest_conditions: list[DbConditionBase] = [
+        max(cond_group, key=lambda c: parse(c.version))
+        for cond_group in grouped_conditions.values()
+    ]
 
     # STEP 4: SORT/PACKAGE FINAL LIST
     # sort the final list alphabetically by `display_name` for a


refiner/app/db/conditions/db.py
```diff
@@ -34,19 +28,72 @@ async def get_conditions_db(db: AsyncDatabaseConnection) -> list[DbConditionBase
...
+            grouped_conditions[cond.canonical_url] = []
+        grouped_conditions[cond.canonical_url].append(cond)
+
+    # STEP 3: HIGHLIGHT LATEST VERSION FOR EACH GROUP
+    # iterate through the dictionary and use `max()` with `packaging.version.parse`
+    # as the key to correctly identify the latest version for each condition
+    latest_conditions: list[DbConditionBase] = []
+    for url, cond_group in grouped_conditions.items():
+        latest_in_group = max(cond_group, key=lambda c: parse(c.version))
+        latest_conditions.append(latest_in_group)
```
jakewheeler: We could do this, which I think is pretty readable:
```suggestion
    latest_conditions = [
        max(cond_group, key=lambda c: parse(c.version))
        for cond_group in grouped_conditions.values()
    ]
```

If we wanted to keep it closer to what you have now, my only other suggestion would be to use `.values()` instead of `.items()` and delete `url` since it's unused.

refiner/app/db/conditions/db.py
```diff
@@ -66,10 +113,10 @@ async def get_condition_by_id_db(
                 icd10_codes,
                 rxnorm_codes
             FROM conditions
-            WHERE id = %s AND version = %s
+            WHERE id = %s
             """
 
-    params = (id, CURRENT_VERSION)
+    params = (id,)
```
jakewheeler: 🙏

refiner/app/services/testing.py
```diff
@@ -123,70 +126,116 @@ async def independent_testing(
...
     # STEP 1:
-    # extract and filter reportable RC codes for jurisdiction
+    # extract all reportable condition snomed codes from the RR file that are reportable for the given jurisdiction
+    # then, for each code, get a list of all possible condition versions (because we can't know which version is configured a priori)
     reportability_data = _extract_reportable_conditions_for_jurisdiction(
         xml_files, jurisdiction_id
     )
     rc_codes_for_jurisdiction = reportability_data["rc_codes_for_jurisdiction"]
-    rc_to_condition = await _map_rc_codes_to_conditions(
+    rc_to_conditions_list = await _map_rc_codes_to_conditions(
```
jakewheeler: I think it'd make sense to either leave the name as it was or switch `_list` to `_dict` since that's what `_map_rc_codes_to_conditions()` returns.

client/src/api/schemas/validationErrorCtx.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Generated by orval v8.2.0 🍺
+ * Do not edit manually.
+ * DIBBs eCR Refiner
+ * Please visit the repo for more info: https://github.com/CDCgov/dibbs-ecr-refiner
+ * OpenAPI spec version: 1.0.0
+ */
+
+export type ValidationErrorCtx = { [key: string]: unknown };
```
jakewheeler: Where is this coming from? 🤔 I'm not seeing the reason this is getting generated.

jakewheeler (changes requested):
Looks good Robert 👍  Just added a couple of potential cleanup related suggestions.

jakewheeler (approved):
🏆 Well done, thanks!

PR #858, opened 2026-02-19
docs(rfc): `displayName` Enrichment

### LINKED ISSUE #802: [SPIKE] Adding Missing Display Names
## 🚧 What needs to be done

APHL and CDC have asked us to address an issue with eICRs from Epic and Meditech where certain coded entries have a code but do not have a display name. We would need to pull the display names from the TES code sets and custom codes provided by the user (this exists in the tables already).

## 🤔 Why it needs to be done

This is a major issue for some jurisdictions as it prevents them from making use of the eCR data in their downstream systems and processes as display names are the primary way data is represented there.

## ⏰ Timebox

1-2 days

## 📝 To-do list

  - [ ] ...
- [ ] Research + Analysis
  - [ ] ...
  - [ ] ...
  - [ ] ...
- [ ] Review and Finalize
  - [ ] ...

## 📜 Additional context

Some things to consider:
1. How this will affect any narrative/HTML portion refining
2. Performance implications 

[Example file](https://drive.google.com/file/d/1puKvsv-qSFx15FIb1TObrw0T9FMEq3B6/view?usp=drive_link) from Tim from APHL: "examples of entries in an eICR that have @code and @codesystem but not @displayName.  I have attached an example test 3.1 document from Epic (no PII) that has examples for most sections.  Each has a text element with a reference to the Narrative to retrieve the display."

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This RFC will walk through three ways we can do things to solve our `displayName` issue and one way, in my view, that makes the most sense to work on as it sets us up really nicely to be able to potentially tackle the refined narrative generation that is section specific rather than one-size-fits-all approach.

## 🔗 Related Issue

Fixes #802 
- #802

### DIFF
--- a/.github/scripts/security-summary.js
+++ b/.github/scripts/security-summary.js
@@ -1,0 +1,367 @@
+const fs = require("fs");
+
+/**
+ * Parse Trivy scan results and count vulnerabilities
+ */
+function parseScanResults(images) {
+  let totalCritical = 0;
+  let totalHigh = 0;
+  let totalMedium = 0;
+  let totalLow = 0;
+  const imageResults = [];
+
+  for (const image of images) {
+    try {
+      const results = JSON.parse(
+        fs.readFileSync(`trivy-${image}-results.json`, "utf8"),
+      );
+
+      let imageCritical = 0,
+        imageHigh = 0,
+        imageMedium = 0,
+        imageLow = 0;
+
+      if (results.Results) {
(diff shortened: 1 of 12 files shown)


jakewheeler (approved):
Good work outlining a few different options! The direction you chose is exactly the way I'd want to approach this as well.

PR #908, opened 2026-03-03
fix: only use active configs in independent testing

### LINKED ISSUE #309: [USER STORY] Use active configurations in independent Test Flow
## 👤💭 User Story

As a user, I want to see how Refiner will handle my test eCR, so that I can make sure it's functioning the way I want it to

## 🏗️ Acceptance Criteria

- [ ] Independent Testing flow only uses active configurations (mirroring what will happen with the lambda)
- [ ] Detects and indicates whether a condition has an active configuration or not
- [ ] Updated copy around needing an active configuration

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=2902-188329&t=zA8rVCNKp1danm9l-1

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Changes the independent testing flow to only parse active configurations, displaying warning messages accordingly.

### All inactive
Screenshot 2026-03-03 at 9 12 53 AM

### One inactive, one missing
Screenshot 2026-03-03 at 9 11 29 AM

### One active, one inactive
Screenshot 2026-03-03 at 9 12 12 AM

## 🔗 Related Issue

Fixes #309 
- #309 

## ✅ Acceptance Criteria
- [ ] Independent Testing flow only uses active configurations (mirroring what will happen with the lambda)
- [ ] Detects and indicates whether a condition has an active configuration or not
- [ ] Updated copy around needing an active configuration

## 🧪 How to test
1. Create a COVID configuration, but don't activate it
2. Try the independent testing flow with the test file. Notice error message for missing Influenza and inactive COVID.
3. Create and activate Influenza condition
4. Notice the COVID warning persists but the Influenza isn't erroring anymore
5. Activate COVID
6. Notice both conditions don't generate errors

### DIFF
--- a/refiner/requirements-dev.txt
+++ b/refiner/requirements-dev.txt
@@ -1,4 +1,4 @@
-fhir.resources
+fhir.resources==8.2.0
 httpx
 pre-commit
 pytest
--- a/refiner/requirements.txt
+++ b/refiner/requirements.txt
@@ -1,7 +1,7 @@
 boto3==1.42.59
 Authlib==1.6.9
 chardet
-fastapi==0.135.1
+fastapi==0.129.0
 fhir.resources
 httpx
 itsdangerous
@@ -14,5 +14,6 @@ pydantic>=2.0.0
 python-json-logger==4.0.0
 python-multipart
 requests
+types-requests
 rich
 uvicorn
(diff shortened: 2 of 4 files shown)


refiner/app/services/testing.py
```diff
@@ -163,6 +166,7 @@ async def independent_testing(
         return {
             "refined_documents": [],
             "no_matching_configuration_for_conditions": [],
+            "no_active_configuration_for_conditions": [],
```
jakewheeler: I am seeing that this setup (Influenza V2 is active):
Image

is producing this result:
Image

At first glance I think it might be this? We're just grabbing any config we find that is associated with the condition ID currently.
```python
        matching_config = next(
            (
                config
                for config in all_jurisdiction_configs
                if config.condition_id == representative_condition.id
            ),
            None,
        )
```

Issue #917, opened 2026-03-06
[TASK] Remediate DAST scan outputs


fzhao99:
## 🚧 What needs to be done
Fix DAST outputs from [this report](https://drive.google.com/drive/folders/1d0I2ehfbcRJWgK8Vwf_EpGNsLzdyhn1_
)
 
## 🏗️ Acceptance Criteria

- [ ] ...

## ⚙️ Technical Notes

- ...

PR #918, opened 2026-03-06
fix: harden session middleware configuration

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This is an attempt to fix the `SessionFixation` flag we're seeing in the DAST scan. 

I believe this is coming from a cookie being created by `SessionMiddleware`, which is only used to persist the OIDC transaction state during the auth flow. **The app does not actually make use of this cookie once logged in** because we issue our own `refiner-session` cookie.

I've done a few things in this PR:
- Configure a `nonce` value for the auth flow (not sure why this was missing, not directly related to the issue)
- Check that `nonce` has a value before attempting a login via IdP
- Clear the `SessionMiddleware` cookie as soon as the log in has completed (Part of the `SessionFixation` fix)
- Harden the configuration of `SessionMiddleware` to limit flagging (Part of the `SessionFixation` fix)

## 🔗 Related Issue

Attempts to fix the `Medium` level vuln in #917 

## 🧪 How to test

You should not see any changes. The auth flow should be working as it did previously.

## ℹ️ Additional Information

> [!IMPORTANT]
> This does change the auth flow. We will want to test this ASAP in dev once it's merged to make sure everything looks good.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -7169,9 +7169,9 @@
       "license": "ISC"
     },
     "node_modules/immutable": {
-      "version": "5.1.4",
-      "resolved": "https://registry.npmjs.org/immutable/-/immutable-5.1.4.tgz",
-      "integrity": "sha512-p6u1bG3YSnINT5RQmx/yRZBpenIl30kVxkTLDyHLIMk0gict704Q9n+thfDI7lTRm9vXdDYutVzXhzcThxTnXA==",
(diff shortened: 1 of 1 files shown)


fzhao99:
if you wouldn't mind pinging geo in teams once this gets in to let him know we have an image that should pass the DAST scan, would be swell

PR #946, opened 2026-03-13
fix: move lambda prefixes to env vars

fzhao99:
# 🔀 PULL REQUEST
Moving refiner prefixes to env variables per our discussion in slack

Also picked up the renaming of the shadow RR output URL per the discussion of teams here

## 💡 Summary
Lambda prefixes are now read solely from env variables and not reliant on hard-coded defaults

## 🧪 How to test
Local lambda operations should still work as before

## ℹ️ Additional Information
Will send Sean a reminder to double check that these are set in the lambda environment when he's back Monday


jakewheeler (approved):
Looks good, thanks!

PR #962, opened 2026-03-17
chore: add more lambda logs for debugging

fzhao99:
adding some logs I wish I had during the debugging session with Tim.

If folks can se anything else they want to add, feel free to suggest.

### DIFF
--- a/refiner/app/lambda/lambda_function.py
+++ b/refiner/app/lambda/lambda_function.py
@@ -209,9 +209,8 @@ def process_refiner(
     reportable_groups = reportability_result["reportable_conditions"]
     logger.info(
         "Running refinement based on the following information:",
-        key="reportable_groups",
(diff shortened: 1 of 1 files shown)


refiner/app/lambda/lambda_function.py
```diff
@@ -254,14 +262,13 @@ def process_refiner(
                     # keep track of non active conditions for final processing
                     non_active_reportable_conditions[jurisdiction_code].add(rsg_code)
                     logger.info(
-                        "No active configuration for specified code, skipping.",
+                        "RSG code isn't in the CG map, skipping.",
                         key=rsg_code,
```
jakewheeler: I know you didn't change this but would you mind updating this?

```suggestion
                        rsg_code=rsg_code,
```
Must have been copy/paste problems 🙈

PR #1022, opened 2026-04-01
refactor: migrate to HeadlessUI `Input` component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR removes all usages of the `uswds-react` `TextInput` in favor of our own custom `TextInput` component, based off of the HeadlessUI `Input` component. This component also has default styling applied to it.

Since they were needed in a few circumstances, I've created (and used) custom `Field` and `Label` components with some default styling.

Additionally, I've updated the `Search` component to be based off of the HeadlessUI `Input` component as well.

## 🧪 How to test

- Modified components should appear and function the same as before.

### DIFF
--- a/client/src/components/Search/index.tsx
+++ b/client/src/components/Search/index.tsx
@@ -18,7 +18,7 @@ export function Search({
     >
       <SearchIcon />
       <Input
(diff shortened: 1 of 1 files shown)


fzhao99 (approved):
🧹

PR #1048, opened 2026-04-07
refactor: transition to HeadlessUI Modal component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR removes all usages of `@trussworks/react-uswds` modal-related components and fully on to our own custom, declarative modal component based on HeadlessUI's `Dialog` component.

I noticed while doing this work that there is a ton of duplicate code and complex components that could use some clean up but decided not to do it here. That will come in a later PR.

It's also worth noting that our modals were previously really inconsistent in terms of styling. This PR does address a lot of that but it does, of course, make some of the modals appear a little bit differently than before.

## 🧪 How to test

I've gone through and checked all of the modals myself to ensure that they look roughly the same as before and still function as they should.

As mentioned, you'll see that some of the styling inconsistencies have been rectified by using this shared component. A good example of how this can change the appearance is something like this:

Before:
image

After:
image

You'll notice:
- Title is now the expected size
- Modal sizing is more consistent
- Spacing between the header, body, and footer is now consistent across modal components
- USWDS styles aren't automatically being applied to the body text

As for testing you should see that:
- Automated tests should pass
- All modal components should look and work about the same as before

## ℹ️ Additional Information

- Does the modal component API seem flexible enough? I didn't want to make it _too_ flexible because we probably should try to make all of our modals look and function roughly the same.


client/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,161 @@
+import {
+  Dialog,
+  DialogBackdrop,
+  DialogPanel,
+  DialogTitle,
+} from '@headlessui/react';
+import { Icon } from '@trussworks/react-uswds';
```
rogeruiz, line 7: happy to see `react-uswds` isn't completely ripped out yet. but perhaps it'll be eventually in later work during this phase in the migration.

client/src/pages/Configurations/ConfigBuild/CustomCodes/ImportCustomCodes.tsx
```diff
@@ -74,9 +73,9 @@ export function ImportCustomCodes({
       system: 'ICD-10',
       name: '',
     });
-  const previewEditModalRef = useRef<ModalRef | null>(null);
-  const confirmModalRef = useRef<ModalRef | null>(null);
-  const undoModalRef = useRef<ModalRef | null>(null);
+  const [isPreviewModalOpen, setIsPreviewModalOpen] = useState(false);
```
fzhao99, line 76: nice, glad we can make this more idomatic with state setters

client/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+      <Modal open={isOpen} onClose={() => setIsOpen(false)}>
+        <ModalHeader>
+          <ModalTitle>My Custom Modal</ModalTitle>
+          <ModalBody>
+            <p>Content goes here</p>
+          </ModalBody>
+          <ModalFooter>
+            <Button onClick={() => setIsOpen(false)}>Close the modal</Button>
+          </ModalFooter>
+        </ModalHeader>
```
fzhao99: Think this closing tag needs to get moved up to 42

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
```diff
@@ -153,98 +161,65 @@ export function CustomCodeModal({
...
-            <Label>Code name</Label>
-            <TextInput
-              type="text"
-              value={form.name}
-              onChange={handleChange('name')}
-            />
-          </Field>
-        </div>
+    <>
+      <>
```
fzhao99: do we need both of these tag fragments?

client/src/components/Modal/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const ctx = useContext(ModalContext);
+  if (!ctx) {
+    throw new Error('Modal components must be used within <Modal>');
+  }
+  return ctx;
+}
+
+interface ModalProps {
+  open: boolean;
+  onClose: () => void;
```
fzhao99, line 27: this is a much cleaner interface 💪

fzhao99 (approved):
left a minor things, but otherwise lgtm!

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
jakewheeler: Nope! I'll fix it.

rogeruiz (approved):
lgtm 🌈 

Also @fzhao99 has added some other comments that I agree with. Great work on this @jakewheeler. Looking forward to more of this work being done.

PR #1051, opened 2026-04-08
chore: add custom codes regression test

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Adds an e2e to check that a custom code addition works in the independent testing flow. This is a regression test for a bug that got introduced that broke custom codes.

## 🔗 Related Issue

Fixes #1050 
- #1050

## ✅ Acceptance Criteria

- [ ] Test exists to test the custom code flow from creation to refinement
## 🧪 How to test
Run the e2e's and notice that the extra test for custom code addition / running in the independent testing flow is working


client/e2e/03 - independent-testing.spec.ts
```diff
@@ -319,4 +246,135 @@ test.describe('should be able to access independent testing', () => {
       inlineFlowInfluenzaResult
     );
   });
+
+  test('adds a custom code and checks the value gets refined', async ({
```
fzhao99, line 288: Putting this in the independent testing flow tests rather than in the custom codes tests since this relies on the creation / existence of the COVID and Influenza configs. Doing it this way minimizes the impact of possible race conditions, since there's a test-wide setup / teardown of the two configs, but there are some robustness checks for draft states / creation just in case

client/src/components/Button/index.tsx
```diff
@@ -35,6 +36,7 @@ export function Button({
       <Link
         onClick={sideEffect}
         to={to}
+        data-testid={props['data-testid']}
```
jakewheeler: I think it'd be ideal if we can avoid using `data-testid` when possible since we should generally be able to use some user-facing method that's available to us (`getByText`, `getByRole`, etc).

client/e2e/03 - independent-testing.spec.ts
```diff
@@ -74,72 +57,16 @@ test.describe('should be able to access independent testing', () => {
...
+    await createAndActivateCovidConfig(page);
+    await createAndActivateInfluenzaConfig(page);
 
     // go to independent testing flow
     await page.getByRole('link', { name: 'Testing' }).click();
 
     const independentFlowFileInput = page.locator('input#zip-upload');
     await independentFlowFileInput.setInputFiles(filePath);
 
-    await page.getByRole('button', { name: 'Refine .zip file' }).click();
```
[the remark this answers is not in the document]
jakewheeler: I think there's probably still some value in it since we have an actual DB backing the page in this test. It could catch an issue with the data being returned incorrectly that a component test couldn't catch.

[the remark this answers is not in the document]
fzhao99: added them back modulo some factoring improvements

client/e2e/03 - independent-testing.spec.ts
```diff
@@ -316,4 +284,111 @@ test.describe('should be able to access independent testing', () => {
...
+    await expect(page.getByText('Test Refiner')).toBeVisible();
+
+    await uploadTestFile(page);
+    await page.getByText('Refine eCR').click();
+
+    await expect(
+      page.getByRole('heading', { name: 'eCR refinement results' })
+    ).toBeVisible();
+
+    await expect(page.getByText('eICR file size reduced by')).toBeVisible();
```
fzhao99, line 349: With the implementation of lazy loading, there isn't a great way to try and match on the specific element in question that's being refined for, so opting instead to just check for the top-line refinement confirmation without the number. Figure we have plenty of other things for the actual refining logic, so hopefully this accomplishes the goal of the final test

jakewheeler (approved):
Thanks Bob!

PR #1053, opened 2026-04-08
fix: update fuzzy search for numbers

### LINKED ISSUE #960: [BUG] Tweak fuzzy search to differentiate between code search and display name search
## 📜 Description
Fuzzy search for numeric strings should be turned off because otherwise, all codes are returned since all codes contain numbers.

Ideally, we'd be able to exact match anything from the code search and fuzzy search on display name. Unsure if this is possible, or if we need to tweak the fuzzy search threshold. 

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0pmRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b9df535a2465e3d8fe8246353dd20834086746eb/Screen%20Recording%202026-03-25%20at%2012.55.55%E2%80%AFPM.mov)

## 👣 Steps to reproduce

1. Go to the build page
2. Try to search for a numeric string
3. See that there are fuzzy search results for the codes (ie 12345 brings up codes that have that as a substring, etc.)

## ??? Expected Result
Fuzzy search should only apply to display name, whereas exact search is being done for codes

#### Priority

- [x] **Medium priority** _Typically, medium-priority bugs do not affect customers directly and therefore can be fixed in the normal course of testing and development._
- [ ] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

## 📓 Notes
Unclear if there are ways of differentiating exact search by the column of data, but that'd be ideal! Otherwise, another solution that at least makes code search work in more cases would be good (double check with product if we get to this point)

lina-roth:
# 🔀 PULL REQUEST

## Search results on main, observe multiple errant matches
Screenshot 2026-04-08 at 12 57 13 PM

## Search results on my branch, observe only 5 results from prefix match
Screenshot 2026-04-08 at 12 56 15 PM

## 💡 Summary

Refined condition code search to improve handling of numeric/code queries.

- Detects code-like input and applies exact/prefix matching on code instead of fuzzy search
- Keeps existing Fuse fuzzy search for text-based queries

## 🔗 Related Issue

Fixes #960 
- #960


client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -190,16 +192,47 @@ export function ConditionCodeTable({
...
   }
 
-  // Show only the filtered codes if the user isn't searching
-  const visibleCodes = searchText ? results.map((r) => r.item) : filteredCodes;
+  function isCodeLikeQuery(value: string) {
+    const trimmed = value.trim();
+
+    // Treat inputs like 123, A12, J10, Z21.3 as code-like
+    // and avoid fuzzy matching them too loosely.
+    return /^[a-z0-9.-]+$/i.test(trimmed) && !trimmed.includes(' ');
```
fzhao99:
```suggestion
    return /^[a-z0-9.-]+$/i.test(trimmed) && !trimmed.includes(' ') && !/[a-z]{2}/i.test(trimmed);
```

I wonder if it's worth adding some regex for "are there two alphabetic characters next to each other" to try and have a faster match for the display name searches, since otherwise, regular text strings without spaces get matched on the exact prefix match. You can see below "coronavirus" doesn't trigger a match, but once there's a space, the values search as expected

https://github.com/user-attachments/assets/5611e8fc-8d06-4d13-b8e2-a526a5e4c7dc

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -190,16 +192,47 @@ export function ConditionCodeTable({
...
+    // For code-like input, do exact/prefix matching on code only
+    if (isCodeLikeQuery(trimmedSearch)) {
+      const normalizedSearch = trimmedSearch.toLowerCase();
+
+      return filteredCodes.filter((code) => {
+        const normalizedCode = String(code.code).toLowerCase();
+
+        return (
+          normalizedCode === normalizedSearch ||
+          normalizedCode.startsWith(normalizedSearch)
```
fzhao99:
```suggestion
          normalizedCode.includes(normalizedSearch)
```

any issues you foresee with changing this to inclusion rather than prefix matching? Think it might be nice for us to allow a search for `51.1` to match `F51.1`

fzhao99 (commented):
two suggestions for the code matching regex, but otherwise lgtm!

rogeruiz (approved):
lgtm 🌈 

Though, I only just saw @fzhao99 's comments and I think the first one regarding `coronavirus` should probably get fixed if not both.

fzhao99 (dismissed):
gating on the regex changes so this doesn't get merged accidentally

rogeruiz (approved):
lgtm 🌈 

Changes look good. Great work @lina-roth

PR #1056, opened 2026-04-09
refactor: migrate to HeadlessUI Button component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the existing `Button` component to be based on Headless UI's [Button](https://headlessui.com/react/button) instead of the `uswds-react` [Button](https://trussworks.github.io/react-uswds/?path=/docs/components-button--docs). It also replaces all usages of `<button>` elements throughout the app with the `Button` component.

Other changes:
- `Button` can now be rendered as a standard `a` tag for usages such as the `Log in` button
- `Button` now offers an `unstyled` variant for certain situations where we need non-standard button styling
- `Button` can now accept a `ref` (only two instances currently where we needed this)
- Table rows in the configurations table are now links instead of buttons
- Eliminated usage of `getByTestId` in tests that were relevant to this work
- Removed USWDS button styles

## 🧪 How to test

- Buttons should work and look the same as before


client/src/components/ConfigurationsTable/index.tsx
```diff
@@ -43,55 +41,33 @@ export function ConfigurationsTable({ data }: ConfigurationsTableProps) {
...
-                      void navigate(`/configurations/${id}/build`);
-                    }
-                  }}
-                  onClick={() => {
-                    void navigate(`/configurations/${id}/build`);
-                  }}
+                <Link
+                  aria-label={`Configure ${name}`}
+                  to={`/configurations/${id}/build`}
+                  className="flex items-center px-4 py-2 after:absolute after:inset-0 after:content-['']"
```
rogeruiz, line 55: Nice! Glad to see this cleaned up and removing the unnecessary `onKeyDown` event.

client/src/components/Diff/index.tsx
```diff
@@ -80,7 +80,8 @@ export function Diff({
           <div className="flex items-center gap-2">
             <span className="font-bold">Layout</span>
             <div className="border-blue-cool-60 flex gap-1 rounded-sm border p-0">
-              <button
+              <Button
```
rogeruiz, line 83: I didn't realize we were using plain buttons on this page. Glad to see this getting updated too.

rogeruiz (approved):
lgtm 🌈 

Great work on this refactor, @jakewheeler 🎉

PR #1062, opened 2026-04-10
fix: independent test file package download validation is overly strict

### LINKED ISSUE #1059: [BUG] Fix test flow download issue
## 📜 Description

File download for the Pertussis/Hep C eCR in the testing flows doesn't seem to work. 

## 👣 Steps to reproduce

1. Navigate to in-line (Pertussis config) or independent testing flow
2. Upload the Hep C/Pertussis test eCR
3. Attempt to download results

## 🌃 Environment:

AIMS dev

## ??? Expected Result

Output files should download as expected.

## 🧪 Actual result

Attempting to download the files results in an error.

#### Priority

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes an issue where the pertussis sample file zip package was failing to download in the independent test flow. Please see my comment for detail on exactly why this was happening.

## 🔗 Related Issue

Fixes #1059 

## 🧪 How to test

1. Create a Pertussis configuration
2. Try running the Pertussis sample files through the independent testing flow (grab from `refiner/assets/demo`)
3. Try downloading the file package using the `Download results` button. You should see that it's successful and be able to open the zip package

If you want to check that the error messaging works, try adding something like this to the route in `demo.py`:
```python
@router.get(
    "/download/{filename}",
    tags=["demo"],
    operation_id="downloadRefinedEcr",
)
async def download_refined_ecr(
    ...
) -> StreamingResponse:
    ...

    raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Problem occurred.")

    if not SAFE_FILENAME_RE.match(filename):
        ...
```

When you try to download again you'll see something like this:
image

### DIFF
--- a/client/src/components/Diff/index.tsx
+++ b/client/src/components/Diff/index.tsx
@@ -3,6 +3,7 @@ import ReactDiffViewer, { DiffMethod } from 'react-diff-viewer-continued';
 import { IndependentTestUploadResponse } from '../../api/schemas';
 import classNames from 'classnames';
 import { Button } from '../Button';
+import { getDownloadRefinedEcrQueryKey } from '../../api/demo/demo';
 
 type DiffProps = Pick<
   IndependentTestUploadResponse,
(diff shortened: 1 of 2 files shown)


refiner/app/api/v1/demo.py
```diff
@@ -254,21 +263,11 @@ async def download_refined_ecr(
...
-    if "/" in filename or "\\" in filename:
+    if not SAFE_FILENAME_RE.match(filename):
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST, detail="Invalid filename."
         )
 
-    uuid_prefix = filename.removesuffix(".zip")
-
-    try:
-        UUID(uuid_prefix)
```
jakewheeler, line 265: The sample pertussis document was failing this check because the last part of the string had 13 characters instead of 12. I assume this was a mistake during its creation.

While that case appears to be a mistake, I think it generally doesn't make sense for us to validate this because we have no control over what document ID is coming out of the file in a real-life scenario. I think it makes sense to check that the filename contains only allowed characters (which I've added) but the app shouldn't perform additional checks beyond that.

client/src/components/Diff/index.tsx
```diff
@@ -48,14 +58,18 @@ export function Diff({
...
-            <div className="gapx-2 flex flex-col items-start py-1">
+            <div className="flex flex-col items-start gap-1 py-1">
               <Button
                 variant="tertiary"
                 onClick={() => downloadFile(refined_download_key)}
               >
                 Download results
               </Button>
-              {downloadError && <span>File download has expired.</span>}
+              {downloadError ? (
```
jakewheeler, line 68: This will actually be useful now and give the user the exact reason the download is failing 🙃

fzhao99 (approved):

PR #1063, opened 2026-04-10
refactor: migrate to HeadlessUI Tooltip component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a `Tooltip` component built on top of HeadlessUI's `Transition` component. The styling and behavior should be very close to the USWDS `Tooltip` we were using previously.

## 🧪 How to test

Tooltips are currently limited to the sections screen. You should see tooltips on icon hover like before.
image

You should also be able to use your keyboard to tab through to the icon and see the label when focused. It will close the message when focus is lost.
image


client/src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,89 @@
+import { useState, useRef, useId } from 'react';
```
jakewheeler, line 1: This is a lot more basic than the `uswds-react` [implementation](https://github.com/trussworks/react-uswds/blob/main/src/components/Tooltip/Tooltip.tsx) but seems like we can get away with it currently. We could always consider pulling in a tooltip component down the road if things need to get more complex in the future for whatever reason.

rogeruiz (approved):
lgtm 🌈 

Nice and clean. And hopefully we can get away with a simple implementation for a while.

PR #1064, opened 2026-04-10
chore: add `@components` alias for `src/components` and update imports

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a `@components` alias so that when importing from `src/components` we can now import like this:
```ts
import { Tooltip } from '@components/Tooltip';
```

Instead of like this:
```ts
import { Tooltip } from '../../../../components/Tooltip';
```

While primarily stylistic, it should also make moving files around easier in some situations since we don't need to reference the exact path in the import statements anymore.

## 🧪 How to test

This should not introduce any changes in the app.


fzhao99 (approved):
🧹

PR #1066, opened 2026-04-10
feat(server): add section processing to narrative, create CDA R2 conformant narrative in augmented output

### LINKED ISSUE #980: [TASK] Add CDA SDTC XSD validation to the refiner
## 🚧 What needs to be done

Add the XSD files to the validation workflow within the `/scripts` directory and also add these checks to the integration tests for both the app and lambda's outputs.

## 🤔 Why it needs to be done

There's a possibility that some errors from SDTC will keep files from working with downstream processes. This _could_ affect augmented spec but we need to look into this first.

> [!IMPORTANT]
> If the current narrative related functions fail the validation; make changes that get them in compliance with the schema.

## 🏗️ Acceptance Criteria

- [ ] ...

## ⚙️ Technical Notes

- ...

## 📜 Additional context

XSD files are located here: https://github.com/HL7/CDA-core-2.0/tree/master/schema/extensions/SDTC

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Reorganizes `services/ecr/process_eicr.py` and `services/ecr/specification.py` into focused packages (`section/` and `specification/`), then layers two follow-on improvements on top: per-section provenance footnotes that travel with every refined eICR, and a `SectionOutcome` `enum` that surfaces what the refiner actually did versus what the jurisdiction configured. Also fixes a latent bug where the `(retain, narrative=False)` configuration was silently ignored.

## 🔗 Related Issue

Fixes #980 
- #980


refiner/app/services/ecr/section/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+        section=section,
+        codes_to_match=codes_to_match,
+        namespaces=namespaces,
+        section_specification=section_specification,
+        code_system_sets=code_system_sets,
+        include_narrative=include_narrative,
+    )
+
+
+__all__ = [
```
fzhao99, line 101: think this is the first time I'm seeing this export pattern, so wondering when / where we should use this to explicitly define the public export interface?

refiner/app/services/ecr/section/constants.py
```diff
@@ -0,0 +1,88 @@
...
+
+PROVENANCE_OUTCOME_NOTES: Final[dict[SectionOutcome, str]] = {
+    SectionOutcome.REMOVED_BY_CONFIG: "Removed by configuration",
+    SectionOutcome.RETAINED: "Retained as-is",
+    SectionOutcome.RETAINED_NARRATIVE_REMOVED: "Retained; narrative removed",
+    SectionOutcome.REFINED_WITH_MATCHES: "Refined; matches found",
+    SectionOutcome.REFINED_NARRATIVE_REMOVED: "Refined; narrative removed",
+    SectionOutcome.REFINED_NARRATIVE_RECONSTRUCTED: "Refined; narrative reconstructed",
+    SectionOutcome.REFINED_NO_MATCHES_STUBBED: "Refined; no matches found",
+}
```
fzhao99, line 88: 💪

refiner/app/services/ecr/section/entry_matching.py
```diff
@@ -0,0 +1,514 @@
...
+    only elements without @code, the next rule is tried.
+    """
+
+    entry_matches: list[EntryMatch] = []
+
+    for rule in match_rules:
+        # evaluate primary code xpath
+        code_elements = cast(
+            list[_Element],
+            entry.xpath(rule.code_xpath, namespaces=namespaces),
```
fzhao99, line 225: is there a world where we'd need to factor in the value of the of the rule code system OID in the actual XPath with something like `hl7:code[@codeSystem='{rule.CODE_SYSTEM_OID}']`? Thinking of instances where a rule without the codesystem check "claims" an entry, but then doesn't match on the coding check in 232-47 because of a system mismatch, only to have subsequent rules be skipped because the entry is claimed already.

think this is what I found in [this PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1088/changes). Maybe this is covered by the spec conformance where system OID is a known quantity (which is maybe what you meant by Tim swapping the system OID's for the immunization bug?), but flagging in case it raises anything

fzhao99 (approved):
left a few questions, but generally am a fan of the file reorganization and refactoring! think the new footnotes will be quite helpful

refiner/app/services/ecr/section/__init__.py
robertmitchellv: hopefully we'll have more time to do some clean up around what should be public vs private--i saw this example when i was trying to figure out if i should make one module, many modules, etc. trying to find the right organization

refiner/app/services/ecr/section/constants.py
robertmitchellv: we still need to build reconstructed but this at least sets us up for that

refiner/app/services/ecr/section/entry_matching.py
robertmitchellv: removing the OID helps us not have to make as many fixed rules since the major thing with immunizations we IG says that they're supposed to use CVX a lot of them don't use that system so they nullFlavor the code or value and then add an RxNorm translation. when we make the OID a part of the xpath that enforces the really strict rule around:

one, is this in the right order?
two, is this just flipped but correct?
three, is the outside nullFlavor and translation the match?
...any other fallback here

i do think that our matching algo **should** be template driven but maybe not in this way at the code system level--more in the identification of the pattern that is being use in CDA. there **could** be collisions with custom codes but i think we're mostly safe with collisions--especially since a code in our set still means "we need to keep this" hopefully this helps but maybe i wasn't totally understanding your question

refiner/app/services/ecr/section/__init__.py
jakewheeler: From what I understand this is a convention typically used in libraries rather than in app code.

PR #1080, opened 2026-04-13
refactor: migrate to HeadlessUI Combobox

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

Removes the usage of the uswds-react `ComboBox` in favor of a custom component based on Headless UI's `Combobox`.

The only place we use a combo box currently is when creating a new configuration (selecting a condition).

> [!NOTE]
> This component is more complex than some of the other components created recently due to needing to match the features offered by the uswds-react combobox. I've added a test file to make sure the expected behavior is working correctly.

## 📺 Demo
Kapture 2026-04-14 at 13 52 32

## 🧪 How to test

- Experience when creating a new configuration should be very similar to how it was prior to this change


client/src/pages/Configurations/index.tsx
```diff
@@ -135,29 +154,28 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
-          <div>
-            <USWDSLabel
-              htmlFor="new-condition"
-              className="leading-6!"
-              data-focus="true"
+          <Field>
+            <Label>Select condition</Label>
+            <Combobox
+              value={selectedCondition}
+              virtual={{ options: filteredConditions }}
```
jakewheeler, line 161: I noticed some performance issues without using this prop. More info about the `virtual` prop [here](https://headlessui.com/react/combobox#virtual-scrolling).

If/when we need a Combobox with less options we likely will not need to use this.

client/tests/setup.ts
```diff
@@ -59,3 +60,5 @@ Object.defineProperty(navigator, 'sendBeacon', {
   writable: true,
   value: vi.fn(),
 });
+
+mockAnimationsApi();
```
jakewheeler, line 64: HeadlessUI is using a browser API for animations that needs to be made available in jsdom when testing. This warning appears when running UI tests without this library being installed.

```
Headless UI has polyfilled `Element.prototype.getAnimations` for your tests.
Please install a proper polyfill e.g. `jsdom-testing-mocks`, to silence these warnings.

Example usage:

import { mockAnimationsApi } from 'jsdom-testing-mocks'
mockAnimationsApi()

Headless UI has polyfilled `Element.prototype.getAnimations` for your tests.
Please install a proper polyfill e.g. `jsdom-testing-mocks`, to silence these warnings.

Example usage:

import { mockAnimationsApi } from 'jsdom-testing-mocks'
mockAnimationsApi()

```

rogeruiz (approved):
lgtm 🌈 

Looks good. Also seems to have fixed a nasty a11y issue when it comes to tabbing and selecting.

PR #1079, opened 2026-04-13
fix: Support multiple chip architectures for dev

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

I ran into issues running Docker Compose with my new laptop's chip architecture.
This helps with that by reading from an environment variable
`DOCKER_DEFAULT_PLATFORM`. If it doesn't exist, it will will default to the
`linux/amd64` platform which is what it would default to prior to this patch.

<details open>
<summary>Example of error(s) I was encountering locally</summary>

The specific issue was with running `just database refresh`. I get the following
error prior to this patch.

```
just database refresh

just db clean && just migrate local && just db seed
docker compose exec db psql -U postgres -d refiner -f drop-all.sql
DROP SCHEMA
CREATE SCHEMA
GRANT
GRANT
just -f migrate.just run local migrate
sh -c '../sh/get_db_url.sh local | xargs -I {} docker compose exec -T migrate dbmate --migrations-dir /app/refiner/migrations --schema-file /app/refiner/schema.sql --url {} migrate'
Applying: 20260226212322_initial_schema.sql
Applied: 20260226212322_initial_schema.sql in 26.004494ms
Applying: 20260305221859_add_new_section_props.sql
Applied: 20260305221859_add_new_section_props.sql in 1.393349ms
Applying: 20260309001833_add_cvx_code_system.sql
Applied: 20260309001833_add_cvx_code_system.sql in 655.716µs
Applying: 20260309143000_add_bulk_add_custom_code_event_type.sql
Applied: 20260309143000_add_bulk_add_custom_code_event_type.sql in 448.797µs
Applying: 20260309151338_add_config_sections_table.sql
Applied: 20260309151338_add_config_sections_table.sql in 6.568281ms
Applying: 20260318125201_custom_sections.sql
Applied: 20260318125201_custom_sections.sql in 3.000116ms
Applying: 20260402215326_add_coverage_status.sql
Applied: 20260402215326_add_coverage_status.sql in 3.942501ms
Writing: /app/refiner/schema.sql
docker compose run --rm seed
WARN[0000] Found orphan containers ([dibbs-erc-refiner-auth-run-467d0d81607f]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
[+]  1/1t 1/11
 ✔ Container dibbs-erc-refiner-db-1 Running                                                                      0.0s
Container dibbs-erc-refiner-seed-run-86d5f7226c0c Creating
[...]

PR #1081, opened 2026-04-13
fix: Removes `platform:` stanza and updates documentation around VM resources

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

During the review of #1079, it was brought up that we don't need to define the
`platform:` stanza for the `refiner-service`. This patch removes that stanza and
also adds some documentation to ensure that developers use a host VM with enough
resources for our scripts and applications to run.

## 🔗 Related Issue

Supersedes #1079

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

Everything should run as expected across different architectures without issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,13 @@ For more detailed information about the relationship between the eICR and RR doc
 
 ## Running the project locally
 
+> [!TIP]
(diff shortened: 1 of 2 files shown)


jakewheeler (approved):
Thanks!

PR #1089, opened 2026-04-14
chore(test): navigate e2e auth flow only once

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR modifies the e2e tests so that they acquire a single Refiner session through Keycloak and then reuse it across all tests. This prevents the tests from needing to re-authenticate over and over as they run. This should help to make the test suite a little more reliable when running in the pipeline since the Keycloak service can be flaky at times in this environment.

## 🔗 Related Issue

Part of the app testing ticket.

## 🧪 How to test

If you want to try this locally:

1. Run `npm run e2e:dev`
2. Try running a single test or all tests via the testing UI
3. You should see `auth.test.ts` will always run first, which authenticates and then stores the session to `e2e/.auth/user.json`
4. Test(s) should pass as normal

### DIFF
--- a/client/playwright.config.ts
+++ b/client/playwright.config.ts
@@ -18,8 +18,18 @@ export default defineConfig({
   },
   projects: [
     {
-      name: 'Chromium',
+      name: 'setup',
+      testMatch: /auth\.setup\.ts/,
       use: { ...devices['Desktop Chrome'] },
+    },
+    {
+      name: 'e2e',
+      testMatch: /.*\.spec\.ts/,
+      dependencies: ['setup'],
+      use: {
+        ...devices['Desktop Chrome'],
+        storageState: 'e2e/.auth/user.json',
+      },
     },
   ],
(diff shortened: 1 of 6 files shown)


client/playwright.config.ts
```diff
@@ -18,8 +18,18 @@ export default defineConfig({
...
     {
-      name: 'Chromium',
+      name: 'setup',
+      testMatch: /auth\.setup\.ts/,
       use: { ...devices['Desktop Chrome'] },
     },
+    {
+      name: 'e2e',
+      testMatch: /.*\.spec\.ts/,
+      dependencies: ['setup'],
```
jakewheeler, line 28: This is how the `setup` project always executes first.

fzhao99 (approved):
thanks for figuring this out! I remember banging my head against this wall back when we made these changes, so glad we came back around to it

PR #1094, opened 2026-04-15
fix: Live environment test flows are over-refining

### LINKED ISSUE #1082: [BUG] Adding code sets can drastically change refinement results when it shouldn't
## 📜 Description

Adding a code set to a configuration - for example adding Agricultural Chemicals (Fertilizer) Poisoning (9 codes) to COVID - boosts the size reduction metric singificantly (ex: 39% to 71%). Removes relevant data, affecting Encounters, Procedures, Problems, Results, Immunizations, potentially others.

## 👣 Steps to reproduce

1. Load up AIMS dev
2. Test COVID with the default set-up (refine all data, retain all narrative)
3. Observe size reduction and retained data
4. Add a code set (ex: Agricultural Chemicals (Fertilizer) Poisoning)
5. Observe size reduction and retained data

## 🌃 Environment:

- AIMS dev
- version: db36f9f

## ??? Expected Result

Expectation is reduction % would stay the same or decrease since codes are added.

## 🧪 Actual result

Reduction % sometimes increases dramatically.

#### Priority

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes a bug that was noticed during refinement in the app where we were seeing a lot more file content refined out than what was expected.

An example of what we were seeing: when running a test using a COVID-19 configuration that also had the Agricultural Chemicals (Fertilizer) Poisoning code set associated with it, we would see a percent reduction in the 70s vs. the 36% we were expecting (at least initially, I've corrected this % here as well).

The reason this was happening was because of two issues:
1. Conditions could be returned in any order (this is why we saw a difference locally vs. in a live environment)
2. `ProcessedConfiguration.from_payload()` had a bug when building the `coding_by_code_system` where the coding list was being re-assigned instead of extended

So what was happening in live environments was this:
1. Conditions are fetched from the DB, COVID-19 is returned first and Agricultural Chemicals (Fertilizer) Poisoning is second
2. `ProcessedConfiguration.from_payload()` loops through conditions to build out `coding_by_code_system`
3. Instead of adding ALL ICD-10 codings to the list, we overwrite the COVID-19 ICD-10 codings (these came first) and keep only the Agricultural Chemicals (Fertilizer) Poisoning codings (which come second and "win")

## 🔗 Related Issue

Fixes #1082 

### View the current issue

1. Log into https://refiner.dev.sandbox2.aimsplatform.org/
2. Create (or use an existing draft) COVID-19 configuration that has  the `Agricultural Chemicals (Fertilizer) Poisoning` code set associated with it
3. Run the `TEST_eCR_Refiner_COVID_Influenza_Sample_Files.zip` file through the inline test flow
4. You should get a 71% reduction which is incorrect

### View the fix in this branch

1. Log into https://refiner.dibbs.tools/
2. Follow step 2 above
3. Run the `SKL_eCR_Refiner_COVID_Influenza_Sample_Files.zip` file through the inline test flow
4. You should get a 44% reduction, which is what we expect


refiner/app/services/terminology.py
```diff
@@ -452,12 +452,16 @@ def from_payload(cls, payload: ConfigurationPayload) -> "ProcessedConfiguration"
...
+                ].extend(
+                    [
+                        asdict(
+                            Coding(
+                                code=c.code, display=c.display, system=code_system.oid
+                            )
+                        )
+                        for c in code_list
+                    ]
+                )
```
jakewheeler, line 464: This is the fix. We need to add to this list instead of re-assigning it.

refiner/app/api/v1/configurations/testing.py
```diff
@@ -249,8 +249,8 @@ async def run_configuration_test(
             stats=[
                 f"eICR file size reduced by {
                     get_file_size_reduction_percentage(
-                        unrefined_eicr=formatted_unrefined_eicr,
-                        refined_eicr=formatted_refined_eicr,
+                        unrefined_eicr=original_xml_files.eicr,
+                        refined_eicr=refined_document.refined_eicr,
```
jakewheeler, line 253: This is consistent with how it gets calculated in "trace".

fzhao99 (approved):
thanks for figuring this out!

PR #1096, opened 2026-04-15
fix: fix rendering bugs

fzhao99:
Needed a quick win after doing some debugging this afternoon, so took care of the bugs Kyle discovered [here](https://skylight-hq.slack.com/archives/C08H7EV6M37/p1776285033678589)

1. [Safari seems to have trouble with the `:after` psuedo element](https://skylight-hq.slack.com/archives/C08H7EV6M37/p1776285033678589), so adding some positioning to the link element seems to have fixed things

https://github.com/user-attachments/assets/99b6ca3b-ae5a-4efd-808a-37b0cb3da294

2. Tweaked some validation / formatting logic around to
    - Validate on both blur and update
    - Always set the code on update (as opposed to only when there isn't an error) so users can always see what they're typing
    - Set an error in the event that there's a dupe (and thus gate saving / updating)
    - Only clear the error when there's an update to the code 
    - Do a final trimming of name and code on save / update

Think this is the most intuitive set of updates, but if folks can think of an issue, lmk

https://github.com/user-attachments/assets/8127b7e0-417f-4c07-a0d0-b0567ecdba89


rogeruiz (approved):
lgtm 🌈 

looks much nicer to read and the functionality is still there. good work on this @fzhao99

PR #1106, opened 2026-04-17
test: run e2e tests against a production-like app build

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

The goal of this PR is to run the end-to-end Playwright tests against a container built from `Dockerfile.app` instead of running them against the local development environment. In development, we run the client as its own container which proxies the backend running in its own container.

With this change we can feel more confident that the e2e tests are running against the image that we may ship to production.

Additionally, when running the e2e tests in CI, we now make use of `Dockerfile.ops` to migrate and seed the database. This will enable us to catch any issues with this container that we would have previously missed since we were never running it in CI before this.

### How this works

This PR splits up the Docker Compose setup into three files:

- `docker-compose.yml` (the base configuration)
- `docker-compose.override.yml` (our local dev environment configuration)
- `docker-compose.ci.yml` (the CI environment configuration)

Luckily, this doesn't change much for local development. We can still simply run `docker compose up` to start the local environment up and everything should work as normal. Docker will automatically merge `docker-compose.yml` with `docker-compose.override.yml` for us.

CI is a bit different since we need to specify which two files to merge (base and CI). See the updates to the GH Action to understand what this looks like.

## 🔗 Related Issue

Testing improvements ticket.

## ✅ Acceptance Criteria

1. Shut down your containers
2. Switch to this branch and run `docker compose up`
3. You should see all of the containers come up as usual
4. Run `just db refresh` to be sure this command still works
5. Please try any other local commands you use to be sure those also still work as expected

If you'd like, you could also try shutting down your containers and then running `pytest tests` from within `/refiner`. This should start up the local environment, seed the database, and run the tests as you'd expect.


.github/workflows/tests.yml
```diff
@@ -93,28 +96,28 @@ jobs:
...
             ${{ runner.os }}-playwright-
             ${{ runner.os }}-
+
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps
         working-directory: client
+
       - name: Start Application
-        run: just dev up --detach
+        run: docker compose -f docker-compose.yml -f docker-compose.ci.yml up --detach --build
```
jakewheeler, line 105: While this doesn't need to be run locally, it does actually give us a convenient way to build/run our prod containers locally which is a little bit of a pain to do prior to this.

refiner/tests/integration/conftest.py
```diff
@@ -192,8 +192,10 @@ def setup(request):
 
     print("🚀 Setting up tests...")
     path = Path(__file__).resolve().parent.parent.parent.parent
-    compose_file_name = os.path.join(path, "docker-compose.yaml")
-    refiner_service = DockerCompose(path, compose_file_name=compose_file_name)
+    refiner_service = DockerCompose(
+        path,
+        compose_file_name=["docker-compose.yml", "docker-compose.override.yml"],
```
jakewheeler, line 197: `DockerCompose` will not automatically merge the files for us like the `docker compose` command will.

docker-compose.yml
```diff
@@ -1,29 +1,6 @@
...
-      - ./refiner:/app
-    ports:
-      - "8080:8080"
-    extra_hosts:
-      - "localhost:host-gateway"
-    environment:
-      - PIP_ROOT_USER_ACTION=ignore
-      - ENV=local
-      - VERSION=local
+  server:
```
jakewheeler, line 4: Taking this opportunity to finally rename `refiner-service` to `server`.

docker-compose.override.yml
fzhao99, line 1: could we just name this file `docker-compose.local.yml` and maybe add a comment in the relevant line of the conftest that we want to override the default setup for testing? Wasn't immediately obvious to me that this file was replacing the old compose with the new local setup, and it seems like the "override" idea is only relevant in that one place

jakewheeler: The [override naming](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#additional-information) is what Docker automatically looks for rather than something I decided on to be clear.

I was thinking this setup made the most sense so that locally we can continue to boot up the environment using this command:
```sh
docker compose up -d
```

If we made the suggested name change we'd lose the automatic merging and instead need to run:
```sh
docker compose -f docker-compose.yml -f docker-compose.local.yml up -d
```

Does that makes sense? I can definitely add a comment to each of the files if that'd help clarify their purposes.

fzhao99: ahh ok. a comment would be great in that case!

.justscripts/just/database.just
```diff
@@ -1,92 +1,90 @@
...
 
-[group('db')]
 [doc('Runs a full database refresh (wipe, migrate, seed)')]
+[group('db')]
 refresh:
     just db clean && just migrate local && just db seed
 
-[group('db')]
 [doc('Run all DB sanity checks (seeding, integrity, etc)')]
+[group('db')]
```
fzhao99, line 37: is there any linting setup for just that you have configured? guessing we have a dueling config issue that is resulting in these whitespace changes

fzhao99 (approved):
lgtm! thanks so much for cleaning all these pieces of up 🧹

.justscripts/just/database.just
jakewheeler: Whoops I didn't notice this happened. Looks like I have [this extension](https://marketplace.visualstudio.com/items?itemName=nefrob.vscode-just-syntax) installed. Should I undo?

fzhao99: i think as long as we all have the same setup, I'm fine with it. We can add this to the relevant VSCode file in a followup

fzhao99: https://github.com/CDCgov/dibbs-ecr-refiner/pull/1109/

jakewheeler:
@fzhao99 Added comments to the top of each of the DC files and also added some notes about this setup to the README. Thanks for the review!

PR #1108, opened 2026-04-20
test: Python integration test improvements

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR makes a few changes, primarily focused on Python integration testing:

- Migration script has been added to delete entries from `configurations_locks` and `events` when a configuration is deleted
  - Updated a couple of queries related to this change
  - Improved `events` constraint naming
- Added new fixtures to `integration/conftest.py` to simplify common testing needs
- Configuration data will be automatically cleaned up between individual test runs, similar to how the e2e tests work
- Resolved an occasional issue related to Postgres caching (tests that work would randomly fail due to this)
- Reduced custom SQL used in `integration/conftest.py` for seeding test data
-  Updated integration tests to use the new fixtures
- Updated test content where it seemed needed

## 🔗 Related Issue

Testing improvements ticket.

## 🧪 How to test

Running the Python tests should work as expected.

```sh
pytest tests -v
```

## ℹ️ Additional Information

Are there other improvements you'd like to see as a part of this work?


refiner/migrations/20260420140437_config-delete-cascade.sql
```diff
@@ -0,0 +1,28 @@
...
+-- migrate:down
+ALTER TABLE configurations_locks
+  DROP CONSTRAINT configurations_locks_configuration_id_fkey,
+  ADD CONSTRAINT configurations_locks_configuration_id_fkey
+    FOREIGN KEY (configuration_id)
+    REFERENCES configurations(id);
+
+ALTER TABLE events
+  DROP CONSTRAINT fk_events_configurations,
+  ADD CONSTRAINT fk_events_configurations
```
jakewheeler, line 26: Note that I've kept the improved name even if we run a `down`. No real reason to switch back to the old name.

refiner/migrations/20260420140437_config-delete-cascade.sql
```diff
@@ -0,0 +1,28 @@
+-- migrate:up
+ALTER TABLE configurations_locks
+  DROP CONSTRAINT configurations_locks_configuration_id_fkey,
+  ADD CONSTRAINT configurations_locks_configuration_id_fkey
+    FOREIGN KEY (configuration_id)
+    REFERENCES configurations(id)
+    ON DELETE CASCADE;
```
fzhao99, line 7: 💪

refiner/tests/integration/conftest.py
```diff
@@ -59,8 +59,125 @@
 TEST_JD_STATE_CODE = "GC"
 
 
+@pytest_asyncio.fixture
+async def activate_config(authed_client):
```
fzhao99, line 63: nice, love the way we're setting these fixtures up with hooks for the individual tests to setup / tear down their data

refiner/tests/integration/test_configurations.py
```diff
@@ -494,23 +447,27 @@ async def test_transaction_rollback_on_activation_failure(self, db_pool):
             new_config = await get_configuration_by_id_db(
                 id=new_config_id, jurisdiction_id="SDDH", db=db_pool
             )
-            assert new_config.status == "inactive"  # Should remain inactive
```
fzhao99, line 497: is there a reason this was passing with inactive before? think what you changed it to was would I would have expected

fzhao99 (approved):
thanks for working on this! glad all our tests are in much better shape now 💪

refiner/tests/integration/test_configurations.py
jakewheeler: That's a great question. In the previous setup the config data was shared between all tests since the database was never reset in any way. By the time we got to this test, there was some other test that ran before this one that modified the config we we were using here to be `inactive`, which is why we were trying to promote the config from `inactive` to `active`.

Since this test is now in control of creating its own configuration data we can simply create a draft and then attempt to promote it to `active` instead of doing the confusing work-around we were doing before.

PR #1109, opened 2026-04-20
chore: add just linter extension and settings

fzhao99:
add formatter and setting for just as a follow up to [this thread](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1106#discussion_r3111297372)

cc @rogeruiz  since you're the most likely to have opinions on this!


.justscripts/just/dev.just
```diff
@@ -1,31 +1,32 @@
...
 
 alias u := up
 alias d := down
 
-[group('docker')]
 [doc("Start only the project's core services with docker compose.")]
+[group('docker')]
 up *ARGS:
-    {{ docker }} compose -f {{ absolute_path('./docker-compose.yaml') }} up {{ ARGS }}
+    {{ docker }} compose {{ compose_files }} up {{ ARGS }}
```
fzhao99: these changes are non-whitespace after I realized the new override setup doesn't automatically get accounted for

jakewheeler: I didn't realize we were doing this. Should we follow this format instead? Using `-f` is why this isn't automatic.

```suggestion
    {{ docker }} compose up {{ ARGS }}
```

rogeruiz: I like this approach you've done here @fzhao99.

Also @jakewheeler will no `-f` flags mean that it'll also pick up the `docker-compose.overrides.yml`? I thought it would only automatically pull in `docker-compose.yml`.

rogeruiz (approved):
lgtm 🌈 

Tested this locally and things are looking good. Thanks for setting this up @fzhao99

.justscripts/just/dev.just
jakewheeler: Yes, there's a note in Docker's [docs](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#additional-information) about this.

PR #1130, opened 2026-04-21
feat(ui, api): Show TES Code Set completeness with Status Badge component and Modal

### LINKED ISSUE #819: [USER STORY] TES Code Set Completeness
## 👤💭 User Story

As a user, I want to know whether the default code set for a condition is ready for use or not, so that I know how much information I may need to provide myself

## 🏗️ Acceptance Criteria

- [x] Indication whether TES code set is not expanded, partially complete, or fully complete
- [x] Details per additional context grouper of whether they're included or not
- [x] Should be properly associated with the variables from the TES code set of the version being viewed
- [x] Show "not expanded" flag for code sets that don't have the completeness variable
- [x] Can ignore Version in modal for now

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=6946-43440&t=UHSnUxraCMyUBLAf-1

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a status badge to every code set on the configuration build page to indicate its level of completeness. It also adds a `Details` button next to the badge to indicate the completeness of the context groupers associated with the code set.

> [!IMPORTANT]
> We decided to drop both the date element and version element that are shown in the modal design. These elements will be reevaluated and may show up again in a future ticket. Let me know if you have any questions about this.

## 📺  Demo
Kapture 2026-05-06 at 09 09 25

## 🔗 Related Issue

Fixes #819

## 🧪 How to test

- Feel free to try recreating the demo config above yourself
- Take a look at the e2e test included in this PR

### Older condition versions

If an older condition version is being used for a configuration (like from TES `1.0.0`) the code set badge will say `not expanded` and all of the individual statuses in the modal will be set to `not included`.
image
image


client/src/components/StatusBadge/index.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
```
jakewheeler: Can we please move this file into `ConfigBuild/CustomCodes` so it's closer to where it is used? I don't know that we're fully there yet but I think longer term we will use `components` exclusively for shared components.

jakewheeler: Moving all of this into `ConfigBuild/CustomCodes/index.tsx` would also be fine.

client/src/components/StatusBadge/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  NotExpanded = 'not_expanded',
+  PartiallyIncomplete = 'partial',
+  FullyComplete = 'complete',
+}
+
+// TODO: Status will eventually come from the server response instead of this enum.
+
+export interface StatusBadgeProps {
+  text: string;
+  status: BadgeStatus;
```
jakewheeler: I'd suggest avoiding the enum here and using string literals.
```suggestion
  status: 'incomplete' | 'partial' | 'complete';
```

It won't matter if we make the switch but did want to note I'm seeing an linting error coming from the enum.
```
Fast refresh only works when a file only exports components. Use a new file to share constants or functions between components.eslint(react-refresh/only-export-components)

enum BadgeStatus
```

client/src/components/StatusBadge/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  className?: string;
+}
+
+const BG_BY_STATUS: Record<BadgeStatus, string> = {
+  [BadgeStatus.NotExpanded]: 'bg-red-400',
+  [BadgeStatus.PartiallyIncomplete]: 'bg-yellow-300',
+  [BadgeStatus.FullyComplete]: 'bg-green-300',
+};
+
+export const StatusBadge = ({
```
jakewheeler: Can we please change this to use `function`?

client/src/components/StatusBadge/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  detailsContent,
+  className = '',
+}: StatusBadgeProps) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <div className="inline-flex items-center pb-4">
+      <span
+        className={`rounded-full px-3 py-1 font-medium text-black lowercase ${BG_BY_STATUS[status]} ${className} `}
+        data-testid="status-badge"
```
jakewheeler: Can we remove this please? It'd be good to move away from using test IDs when we can.

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -239,8 +240,35 @@ export function ConditionCodeTable({
...
+              <p>{coverage.coverage_level_reason}</p>
+              {coverage.coverage_level_date && (
+                <p className="text-sm text-gray-600">
+                  As of {coverage.coverage_level_date}
+                </p>
+              )}
+            </div>
+          }
+        />
+      )}
```
jakewheeler: I'd say go ahead and bake as much of this into the component as we can since it's being created specifically for a single purpose anyway. For example, we could pass in only `coverage` and let the component determine what it needs to render. It'd make using this component a lot simpler.

refiner/app/db/conditions/db.py
```diff
@@ -184,12 +187,15 @@ async def get_condition_codes_by_condition_id_db(
                 FROM conditions
                 WHERE id = %s
             )
-            SELECT DISTINCT code, system, description
+            SELECT DISTINCT code, system, description, coverage_level, coverage_level_reason, coverage_level_date
```
jakewheeler: What is the reasoning to do this at the individual condition code level rather than the condition code set level?

refiner/tests/unit/test_conditions.py
```diff
@@ -100,3 +101,59 @@ async def fake_get_condition_by_id_db(id, db):
 
     assert response.status_code == status.HTTP_404_NOT_FOUND
     assert response.json()["detail"] == "Condition not found."
+
+
+@pytest.mark.asyncio
+async def test_get_condition_coverage_data(monkeypatch, authed_client):
```
jakewheeler: I would highly suggest we scrap unit testing here and prefer an e2e test instead since this is a 100% web app only feature that is used to show detailed info directly to users in the client.

client/src/components/StatusBadge/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+        onClick={() => setIsOpen(true)}
+        aria-label="Open details modal"
+      >
+        Details
+      </Button>
+      {/* Modal opens when Details is clicked */}
+      {isOpen && (
+        <Modal open={isOpen} onClose={() => setIsOpen(false)}>
+          <div className="p-4">{detailsContent}</div>
+        </Modal>
```
jakewheeler: We should use the other `Modal` components here to make sure the title, body, etc. all look similarly to the app's other modals. There's an example in the modal file if you need a good reference.

refiner/app/db/conditions/db.py
rogeruiz: Without this, I wasn't able to have the `coverage_level*` fields in the API response. Is there another way I should do this?

My understanding is that if the outer `SELECT` is still `SELECT DISTINCT code, system, description`, the coverage fields are dropped before they reach the dataclass / model.

client/e2e/configuration.spec.ts
```diff
@@ -10,6 +10,61 @@ test.describe('Configuration detail flow', () => {
...
+    page,
+    configurationsPage,
+    configurationPage,
+  }) => {
+    const condition = 'COVID-19';
+    await configurationsPage.createConfiguration(condition);
+    await configurationPage.goToBuildTab();
+
+    await expect(
+      page.getByLabel('Code set completion status:', {
```
jakewheeler, line 23: I didn't see much value in searching the page for super specific elements (like `not expanded`) since we are anticipating that these values will change over time (and likely soon). Let me know if you have thoughts/suggestions.

client/src/api/schemas/codeSetStatus.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Generated by orval v8.9.0 🍺
```
fzhao99: no need to handle here, but do we know how to get rid of these comments per our convo in the other PR thread? Technically [we're on `8.9.1` now](https://github.com/CDCgov/dibbs-ecr-refiner/pulls?q=is%3Apr+orval+), so this is "out of date", though I think the correct fix is probably just to get rid of these comments

refiner/app/api/v1/conditions.py
```diff
@@ -53,6 +57,30 @@ async def get_conditions(
...
+    included: bool
+
+
+@dataclass
+class CompletenessStatus:
+    """
+    Condition completeness status model.
+    """
+
+    overall_status: CodeSetStatus
```
fzhao99: think it's a bit more readable to me if the key is also `codeset_status`, to mirror how we're labeling the statuses for the code categories

refiner/app/api/v1/conditions.py
```diff
@@ -61,9 +89,44 @@ class GetConditionResponse:
...
+    groupers: list[DbConditionsContextGrouper],
+) -> list[CodeCategoryCompletenessStatus]:
+    category_names = {
+        "symptom": "Symptom codes",
+        "medication": "Medication codes",
+        "diagnosis": "Diagnosis codes",
+        "clinical_lab_result": "Clinical lab result codes",
+        "immunization": "Immunization codes",
+        "specimen_source": "Specimen source codes",
+    }
```
fzhao99, line 116: mostly for my understanding: is there a possibility of these categories changing, and if so, do we know in which cases? Updates to the TES? Something else?

client/src/api/schemas/codeSetStatus.ts
jakewheeler: We can disable `header` in the Orval config file to do this
```ts
    output: {
      mode: 'tags-split',
      target: './src/api', // react hook output
      schemas: './src/api/schemas',
      client: 'react-query',
      httpClient: 'axios',
      override: {
        header: false, // <--
      },
    },
```

refiner/app/api/v1/conditions.py
jakewheeler: The TES would need to be updated to include new categories. The current setup is a little strange because not all conditions have data for all 6 categories. For example, Cholera only has data for `medication` and `immunization`.

I considered making this more dynamic to handle new the possibility of new categories coming in but decided it's too unreliable/unknown right now to do that. Might be worth thinking about later if we see new categories coming in at some point.

fzhao99 (approved):

PR #1133, opened 2026-04-22
docs: code normalization rfc

### LINKED ISSUE #1110: [SPIKE] Write up RFC for codes normalization
## 🚧 What needs to be done
The team is using the feature requested around rendering child RSG information in the application to change the way we're storing code information. This will support this and other features that require more granular retrival / storage of code information

## 🤔 Why it needs to be done

- Make code storage and retrieval more scalable and easier to maintain
- Unlock forthcoming features that require manipulation of code information across categories
- Simplify and make our overall storage schema more performant

## ⏰ Timebox

3 days

## 📝 To-do list

- Communicate with the team about your plan
- Fix what you can as a part of this work
- Create detailed tech debt tickets for any major tasks that may need more time

## 📜 Additional context

- Came out of an eng discussion of the needed changes for the RFC rendering story

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

RFC writeup  for our code normalization effort

## 🔗 Related Issue

Fixes #1110
- #1110

## 🧪 How to test
Noted a couple places for feedback, so would love input on those (and other) areas of the writeup.

## ℹ️ Additional Information
Did make a slight typing tweak on the seeding script, so would appreciate folks double checking the seeding command didn't break.

### DIFF
--- a/docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
+++ b/docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
@@ -8,11 +8,9 @@ Proposed
 
 ## Context and Problem Statement
 
-Following a Slack discussion about implementation of [1099](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/1099), it was determined that this feature may be a chance to implement a long-desired refactor of the schema we're using to store codesets within the application. Several forthcoming features will need to retreive this data more granularly than the current storage schema easily allows, and the engineering team has acknowledged for a while that this area of the codebase has needed refactoring.
+Following a Slack discussion about implementation of [1099](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/1099), it was determined that this feature may be a chance to implement a long-desired refactor of the schema we're using to store codesets. Several forthcoming features will need to retreive granular versions of this data beyond current storage schema allows. The engineering team has acknowledged for a while that this area of the codebase has needed refactoring and is using this opportunity to explore approaches. The current decision to store codes as JSONB was made when the Refiner was a much different application, where codesets reads from the database scoped to a configuration were the optimized operation, over writes/updates and reads outside the configuration context. With the evolution of the lambda, syncing operations for TES updates, and the development of current and upcoming web app features, this storage decision is in need of revisiting.
 
-The current decision to store codes as JSONB was made when the Refiner was a much different application, where codesets reads from the database were the optimized operation, while writes / updates were deprioritized. With the evolution of the lambda, application update operations following the TES, and the development of current and upcoming web app features, this storage decision is in need of revisiting.
-
-Below are several decision drivers that will inform our decision of 1. Whether / how to refactor our schema to better support codeset information and 2. How to roll out the proposed refactor across the relevant modules of the codebase
+Below are our exploration of 1. Whether / how to refactor our schema to better support codeset information and 2. How to roll out the proposed refactorcodebase.
 
 ## Decision Drivers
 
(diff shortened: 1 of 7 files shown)


justfile
```diff
@@ -126,12 +126,16 @@ _new title type folder:
       const re = /[0-9]{4}/g;
       return fs.lstatSync(fileName).isFile() && fileName.match(re);
     };
+    let resolvedPath = ''
```
fzhao99, line 129: the script to generate an RFC had a few namespace bugs, so I went ahead and fixed them here

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+The core of the stored code information would include the following columns
+
+| column       | datatype                |
+| ------------ | ----------------------- |
+| id           | UUID                    |
+| displayName  | string                  |
+| value        | string                  |
+| system       | `fkey to systems table` |
+| created_at   | DateTime                |
+| last_updated | DateTime                |
```
fzhao99, line 74: I realized when I went to go write the TES seeding script that having foreign keys in the codes table like we discussed earlier in the week would create circular dependencies in the seeding. Assuming that we'd need conditions to hold foreign key relationships into codes, in that world:

- We'd go to insert conditions, which would require code ID's as foreign keys
- We'd go to insert codes, which would need foreign keys into conditions, which haven't been inserted yet. 

Instead, opting to insert codes atomically and maintain those relationships in the parent entity

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Decision Outcome
+
+### Store normalized codesets in a single code table
+
+The team will store normalized codes over extending the existing JSON configuration to better enable future development and maintenance. Furthermore, TES and custom codesets will be stored in a single table, which allows for codeset relationships to be centralized and searched more convieniently. Storing things in a single table simplifies needed features like codeset uniqueness validation and codeset parsing and search, since a single query can be made rather than aggregating queries across tables. We suspect that keeping all codes stored within a single table won't create complexity issues given the scale and volume of data we're maintaining, while offering simplicity and centralization improvements. One table is also more easily splittable into two should we change our minds in the future rather than joining two tables into one, so this decision can be revisted if needed in the future.
+
+### Manage joins via a junction table
```
fzhao99, line 139: opted to use the join table approach in the world where we're not foreign keying out from the codes table. Rationale below, but open to discussion since it's different than what we were discussing in eng sync

refiner/scripts/seeding/lib.py
```diff
@@ -173,14 +178,21 @@ def _sort_codes(self, codes: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
         )
 
+    @property
+    def all_codes(self) -> set[FhirCodeTuple]:
+        """
+        Generates a union of all codes belonging to the condition.
+        """
+        # combine all codes; the union operator `|` correctly merges the sets
+        return self.child_codes | self.sibling_codes
+
```
fzhao99, line 188: did add some minor QoL improvements to our seeding file. don't think this should impact anything, but lmk if something broke for yall locally

refiner/scripts/seeding/load_tes_data_into_normalized_table.py
```diff
@@ -0,0 +1,132 @@
...
+)
+
+from scripts.seeding.load_tes_data import _build_condition_groupers
+
+type UrlVersionTuple = tuple[VsCanonicalUrl, VsVersion]
+type ConditionId = UUID
+type CodeId = UUID
+
+
+# TODO: to be fleshed out in the actual script
```
fzhao99, line 20: added some throwaway code to check our assumptions on how we'd ingest / seed the data from the TES. Main thing here that I'd love feedback on is whether folks can see issues with generating the UUID that we'd insert into SQL in code rather than letting Postgres do so.

Mainly did it this way so it's easier to track / post-process the foreign key relationships between condition <> codes, rather than having to go back and forth to the db between the code insert and condition insert steps to build the final relationship map.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+| created_at   | DateTime                |
+| last_updated | DateTime                |
+
+To accomplish the goal of migration-free code system addition while maintaining maximal data guarentees, the `systems` column of the new codes table will reference a foreign key of a new table that stores system metadata that would look something like the below:
+
+| column | datatype |
+| ------ | -------- |
+| id     | UUID     |
+| name   | string   |
+| oid    | string   |
```
jakewheeler: Do you think there is any value in storing a display name (`ICD-10`) and a "sanitized" name (`icd10`)?

I'm wondering if it'd be valuable to do this at the time of insertion and then not have to think much about it in the app besides pulling it and using it.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+This option stores more code information than strictly necessary, but allows for row-level relationships via foreign keys to drive the configuration and condition relationships between custom and TES-derived codes. In this pattern, the same code could exist in multiple rows, but would be unique within the configuration / condition entity it's stored within.
+
+### 2.2 Code storage schema
+
+The core of the stored code information would include the following columns
+
+| column       | datatype                |
+| ------------ | ----------------------- |
+| id           | UUID                    |
+| displayName  | string                  |
```
jakewheeler: Super small nitpick: I'd want this to be `display_name` in a migration script.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+| value        | string                  |
+| system       | `fkey to systems table` |
```
jakewheeler: This would be `system_id` right?

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+| ------ | -------- |
+| id     | UUID     |
+| name   | string   |
+| oid    | string   |
+
+Future code system addition will be enabled by adding a new row in the systems table that we can populate via a seeding script. To fully align this new table with the backend code, some refactoring will need to be done in the `terminology.py` file to derive backend `CodeSystem` enum values with the values in the new table.
+
+#### 2.2.1 Storing TES codes together with custom codes
+
+The most straightforward option for code storage is to store custom codes and TES codes in a single table, using boolean columns to differentiate the two. The main complication would be the need to link custom code rows to configurations and TES code rows to conditions, since custom codes only exist in a configuration context while TES codes exist to condition codesets. The extended schema under this option would look something like the following:
```
jakewheeler, line 91: This is a great point and I think it actually makes things less straightforward than I was initially thinking.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+| last_updated  | DateTime       |
+| **is_custom** | **Boolean**    |
+
+In return, reads and writes from / to this table would be much more straightforward, since they can be manged by a single database service shared by both the condition and configuration modules. The codes object would provide only the necessary information around custom code / TES status to the caller, which may allow us to simplify the differentiation between the our models of codes.
+
+#### 2.2.2 Storing TES codes separately from custom codes
+
+A secondary option is to store custom codes and TES codes separately, duplicating the base schema in both tables and only adding a foreign key to the relevant configuration or conditions table. The main complication would be the need to link custom code rows to configurations and TES code rows to conditions, since custom codes only exist in a configuration context while TES codes exist to condition codesets. The extended schema under this option would look something like the following:
+
+##### Custom codes table
```
jakewheeler, line 109: I'm leaning more in this direction now. I think your point above (custom codes map to a configuration, TES codes map to a condition) and the fact that we handle custom codes separately from TES codes is enough for me to think this is a better path.

It's also just kind of nice to know that rows in a custom codes table would be editable by users while rows in the TES codes table would not be Just an easy way to think about things in my mind.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -0,0 +1,161 @@
...
+| last_updated | DateTime       |
+
+##### TES codes table
+
+| column       | datatype       |
+| ------------ | -------------- |
+| id           | UUID           |
+| display_name | string         |
+| value        | string         |
+| system       | string or Enum |
```
jakewheeler: Is there a reason this can't be `system_id` and link to a system record? Not sure if I'm misunderstanding your comment about FKs.

fzhao99: nope just an oversight! will fix

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
fzhao99: I think there are def pros and cons to each! Think the main benefit of two table approach creates the cleanest data for sure by reducing the nullable columns by baking the code <> parent relationship into the schema, as well as the editable vs uneditable table point you brought up. Good to know that it's easier for you to grok, since I wasn't sure if extra table = extra complexity here

The main tradeoff in my mind would be the marginal difficulty of code uniqueness validation vs the one table, since (we now have codes split across columns), though it's obviously not impossible to do if we had two tables. Also, going from two to one is slightly more difficult, but don't think that should be anything else other than a tie breaker. 

Ultimately think we should be optimizing for intuitiveness, so would be curious if others also find the two tables setup easier to understand. If it is for others, think it's the right way to go.

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
fzhao99: this is a good idea! will add

docs/decisions/0009_2026-04-17_normalizing-storage-of-codes.md
```diff
@@ -67,35 +67,38 @@ The core of the stored code information would include the following columns
...
-| column | datatype |
-| ------ | -------- |
-| id     | UUID     |
-| name   | string   |
-| oid    | string   |
+| column       | datatype |
+| ------------ | -------- |
+| id           | UUID     |
+| display_name | string   |
+| name         | string   |
```
fzhao99, line 82: @jakewheeler  opted to go with a default `name` and a modified `display_name` here. Lmk if we think this is unclear

jakewheeler: This feels clear enough to me 👍

jakewheeler (approved):
Thanks Bob! 🫡

Issue #1124, opened 2026-04-20
[BUG] Can't edit custom code system or name


kplaneaux:
## 📜 Description

When editing custom codes, only the code can be updated. Trying to change the code system or display name results in a validation error that the code already exists in the code set.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNkdrRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5d7f7f67a579e719f1dc63af9440b0b3e38c166e/Screen%20Recording%202026-04-20%20at%208.56.13%E2%80%AFAM.mov)

## 👣 Steps to reproduce

1. Add a custom code to a new or existing configuration
2. Attempt to edit that code to change code system or display name
3. Observe that you're not able to save the changes

## 🌃 Environment:

- MacOS
- Chrome
- AIMS dev

## ??? Expected Result

Expect to be able to edit code system or display name

## 🧪 Actual result

Not able to

#### Priority

- [ ] **Low priority** _The bug does not have to be fixed immediately. High and medium priority bugs should be addressed first._
- [ ] **Medium priority** _Typically, medium-priority bugs do not affect customers directly and therefore can be fixed in the normal course of testing and development._
- [x] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

Issue #560, opened 2025-10-30
Checking if a custom code already exists should happen on the server


jakewheeler:
# Problem Overview

#482 added the ability to give users immediate feedback when attempting to add a custom code that already exists. While this does work as-is, the current implementation queries for and sends every code from every associated code set to the client for the sole purpose of doing this check. Depending on the number of associated code sets, this could mean 10s of thousands of codes making their way to the client for no purpose other than this check.

This ultimately results in slower performance in the client by a few hundred milliseconds.

In this ticket, we should update this functionality to happen at the server level. This means that the custom code form will need to make a request (either `onBlur` or on `submit - TBD) to show the error to the user. This will allow us to send a `true` or `false` value to the client instead of thousands of codes.

## Acceptance Criteria

- [ ] Codes to perform custom code check are no longer sent to the client
- [ ] Any application code put in place to send codes to the client has been removed from the client and server
- [ ] Custom code modal is updated to send a request to the server asking whether or not this code has already been added
  - [ ] If it has been added, alert the user and don't allow the form to be submitted (if we stick with the same UX - TBD)
  - [ ] If it has not been added, user is allowed to submit the form

## Notes

- It was proposed to change this design to make this check happen when the user clicks the `Add custom code` or `Update` button in the modal. Do we want to go this route instead?

PR #1135, opened 2026-04-22
fix: improvements to custom code validation

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes an issue with the custom code form where the form required a code to change when being edited. This prevented users from leaving the code as-is when editing, even if they only wanted to make an update to the name or system.

This PR contains other improvements related to custom codes as well:

- Validation now happens server-side
  - `deduplicated_codes` configuration property has been removed
-  ~~Fixed a bug where some characters in a custom code (such as `#`) could break custom code deletion functionality~~ There is more that needs to be done here, will fix it in a follow-up PR
- `custom_codes/base.py` is now `custom_codes.py`, tests have been updated where needed
- New fixtures have been added to `integration/conftest.py` to help with modifying a configuration via the API
- Integration tests have been added to exercise the custom code validation logic
- A dedicated custom codes e2e test has been added to check various possible states

## 🔗 Related Issue

Fixes
- #1124 
- #560 

## 📺 Demo

Kapture 2026-04-24 at 11 59 45

## 🧪 How to test

1. Editing a custom code should let you leave the code as-is
2. Attempting to add a custom code that exists already (as a custom code or as part of a code set) should give an error message
3. Update / add button should not be enabled until validation is successful


client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -91,7 +89,8 @@ export function CustomCodesDetail({
                         if (disabled) return;
                         deleteCode(
                           {
-                            code: customCode.code,
+                            // encode to prevent special characters from breaking the action
+                            code: encodeURIComponent(customCode.code),
```
jakewheeler, line 93: This doesn't actually fully address the issue. The "correct" fix would be to use the UUID of the custom code to indicate which record needs to be deleted. However, we can't do this yet)so that'll need to come later.

I'm going to follow this PR up with a temporary fix.

client/e2e/pages/ConfigurationPage.ts
```diff
@@ -76,9 +76,26 @@ export class ConfigurationPage {
     await this.page.getByRole('button', { name: 'Add custom code' }).click();
   }
 
-  async deleteCustomCode(code: string) {
+  async editCustomCode(
```
fzhao99, line 79: 💪

client/e2e/configuration.spec.ts
```diff
@@ -82,9 +204,9 @@ test.describe('Configuration detail flow', () => {
     await test.step('Configure a custom code', async () => {
       await page.getByRole('button', { name: 'Custom codes' }).click();
       await configurationPage.addCustomCode(
-        customCodeName,
+        customCode,
         customCodeSystem,
-        customCode
+        customCodeName
```
fzhao99, line 209: 😅

refiner/tests/unit/test_configurations.py
```diff
@@ -154,7 +154,7 @@ def mock_db_functions(
 
     # for get_total_condition_code_counts_by_configuration_db, could use a list of count objects if needed
```
fzhao99, line 155: is this an example of the unit tests that you'd be ok deleting at some point? looking at them and it looks like the surface area of things we're testing is pretty small

fzhao99 (approved):
appreciate the extra testing improvements we've added! appreciate you cleaning up this bug that I introduced 😬

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
fzhao99: code table normalization ftw! think this is a benefit of including at least a flag in the code table for custom, since it gives us a simple way of querying for all the relevant codes we'd need to check. can imagine that that table would get pretty big over time, so would be beneficial as a column to filter on.

client/e2e/configuration.spec.ts
jakewheeler: Whoops 😆

refiner/tests/unit/test_configurations.py
jakewheeler: Yes, this is a great example of what I was talking about.

1. So much is mocked that we are hardly testing any actual functionality
2. It's hard to tell exactly what needs to be mocked without a lot of trial and error
3. The mocks are difficult / frustrating to maintain (especially anything that uses `monkeypatch` since it's just a string)
4. The "burden to value" ratio from these kinds of tests just don't feel worth it to me

PR #1136, opened 2026-04-22
fix: correct serialization of custom codes into lambda

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Fixing an issue Tim discovered in [this Slack thread](https://skylight-hq.slack.com/archives/C08H7EV6M37/p1776886093130569) around the way we were assigning custom codes to the lambda config files. 

## 🧪 How to test

1. Spin up the application locally
2. In the `localstack/seed.py` file, swap out the assignment on line 88-91 with by uncommenting 94-5. Run the `just server seed-localstack` command.
3. Create a COVID config and add some or all of the following custom codes 
     - Colonic polypectomy (procedure) (274025005) SNOMED
     - Smoking history (11367-0) LOINC 
     - sex assigned at birth (76689-9) SNOMED
     - Systolic (8480-6) LOINC
     - Diastolic (8462-4) SNOMED
1. Activate the config and confirm that the code shows up in the active file URL (should be http://localhost:4566/local-config-bucket/configurations/SDDH/07221093-b8a1-4b1d-8678-259277bfba64/1/active.json) 
1. Fire off the lambda and navigate to the output URL (should be http://localhost:4566/local-config-bucket/RefinerOutput/persistence/id/SDDH/COVID19/refined_eICR.xml)
2. Observe that those codes are retained


refiner/app/services/configurations.py
```diff
@@ -190,7 +191,8 @@ async def convert_config_to_storage_payload(
 
         for code_system, code_list in code_system_map.items():
             codes = codes | {c.code for c in code_list}
-            coding_by_code_system[CodeSystem(code_system).format_system_string()] = [
+            system_to_extend = CodeSystem(code_system).format_system_string()
+            coding_by_code_system[system_to_extend] += [
```
fzhao99: main issue was that we were replacing and not appending to the list in question

refiner/app/services/configurations.py
```diff
@@ -164,15 +164,16 @@ async def convert_config_to_storage_payload(
     # custom codes
     for cc in configuration.custom_codes:
         codes.add(cc.code)
-        cur_code_system = cc.system
+        cur_code_system = CodeSystem(cc.system)
+        system_to_extend = cur_code_system.format_system_string()
```
fzhao99, line 168: there was a formatting issue (capital vs lower case) with the way we were indexing the dictionary in question that we were serializing into the lambda activation file. Using the utility method off the object gives us consistent dictionary key values

rogeruiz (approved):
lgtm 🌈

PR #1137, opened 2026-04-23
fix: inline and independent test configurations use the same serialization/deserialization process as Lambda

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the inline and independent test flows so that the `ProcessedConfiguration` used for refinement is created by serializing to a `ConfigurationStoragePayload` and then deserializing this object.

This brings the web app's refinement process much closer to the way that refinement works in Lambda. The only main difference now is that the `ConfigurationStoragePayload` is used immediately rather than being written to a file.

## 🧪 How to test

- Both inline and independent testing should function as they did before
- All automated tests should continue to pass


fzhao99 (approved):
Thanks so much!

PR #1144, opened 2026-04-27
refactor: Remove `ProcessedConfiguration.from_payload()`

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR removes the option to create a `ProcessedConfiguration` using the `from_payload()` method. 

Last week we noticed the refined results of Lambda differed from those of the web app due to a divergence in how `ProcessedConfiguration` objects were being created in each. We merged two PRs to address this:
- #1136 (bug fix for creation)
- #1137 (ensure web app and Lambda create `ProcessedConfiguration` the same way)

This PR goes a step further so that there is now only a single way to create a `ProcessedConfiguration`, which is by using the `from_dict()` method. 

## 🧪 How to test

- App and tests should work as expected

We stopped using `from_payload()` in #1137. This PR only deletes the option so there should be no changes to the app itself.

### DIFF
--- a/refiner/tests/unit/test_service_refine.py
+++ b/refiner/tests/unit/test_service_refine.py
@@ -1,3 +1,6 @@
+import unittest
+
+import pytest
 from lxml import etree
 
 from app.db.conditions.model import DbCondition, DbConditionCoding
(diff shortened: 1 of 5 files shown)


refiner/tests/unit/test_service_refine.py
```diff
@@ -24,6 +26,9 @@
 
 _PLACEHOLDER_AUGMENTATION_TIMESTAMP = "19700101000000+0000"
 
+# ProcessedConfiguration must contain at least one code or otherwise is considered
+# to be invalid. We are using a code that definitely will not appear in a real code set.
+_MOCK_CONDITION_CODES = [DbConditionCoding(code="fake-code-0!", display="")]
```
jakewheeler, line 31: Does this make sense? This is a clearly fake code to meet the validation requirement of `ProcessedConfigurationData` (`codes` must have at least 1 item).

I questioned if `ProcessedConfigurationData`'s `Field(min_length=1)` requirement on `codes` made sense but I think it does so I left it be. `codes` being blank in an active configuration file seems like it'd always indicate an issue. 

Let me know if you disagree.

fzhao99: agreed, think there should always be at least one code in the activation file. don't think we'd ever be in a situation in production with an empty codes list, so this feels like a good workaround

fzhao99 (approved):

PR #1157, opened 2026-04-27
feat: view individual custom code events associated with a CSV bulk upload

### LINKED ISSUE #940: [TASK] Break out custom code upload activity log entries
## 🚧 What needs to be done

Activity log entries for custom code uploads need to reflect the individual codes added.

## 🤔 Why it needs to be done

It's important to have traceability of codes added to a configuration so there's record for the PHAs of what's changed version to version.

## 🏗️ Acceptance Criteria

- [ ] Custom code upload activity log entries must be code-by-code

## ⚙️ Technical Notes

- ...

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a new feature where custom code CSV upload will produce events for the individual custom codes that were uploaded. These individual events are tied to a parent `bulk_add_custom_code`.

When a bulk upload event is in the activity log table, there is now a `View all` button that opens a modal displaying all custom codes that were included in that CSV upload.

## 🔗 Related Issue

Fixes #940 

## 📺  Demo

Kapture 2026-04-28 at 15 19 02

## 🧪 How to test

1. Run `just migrate local` (feel free to also try `just migrate local down` if you'd like)
2. Use the custom code bulk upload feature (this is my demo file if you want to try it -> [custom_code_upload_template.csv](https://github.com/user-attachments/files/27178590/custom_code_upload_template.csv))
3. Check the activity log page. You should see there is a `View all` button available
4. Click the button and you'll see the entries load into the modal


refiner/app/db/events/db.py
```diff
@@ -71,7 +108,10 @@ async def get_events_by_jd_db(
         c.version as configuration_version,
         c.id AS condition_id,
         e.action_text,
-        e.created_at
+        e.created_at,
+        EXISTS (
+            SELECT 1 FROM events_custom_code_uploads ecu WHERE ecu.event_id = e.id
+        ) AS has_custom_code_upload_events
```
jakewheeler, line 114: I've checked that this appropriately handles bulk upload events prior to this change. In that case, this will be `false` and no `View all` will display in the UI.

Here's an example (the event text was previously `Added 23 custom codes`):
Image

refiner/app/db/events/db.py
```diff
@@ -89,10 +129,69 @@ async def get_events_by_jd_db(
...
+            await cur.execute(query, params)
+            rows = await cur.fetchall()
+            return rows
+
+
+async def insert_custom_code_upload_events_db(
+    configuration: DbConfiguration,
+    user_id: UUID,
+    custom_codes: list[DbConfigurationCustomCode],
+    cursor: AsyncCursor[Any],
```
jakewheeler, line 161: Bulk event + individual code events are written in the same transaction.

client/e2e/activity-log.spec.ts
```diff
@@ -45,4 +45,57 @@ test.describe('Activity log', () => {
...
+    activityLogPage,
+  }) => {
+    const condition = 'Lead in Blood';
+    const config = await api.createConfiguration(condition);
+
+    // Create 50 codes to upload
+    const customCodes = Array.from({ length: 50 }, (_, i) => ({
+      code: `mc-${i + 1}`,
+      name: `mock code ${i + 1}`,
+      system: 'SNOMED',
```
fzhao99: any chance we could vary the list of possible systems here, maybe using an index hash from the possible set of systems? might make for a slightly more robust test to any future code system issues

client/src/pages/ActivityLog/ActivityLogEntries.tsx
```diff
@@ -64,3 +79,84 @@ export function ActivityLogEntries({
...
+                        <th scope="col">Display name</th>
+                      </tr>
+                    </thead>
+                    <tbody className="divide-gray-cool-20 divide-y">
+                      {events.data.map((cc) => (
+                        <tr key={cc.id}>
+                          <td className="py-3">{cc.system}</td>
+                          <td>{cc.code}</td>
+                          <td>{cc.name}</td>
+                        </tr>
```
fzhao99:
```suggestion
                        <th scope="col">
                          Code system
                        </th>
                        <th scope="col">Code</th>
                        <th scope="col" className="py-3">Display name</th>
                      </tr>
                    </thead>
                    <tbody className="divide-gray-cool-20 divide-y">
                      {events.data.map((cc) => (
                        <tr key={cc.id}>
                          <td >{cc.system}</td>
                          <td>{cc.code}</td>
                          <td className="py-3">{cc.name}</td>
                        </tr>
```

in cases where there are long display names, think the modal table looks slightly better if we put the padding on the name column (which would have the longest string in the cases where there's wrapping)

Before:
Image

After:
Image

fzhao99 (commented):
two minor suggestions but otherwise lgtm! thanks :)

client/e2e/activity-log.spec.ts
jakewheeler: I've updated it to cycle through a list of different systems. Let me know if there's still anything additional you're looking for here.

client/src/pages/ActivityLog/ActivityLogEntries.tsx
jakewheeler: Good idea. Thanks!

fzhao99 (approved):

PR #1160, opened 2026-04-28
chore: integrate new systems table 

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Refactor that builds off the seeding of the systems table to integrate data into the rest of the application code

## 🔗 Related Issue

Fixes #1285
- #1285

## ✅ Acceptance Criteria
- [ ] Code system information is driven from the database table as much as possible. 
- [ ] Existing functionality is maintained

## 🧪 How to test

Test the flows reliant on code system addition (custom code single add / bulk upload, code set filtering on code system


client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
```diff
@@ -193,6 +187,24 @@ function CustomCodeForm({
     }
   };
 
+  if (isPending)
```
fzhao99, line 189: wondering if it may be good to factor this pending / error UI into a component? We seem to be using it in a bunch of places

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
```diff
@@ -145,10 +139,10 @@ function CustomCodeForm({
           configurationId,
           data: {
             code: selectedCustomCode.code,
-            system: normalizeSystem(selectedCustomCode.system),
+            system: selectedCustomCode.system,
```
fzhao99: opted to do away with the normalization here since these system values are coming from the server now, which has a layer of sanitation / coersion in the new sanitization functions to be more flexible. Let me know if folks forsee issues here though.

refiner/scripts/seeding/load_system_data.py
```diff
@@ -0,0 +1,66 @@
...
+    get_db_connection,
+)
+
+CODE_SYSTEM_DATA = {
+    "2.16.840.1.113883.6.96": {"name": "snomed", "display_name": "SNOMED"},
+    "2.16.840.1.113883.6.1": {"name": "loinc", "display_name": "LOINC"},
+    "2.16.840.1.113883.6.90": {"name": "icd10", "display_name": "ICD-10"},
+    "2.16.840.1.113883.6.88": {"name": "rxnorm", "display_name": "RxNorm"},
+    "2.16.840.1.113883.12.292": {"name": "cvx", "display_name": "CVX"},
+    "2.16.840.1.113883.5.1008": {"name": "other", "display_name": "Other"},
```
fzhao99: I opted to use [the OID for NullFlavor OTH](https://terminology.hl7.org/1.0.0/CodeSystem-v3-NullFlavor.html) here. lmk if this would cause issues down the line though

refiner/app/services/terminology.py
```diff
@@ -18,108 +24,129 @@
...
+    ) -> DbCodeSystem | None:
         """
-        Get the corresponding code system OID.
+        Get a specific code system based on its name.
         """
+        value = [system for system in cls._registry.values() if system.oid == oid]
+        if len(value) == 0:
+            return None
+        if len(value) > 1:
+            logger.warning(
```
fzhao99: opted to log a warning rather than raising here and in the display name case since we're controlling the reference set. Lmk if folks think raising is preferable though

refiner/app/main.py
```diff
@@ -76,6 +78,10 @@ async def _lifespan(app: FastAPI):
         app.state.db = db
         await db.connect()
         logger.info("Database pool opened", extra={"db_pool_stats": db.get_stats()})
+
+        # load app-wide registry info from the DB
+        await SupportedCodeSystems.load_from_db(db)
```
fzhao99: Opted to load the systems table into application memory once on startup but think this may be bad? Would love feedback as to whether this is inadvisable / there's a better place for this load. Could also envision some sort of cache invalidation / re-read being desirable.

refiner/scripts/seeding/load_system_data.py
```diff
@@ -0,0 +1,66 @@
+from config import logger
+from lib import (
+    get_db_connection,
+)
+
+CODE_SYSTEM_DATA = {
+    "2.16.840.1.113883.6.96": {"name": "snomed", "display_name": "SNOMED"},
+    "2.16.840.1.113883.6.1": {"name": "loinc", "display_name": "LOINC"},
+    "2.16.840.1.113883.6.90": {"name": "icd-10", "display_name": "ICD-10"},
```
fzhao99: opted here just to use the lowercase `icd-10` with the hyphen included since the misalignment was causing a bunch of issues. Did add [a check for the old way](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160/changes#diff-21908ed1b118de5ade618e8e597ba5743f693497b2d6dd19423c1ab264dd8457R55-R57) in the sanitization code though to make this robust as possible.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
rogeruiz: I think it's a good idea since we're using it in a few places. But I do think it could be part of another ticket rather than making it part of this larger refactor. Open to suggestions from others though if folks believe it should be done in this PR.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
rogeruiz: 🚀 Yeah, this is good since we don't have to have duplicated logic around this. thanks for clearing this up.

refiner/scripts/seeding/load_system_data.py
rogeruiz: I think this is great idea for the `Other` OID since it seems like it's all about representing the lack of a valid value and has lots of codes to further specify that. Should help downstream integrations too as long as the consuming systems are aware of us using that OID.

refiner/app/services/terminology.py
rogeruiz: I think this is okay 🤷🏼.I think this solution is reasonable for now. I can't foresee any reason why we'd need to raise here.

refiner/app/main.py
rogeruiz: Yeah you hit all the points I'd have for this. I think it's solid for now as our data volume is small and shouldn't change often. I think we may run into issues if we horizontally scale the application and for whatever reason the long-running container and the newer containers are operating on different data from the database. So having some kind of cache invalidation makes sense but that's not an issue I foresee right now. Also, probably documenting this somewhere would be good in case we find it is a problem in the future.

refiner/scripts/seeding/load_system_data.py
rogeruiz: Oh this is great! Thanks @fzhao99. I love the idea of having the fallback here rather than having to try and validate things in the client code.

client/src/pages/Configurations/ConfigBuild/CsvCodeUploadModal.tsx
```diff
@@ -105,13 +106,27 @@ export function PreviewEditModal({
...
+  } = useGetCodeSystems();
+
+  if (isPending)
+    return (
+      <div className="flex w-full justify-center">
+        <Spinner />
+      </div>
+    );
+
+  if (isError || !supportedCodeSystems) return 'Error!';
```
rogeruiz: I wonder if we should revisit this messaging UX-wise in the future. Though I think the string return is appropriate for now.

refiner/migrations/20260428155732_add_system_table.sql
```diff
@@ -0,0 +1,11 @@
+-- migrate:up
+CREATE TABLE systems (
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+    name text NOT NULL,
+    display_name text NOT NULL,
+    oid text NOT NULL UNIQUE
```
rogeruiz: I wonder if we should add a constraint to validate it's HL7 OID style? Perhaps something like this to match the documentation here: https://build.fhir.org/ig/HL7/CDA-core-sd/StructureDefinition-oid.html


```suggestion
    oid text NOT NULL UNIQUE
    CONSTRAINT oid_format_valid CHECK (
      oid ~ '^[0-2](\.(0|[1-9][0-9]*))+$'
    )
```

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -146,15 +117,14 @@ async def add_custom_code(
         )
 
     # Create a custom code object
-
-    if sanitized_system_name not in ALLOWED_CUSTOM_CODE_SYSTEMS:
+    if not selected_code_system:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
-            detail=f"System must be one of [{ALLOWED_CUSTOM_CODE_SYSTEM_NAMES}]. Got: {sanitized_system_name.value}",
+            detail=f"System must be one of [{SupportedCodeSystems.allowed()}]. Got: {selected_code_system.name}",
```
rogeruiz: I like this approach. But I've been thinking if we should have a canonical list of error codes to make debugging easier that we can attach to things like this. Some internal list that would make debugging things from the client application easier. We could probably use them in the refiner/lambda logs too for easier discoverability when combing through logs. But this is out of scope for this PR. Figured I'd add it and maybe we can make it a new ticket?

rogeruiz (commented):
Looking good so far @fzhao99. I left some feedback on some things. I do think the constraint on the `oid` column would be a good idea but I'm not sure if we want more flexibility here. Looking forward to reviewing this again. 👍🏼

refiner/migrations/20260428155732_add_system_table.sql
fzhao99: I like the thought! Main thing is that I think something like this might be better as a validator function in the ingest interface / within Python so we could have proper error handling fallbacks, rather than at the data level where it may be hard to debug. 

I'll look into where it might make sense to add it

refiner/app/api/v1/configurations/custom_codes.py
fzhao99: Yeah we have [a file for this](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/core/exceptions.py) that we haven't been very disciplined in using. Think it'd be worthwhile for us to do a larger refactor for us to define all those errors centrally / have a unified place for us to define and pick up error logic.

A new ticket would be welcome!

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
jakewheeler: I wouldn't do it in this PR. It might be worth asking design if they have something in mind for a default error component. This was a placeholder I figured would go away at some point 😅

refiner/app/main.py
jakewheeler: One thing that's worth noting here is that we can't currently guarantee these things by the time the app goes to start up:
1. The `systems` table exists
2. The `systems` table has the latest data

We will very likely run into both of these things since the app and the ops container both go out and attempt to start up at the same time.

Aside from that, fetching the systems from the DB is not very expensive. I think it'd be simpler to fetch them as needed.

refiner/scripts/seeding/load_system_data.py
jakewheeler: I have a couple of thoughts on this:
1. Should `Other` even be a row in the `systems` table? What if an `Other` selection in the UI meant "no code system associated with the custom code"?
2. If we do store it, any reason not to give it an `OID` of `NULL` instead? This seems more clear at a glance to me.

refiner/scripts/seeding/load_system_data.py
```diff
@@ -0,0 +1,66 @@
+from config import logger
```
jakewheeler: Should the contents of this file be a part of `load_tes_data.py`?

refiner/app/db/code_systems/db.py
```diff
@@ -0,0 +1,51 @@
...
+    A code system row from the `systems` table.
+    """
+
+    id: UUID
+    name: str
+    display_name: str
+    oid: str
+
+
+type CodeSystemName = str
```
jakewheeler: What if we made this a `Literal` and used it when we needed to reference a system? If we give limited and specific options this would eliminate a lot of the logic around determining whether a code system is "supported" or not.

```python
type CodeSystemName = Literal["ICD-10", "SNOMED", ...]
```

refiner/app/api/v1/codes/model.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
```
jakewheeler: I have a few general suggestions here:
1. Delete the `codes` directory, move `code_systems.py` into `/v1`
2. Delete this file and move `GetSupportedCodeSystemsReponse` directly above the route in `code_systems.py`
3. Rename the model to `CodeSystemsResponse`

Let me know if you disagree with any of this. I feel this follows the pattern of most of the other routes a bit more closely.

refiner/app/api/v1/codes/code_systems.py
```diff
@@ -0,0 +1,30 @@
+from fastapi import APIRouter
+
+from app.api.v1.codes.model import GetSupportedCodeSystemsReponse
+from app.services.terminology import SupportedCodeSystems
+
+router = APIRouter()
```
jakewheeler: This router needs a prefix. I noticed the data was coming back from `/v1`. One of these would work.

```suggestion
router = APIRouter(prefix="/systems")
```
```suggestion
router = APIRouter(prefix="/code-systems")
```

refiner/app/db/code_systems/db.py
fzhao99: Could you tell me more about how this would differ from our old approach of specifying the list of codes as an enum? Was under the impression that the point of storing the code system names was to get rid of the in-application specification of code information and move that into the seeding step, in preparation of eventually getting that from an API call in much the same way we do for TES data. Specifying them in application code seems to get away from the spirit of that and would almost lead me to just keep the old enum approach, since it allows us also to do more robust class-based operations.

Know this is a bit of an architectural-level response to a tactical-level suggestion, so lmk if there's something I'm missing / a misaligned assumption at the architectural level that's behind this suggestion

refiner/scripts/seeding/load_system_data.py
fzhao99: The main reason I chose this OID rather than null is mainly to follow CDA patterns for [providing NullFlavor when something isn't specified](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/DIBBs-eCR-and-CDA-for-Engineers.md#15-nullflavor-how-cda-says-i-dont-know). Since these OID's are only really being used in those CDA contexts, following the grammar around NullFlavor / providing that OID seemed appropriate. @robertmitchellv feel free to weigh in on this if we're missing something.

> Should Other even be a row in the systems table? What if an Other selection in the UI meant "no code system associated with the custom code"?

I ultimately wanted to add this Other row since it'd make validation in the context of things like CSV upload easier, since otherwise, we'd need to layer in this "Other" option in that case as well as when displaying things in the frontend. With those two use cases in mind, "Other" is a legitimate catchall option rather than a second-class systems citizen, so I think storing it would be easier rather than having to add it in whenever we send / receive this information from the frontend.

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -350,3 +337,48 @@ function ConditionCodeRow({
...
+
+function CodeSystemSelection({
+  selectedCodeSystem,
+  handleCodeSystemSelect,
+}: CodeSystemSelectionProps) {
+  const {
+    data: supportedCodeSystems,
+    isPending,
+    isError,
+  } = useGetCodeSystems();
```
jakewheeler: I might suggest attaching system info to the response of `useGetCondition()` (`GET api/v1/conditions/:id`) instead. Not a huge deal but eliminates an additional network request each time the client requests a condition code set.

It also eliminates the need to handle an additional loading/error state since this is already being handled for the condition code set.

refiner/scripts/seeding/load_system_data.py
fzhao99: I guess this is a pattern thing we can decide as a team: whether we want to keep all the "shove the needed static asset scripts into the db" steps in one file, or if we'd rather keep this and any similar future scripts semantically separate and then orchestrate them in the load step. 

I guess I have a weak preference to keeping them separate just for extensibility reasons (in a world where we have other sources of static data to seed, for instance), but not too tied to it if you feel strongly keeping them in one file is better.

refiner/app/main.py
fzhao99: Ahh this is a good point. I'll look into just doing an on-demand fetch in that case

refiner/scripts/seeding/load_system_data.py
jakewheeler: I think as long as we have a single script to handle all of the data setup tasks I'm happy. I suppose the actual content of this file doesn't need to be in `load_tes_data.py` but I think it'd make sense for that script to call the function to handle system loading. This might not be the main setup script forever but it is for now at least 🤷

refiner/app/db/code_systems/db.py
jakewheeler: We might have slightly different outcomes in mind for this work. I don't know that I'm necessarily looking for a fully dynamic, hands-off approach to adding system info.

The main things I'm interested in are:
- Database being the source of truth for system info (in whatever format we like)
- All system info used by the client comes from the server
- Custom codes store their related system by ID

Actually, as I typed that last bullet out, I'm thinking we should be using IDs rather than name which actually would make it fully dynamic in the way you're looking for since loading and using systems would then be very similar to how we work with conditions coming from the TES.

I think I was stuck on using the code system name initially since that's what I was looking at here, but ID seems like the way to go instead.

Does that make? We offer the client a list of system ID options, client makes a selection, server checks if that's a valid system ID and if so, does something with it. If not, return an error to the client.

refiner/scripts/seeding/load_system_data.py
jakewheeler: Does following CDA matter in this instance if the system isn't "real" and won't make it into any output from refiner? If I looked at the database table I would understand a `NULL` oid but wouldn't be able to recognize that `2.16.840.1.113883.5.1008` is a NullFlavor unless I looked it up.

👆 Let me know if I'm misunderstanding something here. It seems like the system data belongs to us and we can handle it in whatever way we want to since it's only used internally.

I'm not necessarily against `Other` being in the table as long as it's clear this isn't a "real" system (which I think a `NULL` oid could fairly obviously represent). I was thinking if another table's row had a `NULL` `system_id` reference that could implicitly mean "Other" rather than have an explicit catch-all value.

refiner/app/db/code_systems/db.py
fzhao99: yeah this makes sense. 

The one difficulty of using ID's over name (which I think I'm going to rename `key` to differentiate it from the `display_name` field, ie `snomed` used in our app code over `SNOMED` displayed to users) is that the serialization into S3 is using the system key. A little worried over the scope of also changing the S3 serialization in the same PR, so I'll just rename them for now and leave an ID refactor as a follow up.

refiner/scripts/seeding/load_system_data.py
fzhao99: ok, reviewing some of the way we're storing the code sets convinced me that null shouldn't be an issue. Will make this change

refiner/scripts/seeding/load_system_data.py
fzhao99: Right, I assumed `seed_db.py` was our catch-all script for data sightup, which is why I[ added the call for `load_system_data` there](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160/changes#diff-eded58150675a583f232a0cef4f6cc76a6ed40199a61b999db62b481972c1936R45) right after the invocation of the tes-related seeding. It looks like that orchestration function is getting called by docker whenever we run seed.

Would we want this function to be called in `load_tes_data.py` instead of `seed_db.py`?

refiner/app/db/code_systems/db.py
```diff
@@ -49,3 +50,198 @@ async def get_all_code_systems_db(
...
+            db=db, logger=logger, name=string_to_search
+        )
+        return by_name
+    except ValueError:
+        # fall back to search by key
+        by_key = await get_code_system_by_key_or_raise_db(db=db, key=string_to_search)
+        return by_key
+
+
+async def get_allowed_code_system_display_names(
```
fzhao99: opted to put these methods in the db file rather than in service since they're just thin wrappers around the other functions. happy to move them if we feel that conforms better with the overall repo pattern though

refiner/scripts/seeding/load_system_data.py
fzhao99: resolved in #1246

refiner/scripts/seeding/load_static_data.py
```diff
@@ -233,7 +233,7 @@ def _build_condition_groupers(
     "icd10": {"oid": "2.16.840.1.113883.6.90", "display_name": "ICD-10"},
     "rxnorm": {"oid": "2.16.840.1.113883.6.88", "display_name": "RxNorm"},
     "cvx": {"oid": "2.16.840.1.113883.12.292", "display_name": "CVX"},
-    "other": {"oid": None, "display_name": "Other"},
+    "other": {"oid": "Other", "display_name": "Other"},
```
fzhao99, line 236: opted to change this to this string since it more closely matches to how we were serializing things before

refiner/app/services/ecr/specification/constants.py
```diff
@@ -39,3 +40,12 @@
     ICD10_OID: "ICD-10",
     CVX_OID: "CVX",
 }
+
+OID_TO_SYSTEM_KEY_MAP: Final[dict[str, str]] = {
```
fzhao99, line 44: this is the one place where I still need to hard code the mapping between code system and OID, so as to maintain the relationship between the refinement passes and the DB-stored data. Right now, the encoding between the codes we store in S3 and the codes used by the refinement runs are linked via the introduced `code_system_key` (the old system name) via OID. This is the last piece we'll need to resolve if we want to make system-related actions fully dynamic as discussed [here](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160#discussion_r3260064703)

The code system concepts in refinement are quite different from the way we're using it within the app in that they have much more sematic meaning within the section / refinement rule we're making, so maintaining that mapping manually, at least for now, feels more appropriate to me as we're building this out, rather than somehow doing a DB read somewhere in the chain to get the system info.

lmk what folks think

refiner/scripts/seeding/load_static_data.py
jakewheeler: This value doesn't seem to update for me when I re-run the script.

refiner/app/db/configurations/db.py
```diff
@@ -308,7 +308,7 @@ async def insert_configuration_db(
                     # custom_codes
                     Jsonb(
                         [
-                            {"name": c.name, "code": c.code, "system": c.system}
+                            {"name": c.name, "code": c.code, "system_key": c.system_key}
```
jakewheeler, line 313: Is this correct? Previously added custom codes are still going to be using `system`.

[This](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160/changes#diff-0d203e9cf5d7d386ad264627cf69503e10132a0e77eabecdfa8387340035f1f9R167) makes me think it's still intended to be `system`?

refiner/app/db/code_systems/db.py
jakewheeler: I'd suggest moving them to `services/code_system.py`. I think they'd fit better there.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -270,15 +253,21 @@ async def upload_custom_codes_csv(
...
             row_errors.append("Missing display_name")
         sanitized_system = None
+        allowed_systems_str = ", ".join(
+            await get_allowed_code_system_display_names(db=db)
+        )
         try:
-            sanitized_system = CodeSystem.sanitize(code_system_raw)
+            sanitized_system = await get_code_system_by_key_or_display_name_or_raise_db(
+                name=code_system_raw, db=db
+            )
```
jakewheeler: I'd recommend that we avoid calling these within the loop since this could pretty easily end up being a lot of DB hits if the list of rows is decently large. We could assemble the data to check against before the loop so it's in memory when we're doing the checks.

refiner/app/api/v1/configurations/model.py
```diff
@@ -222,7 +208,7 @@ class UploadCustomCodesPreviewItem(BaseModel):
     """Validated CSV row ready for confirmation."""
 
     code: str
-    system: CodeSystem
+    system_key: str
```
jakewheeler, line 218: If this does need to change we'll need to write a migration for it.

refiner/app/db/code_systems/db.py
jakewheeler: They are so simple I also wouldn't be opposed to deleting the functions and letting the callers add the additional one line 🤷

refiner/app/services/configurations.py
```diff
@@ -165,8 +171,10 @@ async def convert_config_to_storage_payload(
     # custom codes
     for cc in configuration.custom_codes:
         codes.add(cc.code)
-        cur_code_system = CodeSystem(cc.system)
-        system_to_extend = cur_code_system.format_system_string()
+        cur_code_system = await get_code_system_by_key_or_raise_db(
+            key=cc.system_key, db=db
```
jakewheeler: The `or_raise_db` is a bit unconventional. Any reason not to do the usual `None` check?

refiner/app/api/v1/configurations/base.py
```diff
@@ -361,4 +371,5 @@ async def get_configuration(
         latest_version=latest_version,
         locked_by=locked_by,
         is_locked=is_locked,
+        code_systems=code_systems,
```
jakewheeler: It might be better to group these under `custom_codes`? Configurations don't have any direct link to code systems so this feels a bit off to me that these are at the top level of a config response.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
```diff
@@ -193,6 +195,24 @@ function CustomCodeForm({
...
+    );
+
+  if (isError || !codeSystems) return 'Error!';
+
+  const systemValues: CodeSystemsReponse[] = [
+    {
+      display_name: 'Select system',
+      key: SELECT_DEFAULT_LITERAL,
+      oid: '',
+      id: 'c107a769-4de6-4b3f-bdf0-261284259cfd',
```
jakewheeler: I am not sure I understand this. Why is this configured like it's a choice the user can select? I am confused where the ID is coming from.

fzhao99: oops think I had this as a placeholder and forgot to change it back. fixed it to a better default select pattern.

refiner/scripts/seeding/load_static_data.py
fzhao99: good catch! added an exclude update to the seed query

refiner/app/db/configurations/db.py
fzhao99: ahh good point. I wanted to keep things consistent to how we were using it in the system service code (where everything is using `system_key`), but don't think it's worth the hassle of migrating the data. 

Hopefully the use of `system` in custom codes contexts is clear enough to disambiguate it's expecting the value for `key` and not the object / display name

refiner/app/api/v1/configurations/model.py
fzhao99: assuming this is referring the change from `system` --> `system_key` that we were discussing in the other thread, which I've reverted.

Lmk if we need a migration for some reason to change to the typing.

refiner/app/api/v1/configurations/custom_codes.py
fzhao99: did a temporary change to get the data outside the loop, but think this function needed a larger refactor since I found it kind of hard to follow.

got that started [here](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1310).

refiner/app/db/configurations/db.py
jakewheeler: Should we not migrate `system` values in existing custom codes to use the `key` value instead and then use `key` going forward (for now)?

I think this is what I'm wanting to avoid. First code was added on `main` and second was on this branch:
```json
[
  {
    "code": "old_code",
    "name": "old_code",
    "system": "ICD-10"
  },
  {
    "code": "new_code",
    "name": "new_code",
    "system": "icd10"
  }
]
```

This is not an impossible situation to handle in a future migration but I don't see the benefit in holding off on fixing this. I'd feel better about the data being consistent, especially since we don't have the ability to see DB data in live environments.

This would make a future migration to a custom codes table easier if we can assume `key` is correct and maps to a row in `systems`.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -482,12 +474,14 @@ async def delete_custom_code(
...
         config_id=config.id, db=db
     )
+    code_systems = await get_all_code_systems_by_key(db=db)
 
     return ConfigurationCustomCodeResponse(
         id=updated_config.id,
         display_name=updated_config.name,
         code_sets=config_condition_info,
         custom_codes=updated_config.custom_codes,
+        code_systems=code_systems,
```
jakewheeler: I'm thinking we don't need to return this?

No need to change anything else but I'm not sure why we return a lot of the data that we do here.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -504,21 +498,25 @@ class UpdateCustomCodeInput(BaseModel):
...
+    sanitized_system = await get_code_system_by_key_db(key=updateInput.system, db=db)
+    if sanitized_system is None:
+        raise ValueError(
+            f"System of name {updateInput.system} doesn't match supported systems"
+        )
     code_to_edit = [
         cc
         for cc in custom_codes
-        if cc.system == sanitized_system.value
+        if cc.system == sanitized_system.key
```
jakewheeler: I'm noticing that editing doesn't work for custom codes using data added prior to this PR. I think due to [this](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160#discussion_r3342283350).

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -751,7 +750,7 @@ async def edit_custom_code(
         prev_system=body.system,
         prev_name=body.name,
         new_code=body.new_code,
-        new_system=body.new_system,
+        new_system=body.system,
```
jakewheeler: Accidental change I think?

refiner/app/api/v1/configurations/model.py
```diff
@@ -103,6 +114,7 @@ class ConfigurationCustomCodeResponse:
     display_name: str
     code_sets: list[DbTotalConditionCodeCount]
     custom_codes: list[DbConfigurationCustomCode]
+    code_systems: dict[CodeSystemKey, DbCodeSystem]
```
jakewheeler: This is no longer needed, right?

refiner/app/api/v1/code_systems.py
```diff
@@ -18,7 +18,7 @@ class CodeSystemsReponse:
     id: UUID
     key: str
     display_name: str
-    oid: str
+    oid: str | None
```
jakewheeler: Is this still the expectation if `other` is now the OID for the `Other` system?

refiner/app/api/v1/conditions.py
```diff
@@ -176,4 +178,5 @@ async def get_condition(
             code_category_statuses=code_category_statuses,
         ),
         codes=condition_codes,
+        systems=list(systems.values()),
```
jakewheeler: This should probably be `CodeSystemsResponse`?

refiner/app/api/v1/code_systems.py
jakewheeler: Actually on second thought, it might make the most sense to not include `oid` at all if we don't use it in the client.

refiner/app/db/configurations/db.py
fzhao99: hmm yeah this is bad. I'll add a migration

refiner/app/db/code_systems/db.py
```diff
@@ -39,13 +58,155 @@ async def get_all_code_systems_db(
     """
 
     async with db.get_connection() as conn:
-        async with conn.cursor(row_factory=class_row(DbCodeSystem)) as cur:
```
jakewheeler, line 42: Any reason not to let psycopg handle the mapping? Something like this
```python
        async with conn.cursor(row_factory=class_row(DbCodeSystem)) as cur:
            await cur.execute(query)
            rows = await cur.fetchall()

    return {system.id: system for system in rows}
```

fzhao99: oh I think this was a holdover from when I was typing DbCodeSystem differently. I'll make the switch

refiner/migrations/20260602161536_replace_custom_code_system_key.sql
```diff
@@ -0,0 +1,44 @@
...
+        || CASE
+                WHEN elem->>'system' = 'ICD-10' THEN '{"system_key": "icd10"}'::jsonb
+                WHEN elem->>'system' = 'SNOMED' THEN '{"system_key": "snomed"}'::jsonb
+                WHEN elem->>'system' = 'LOINC' THEN '{"system_key": "loinc"}'::jsonb
+                WHEN elem->>'system' = 'RxNorm' THEN '{"system_key": "rxnorm"}'::jsonb
+                WHEN elem->>'system' = 'CVX' THEN '{"system_key": "cvx"}'::jsonb
+                WHEN elem->>'system' = 'Other' THEN '{"system_key": "other"}'::jsonb
+                ELSE '{}'::jsonb
+            END) - 'system'
+            ORDER BY ord
```
fzhao99, line 18: I don't ordinality strictly matters but adding it bc order makes my brain happy. can remove it if needed tho

refiner/migrations/20260602161536_replace_custom_code_system_key.sql
```diff
@@ -0,0 +1,72 @@
...
+    )
+    FROM jsonb_array_elements(c.custom_codes) WITH ORDINALITY AS t(elem, ord)
+), '[]'::jsonb) 
+WHERE c.custom_codes IS NOT NULL
+  AND jsonb_typeof(c.custom_codes) = 'array';
+
+ALTER TABLE events_custom_code_uploads RENAME COLUMN system TO system_key;
+
+UPDATE events_custom_code_uploads
```
fzhao99: adding one for the upload events table too so things are consistent across `custom_code`-related objects

refiner/tests/integration/test_configurations.py
```diff
@@ -445,6 +449,74 @@ async def test_section_updates_failure(
         # FastAPI shouldn't allow this to work
         assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
 
+    new_code = ["VERY-FAKE-CODE-00000-WITH-UPDATE", None]
+    new_system_key = ["snomed", "icd10"]
+    new_name = ["Mock code with update", None]
+
+    @pytest.mark.parametrize(
```
fzhao99, line 452: we somehow didn't have a test for editing of the custom codes across the various combinations of the possible fields, which I discovered when I accidentally broke system updating. Adding this new test with the combination of possible options to cover the cases for editing custom codes

refiner/app/db/configurations/db.py
fzhao99: updated with the migration for custom codes that also has a related change for the events table so that everything is consistent. I _think_ my testing for it showed it was working, but this is my first one I've done, so please double check me

refiner/app/services/ecr/specification/constants.py
fzhao99: Screenshot of the relevant comment bc GitHub seems to be having trouble with discussion links now 🤷 

Screenshot 2026-06-03 at 9 03 31 AM

refiner/migrations/20260602161536_replace_custom_code_system_key.sql
jakewheeler: I think `system` still made sense here since `system_key` isn't acting as a foreign key. The `system` value is a point in time snapshot and can stand on its own since it's only used for display.

refiner/migrations/20260602161536_replace_custom_code_system_key.sql
jakewheeler: good with me 👍

refiner/tests/integration/conftest.py
```diff
@@ -122,6 +122,21 @@ async def _get(config_id: UUID, custom_code: DbConfigurationCustomCode):
     return _get
 
 
+@pytest_asyncio.fixture
```
jakewheeler, line 125: 👍

refiner/migrations/20260602161536_replace_custom_code_system_key.sql
fzhao99: added some changes to put it back / work with the new shape of custom codes

refiner/app/services/configurations.py
```diff
@@ -227,7 +251,9 @@ async def convert_config_to_storage_payload(
         included_condition_rsg_codes.update(c.child_rsg_snomed_codes)
 
     # STEP 3: build the CodeSystemSets
-    code_system_sets = CodeSystemSets.from_dict(data=coding_by_code_system)
+    code_system_sets = CodeSystemSets.from_dict(
+        s3_data=coding_by_code_system, oid_to_system_map=OID_TO_SYSTEM_KEY_MAP
```
jakewheeler: It'd be helpful if `s3_data` was instead called `coding_by_code_system`. I was confused for a sec about where this data was coming from.

refiner/app/services/configurations.py
```diff
@@ -182,8 +191,15 @@ async def convert_config_to_storage_payload(
     # custom codes
     for cc in configuration.custom_codes:
         codes.add(cc.code)
-        cur_code_system = CodeSystem(cc.system)
-        system_to_extend = cur_code_system.format_system_string()
+        cur_code_system = await get_code_system_by_key_or_display_name(
```
jakewheeler: We will need to hit the DB for every custom code when doing a conversion to a storage payload. Is it possible to build a map outside the loop and reference it within? I think we did something similar elsewhere for the same reason.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -270,15 +253,23 @@ async def upload_custom_codes_csv(
...
         sanitized_system = None
+
         try:
-            sanitized_system = CodeSystem.sanitize(code_system_raw)
+            sanitized_system = systems_by_name.get(
+                code_system_raw
+            ) or systems_by_name.get(code_system_raw)
+            if sanitized_system is None:
+                raise ValueError(
+                    f"System of name {code_system_raw} doesn't match supported systems"
```
jakewheeler, line 260: Should we log this error message instead? This exception is going to be caught by the `except` below.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -527,12 +525,12 @@ def _get_modified_custom_codes(
     if len(code_to_edit) == 0:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
-            detail="Could not find custom code with specified system/code pair.",
+            detail="Could not find custom code with specified system_key/code pair.",
```
jakewheeler: I'm not sure how useful this is to show users. I'd probably leave it as `system` since they won't know what a `system_key` is most likely.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -527,12 +525,12 @@ def _get_modified_custom_codes(
...
         )
     if len(code_to_edit) > 1:
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail="Multiple custom codes with system/code pair found.",
+            detail="Multiple custom codes with system_key/code pair found.",
```
jakewheeler: Same thing here.

refiner/app/services/configurations.py
fzhao99: good catch! fixed

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -543,25 +541,29 @@ def _get_modified_custom_codes(
...
+        )
+    system_key = (
+        updateInput.new_system_key
+        if updateInput.new_system_key
+        else existing_code.system_key
     )
+
+    new_system = await get_code_system_by_key_db(key=system_key, db=db)
+    if new_system is None:
+        raise ValueError(f"System of name {system_key} doesn't match supported systems")
```
jakewheeler: I'd probably throw a 400 exception here instead since the client handed us the wrong system key.

refiner/tests/integration/test_configurations.py
```diff
@@ -445,6 +449,74 @@ async def test_section_updates_failure(
...
+        "new_code, new_system_key, new_name",
+        list(itertools.product(new_code, new_system_key, new_name)),
+    )
```
jakewheeler: (Nitpicking) I wouldn't mind these being defined inline rather than have the few variables floating right above.

```suggestion
    @pytest.mark.parametrize(
        "new_code, new_system_key, new_name",
        list(
            itertools.product(
                ["VERY-FAKE-CODE-00000-WITH-UPDATE", None],
                ["snomed", "icd10"],
                ["Mock code with update", None],
            )
        ),
    )
```

jakewheeler: And thanks for adding the new tests by the way 👍

refiner/tests/unit/helpers/code_systems.py
```diff
@@ -0,0 +1,33 @@
...
+    "snomed": {"oid": "2.16.840.1.113883.6.96", "display_name": "SNOMED"},
+    "loinc": {"oid": "2.16.840.1.113883.6.1", "display_name": "LOINC"},
+    "icd10": {"oid": "2.16.840.1.113883.6.90", "display_name": "ICD-10"},
+    "rxnorm": {"oid": "2.16.840.1.113883.6.88", "display_name": "RxNorm"},
+    "cvx": {"oid": "2.16.840.1.113883.12.292", "display_name": "CVX"},
+    "other": {"oid": "Other", "display_name": "Other"},
+}
+
+
+def get_mock_allowed_system_keys():
```
jakewheeler: Should we turn these functions into fixtures and stick them in `unit/conftest.py`?

refiner/tests/unit/conftest.py
```diff
@@ -110,6 +115,11 @@ def mock_user():
     )
 
 
+@pytest.fixture
```
jakewheeler: Oh 🤔 maybe this answers my question. I don't see any issue with the fixtures handling the logic themselves rather than calling out to other functions.

fzhao99: yep, will clean these up

refiner/app/db/configurations/db.py
```diff
@@ -748,20 +750,22 @@ async def add_bulk_custom_codes_to_configuration_db(
 
     existing_codes = config.custom_codes or []
 
-    # Build a set of (code, system) for fast lookup
-    existing_keys = {(c.code, c.system) for c in existing_codes}
+    # Build a set of (code, system_key) for fast lookup
+    existing_keys = {(c.code, c.system_key) for c in existing_codes}
 
     new_codes_added: list[DbConfigurationCustomCode] = []
+    system_info = await get_all_code_systems_by_key(db=db)
```
jakewheeler: Should we call this `code_systems`? We seem to refer to this `dict[CodeSystemKey, DbCodeSystem]` structure like this elsewhere.

Also wondering if you think there's any value in a defined type for `dict[CodeSystemKey, DbCodeSystem]` since it's a common type we're using across a few different files/functions? Maybe `type CodeSystemIndex = dict[CodeSystemKey, DbCodeSystem]`?

refiner/tests/unit/conftest.py
fzhao99: ok, think I simplified things down a bit. Lmk if what I came up with makes sense

jakewheeler (approved):
Went through it all one last time and everything seems to work 🫡  thanks Bob!

PR #1161, opened 2026-04-28
feat: App Update Banner 

### LINKED ISSUE #1028: [TASK] - Make updates banner that disappears on click per user
Split out of the banner UI in [this ticket](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/528)

lina-roth:
# 🔀 PULL REQUEST

Screenshot 2026-05-05 at 10 51 14 AM

## 🔗 Related Issue

Fixes #1028 
- #1028 

## ✅ Acceptance Criteria

- [ ]  Notifies the user when an update is made to the application through a banner (displayed per user until they click on it
- [ ] Added X to dismiss banner if user doesn't want to look at app updates (previously only way to dismiss would be to click View Updates and redirect to App Updates page)

## 🧪 How to test

To test locally - 

refiner/app/api/v1/releases.py lines 156-163
```
    for release in releases:
#         if release.prerelease:
#             continue

        release_content = release.body

#         if not release_content:
#             continue
```


client/src/pages/Configurations/index.test.tsx
```diff
@@ -73,7 +73,10 @@ const renderPageView = () =>
       <TestQueryClientProvider>
         <ToastContainer />
         <Routes>
-          <Route path="/configurations" element={<Configurations />} />
+          <Route
+            path="/configurations"
+            element={<Configurations user={undefined} />}
```
fzhao99, line 78: instead of passing in an un-initialized user object here, what do you think about calling `useLogin` within Configurations and managing that state within the component? 

Don't think that piece of UI is vital to the page, so we could probably even wrap all the banner logic into a sub-component and conditionally render things inside based on when the user object loads

client/src/pages/Configurations/index.tsx
```diff
@@ -51,10 +54,27 @@ interface ConfigurationsTable {
...
+interface ConfigurationsProps {
+  user?: UserResponse;
+  setUser?: Dispatch<SetStateAction<UserResponse | null>>;
+}
+
+export function Configurations({ user, setUser }: ConfigurationsProps) {
   const { data: response, isPending, isError } = useGetConfigurations();
   const configs = useMemo(() => response?.data ?? [], [response?.data]);
 
+  const { data: releaseFetchResult } = useGetReleases();
```
fzhao99: Think it's a bit cleaner of a separation of concerns if we move this logic into the AppUpdateBanner component and manage all the system-wide state closer to the UI we're rendering inside.

What do you think?

refiner/app/api/auth/handlers.py
```diff
@@ -216,6 +220,7 @@ class UserResponse(BaseModel):
     id: UUID
     username: str
     jurisdiction_id: str
+    dismissed_notifications: dict[str, str] = {}
```
fzhao99: Thinking about it a bit more, wondering if it's better here to explicitly type the field here as a class rather than a generic dict, with the field names being the notification type that we're tracking acknowledgement status for. That way, we're not indexing on loose strings and have Pydantic enforcing things for us

What do you think?

refiner/app/api/auth/handlers.py
```diff
@@ -301,3 +309,67 @@ async def logout(
...
     return response
+
+
+class UpdateDismissedNotificationRequest(BaseModel):
+    """
+    Request to update a dismissed notification timestamp for the current user.
+    """
+
+    key: str
+    value: str
```
fzhao99:
```suggestion
    name: str
    date_acknowledged: str
```

refiner/app/api/auth/handlers.py
```diff
@@ -301,3 +309,67 @@ async def logout(
...
+
+
+@auth_router.patch(
+    "/user/dismissed-notifications",
+    response_model=UserResponse,
+    tags=["user"],
+    operation_id="updateDismissedNotification",
+)
+async def update_dismissed_notification(
```
fzhao99: similar suggestion to use generic `update_user_notifications` language here in the function name without the mention of dismissal, handling that instead in the data itself

fzhao99 (commented):
left a few suggestions, but functionality is looking good!

client/src/pages/Configurations/index.tsx
```diff
@@ -67,41 +87,110 @@ export function Configurations() {
...
-            onChange={(e) => setSearchText(e.target.value)}
-          />
-        ) : null}
-
-        <Button className="m-0!" onClick={() => setIsOpen(true)}>
-          Set up new configuration
-        </Button>
-        <NewConfigModal open={isOpen} onClose={() => setIsOpen(false)} />
+    <>
+      {showAppUpdateBanner && latestRelease && setUser && (
```
fzhao99: thoughts on adding an e2e to test this code? Think we could mock the incoming releases API data using [Playwright mocking functionality](https://playwright.dev/docs/mock) and then assert on the interactions from there

client/src/pages/Configurations/index.test.tsx
jakewheeler: Just wanted to note that `useLogin` needs to be called at the top level of the app right now.

IMO we should have a mutation that runs when the user dismisses the notification and then, once it succeeds, resets the user data on the client. Similar to how we use `onSuccess` in a lot of other places.

For example:
```ts
      {
        onSuccess: async (resp) => {
            queryKey: getUserQueryKey(user.id),
          });
        },
```

This would take some additional work but wanted to throw the idea out there.

client/src/pages/Configurations/index.tsx
jakewheeler: Agreed. Any user facing functionality at this point should have at least one PW test associated with it.

client/src/pages/Configurations/index.tsx
```diff
@@ -67,41 +87,110 @@ export function Configurations() {
...
+              void dismissAppUpdate().catch((error) => {
+                console.error('Failed to update dismissed notification', error);
+              });
+            }}
+            className="font-public-sans text-violet-warm-60 border-violet-warm-60 flex h-[44px] items-center justify-center rounded-[4px] border-[2px] bg-white px-[20px] text-center text-[1rem] leading-[1.4rem] font-bold lining-nums proportional-nums no-underline"
+          >
+            View updates
+          </Link>
+        </div>
+        <button
```
jakewheeler: Can we make sure to use our `Button` component here? There is a `variant='unstyled'` that should work.

client/src/pages/Configurations/index.test.tsx
fzhao99: ahh ok. Lina, feel free to handle this in a follow up in that case

refiner/app/api/auth/handlers.py
```diff
@@ -301,3 +309,67 @@ async def logout(
...
+    request: UpdateDismissedNotificationRequest,
+    http_request: Request,
+    db: AsyncDatabaseConnection = Depends(get_db),
```
jakewheeler: Should we be getting `user` in the same way we get `db`? A user must be logged in to dismiss a notification so we should already have that data via `Depends`.

I'm also wondering if we should move this into a new file since it's not related to authentication.

refiner/migrations/20260505141110_add_dismissed_notifications_to_users.sql
```diff
@@ -0,0 +1,10 @@
+-- migrate:up
+
+ALTER TABLE users
```
jakewheeler: Does this need to be JSONB? Would a boolean do the job?

client/src/pages/Configurations/index.tsx
```diff
@@ -51,10 +54,27 @@ interface ConfigurationsTable {
...
+  const latestRelease = releaseFetchResult?.data.releases?.[0];
+
+  const dismissedMostRecentAppUpdate =
+    user?.dismissed_notifications?.most_recent_app_update;
+
+  const showAppUpdateBanner =
+    latestRelease &&
+    (!dismissedMostRecentAppUpdate ||
+      new Date(latestRelease.created_at) >
+        new Date(dismissedMostRecentAppUpdate));
```
jakewheeler: We should handle the logic for whether or not to show the banner to the user on the server and then return that result to the client to decide what to render.

client/src/pages/Configurations/index.test.tsx
lina-roth: That makes sense. Since `useLogin` currently owns the user state at the top level of `App` and isn’t backed by React Query yet, I handled this by using the `setUser` dispatcher returned from `useLogin`.

So the flow is similar in intent to an `onSuccess` mutation update:

1. User clicks **View updates** or dismisses the banner.
2. `updateDismissedNotification` runs.
3. On success, the response returns the updated `UserResponse`.
4. We call `setUser(resp.data)` so the client-side user state reflects the updated `dismissed_notifications` value immediately.

That avoids needing a hard refresh and keeps the banner hidden after route changes.

I agree that longer-term, moving user/session data into React Query would let us use the more standard `invalidateQueries` pattern here. For this PR, I kept the change scoped to the current `useLogin` pattern rather than introducing a larger refactor.

client/src/pages/Configurations/index.tsx
lina-roth: Agreed thats a good suggestion

refiner/migrations/20260505141110_add_dismissed_notifications_to_users.sql
lina-roth: A boolean works for one notification, JSONB lets us add more without schema changes.

If this were a boolean, we’d need:

a new column per notification type, or
a migration every time we add a new one

With JSONB, we can just add another key:

```
{
  "most_recent_app_update": { "date_acknowledged": "..." },
  "new_feature_banner": { "date_acknowledged": "..." }
}
```

Which I think is what Bob intended this to be

client/src/pages/Configurations/index.tsx
lina-roth: I think that could make sense longer term if we end up with more complex notification/banner rules, but for now the client already has both pieces of data needed to determine visibility:

* latest release metadata
* the user’s acknowledged notification timestamp

So I kept the visibility logic local to the UI rather than introducing additional server-side derived state.

That said, I do agree there’s value in centralizing the logic eventually if we expect more complicated visibility rules.

refiner/migrations/20260505141110_add_dismissed_notifications_to_users.sql
jakewheeler: I think I'd prefer to not introduce JSONB for newer things since we're attempting to move away from it but will defer to @fzhao99 on this one. I'm not 100% against it if others think this is the right path at this time.

client/src/pages/Configurations/index.tsx
jakewheeler: Is there any reason not to make the change now so that we can determine whether or not to show the banner solely based off data coming from the `/api/user` response since this is a request we already need to make?

The main reason I think it'd make the most sense to make this switch is that the configurations page will now need to fetch the entire history of releases only to determine whether or not to show the banner to the user. It seems useful to completely eliminate that request from needing to be made if that determination can happen server-side. This list of releases is small for now but will grow over time.

Does that make sense?

refiner/migrations/20260505141110_add_dismissed_notifications_to_users.sql
fzhao99: I think I'm ok with having JSON here, but let's talk about it at the next eng meeting for where / where not to use it with all the changes we're making

client/src/pages/Configurations/index.tsx
lina-roth: Yep I follow you. I think I have a good solution for it

client/e2e/app-notifications.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  test('sends notification acknowledgement request when app update banner is dismissed', async ({
+    page,
+  }) => {
+    const updateRequestPromise = waitForAcknowledgementRequest(page);
+
+    await page.getByRole('button', { name: 'Dismiss notification' }).click();
+
+    const updateRequest = await updateRequestPromise;
```
fzhao99, line 98: I think instead of asserting on the structure of the request body here, we should just be checking what's on the page before / after the user action. These asserts aren't super useful IMO

refiner/app/api/v1/releases.py
```diff
@@ -207,3 +207,15 @@ def _format_api_body_to_dict(content: str) -> list[ReleaseNotes]:
...
+
+
+def get_latest_release_created_at() -> datetime | None:
+    """
+    Returns the created_at timestamp for the latest app release.
+    """
+    releases = _get_releases_data_from_github(ttl_hash=_get_ttl_hash())
+
+    if not releases:
+        return None
```
fzhao99: instead of returning None here, we can return the minimum epoch time to represent "the latest release was never created". This helps us simplify some things in the caller code

```suggestion
        return datetime.min
```

client/src/pages/Configurations/index.test.tsx
lina-roth: Making a ticket to capture this work [here](https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/1220)

client/e2e/app-notifications.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  test.beforeEach(async ({ page, configurationsPage }) => {
+    await deleteAllConfigurations();
+
+    await page.route(
+      (url) =>
+        url.pathname === '/api/releases/getReleases' ||
+        url.pathname === '/api/releases/getReleases/' ||
+        url.pathname === '/api/releases' ||
+        url.pathname === '/api/releases/',
```
lina-roth, line 24: Hmm I see what you're saying. Let me look into it - I'm not sure how difficult that would be or how feasible

client/e2e/app-notifications.spec.ts
lina-roth: I feel like asserting that the API call to update the user notifications fired off is better than checking to see that an element is no longer visible, but I can add it if you feel that it a stronger assertion

client/e2e/app-notifications.spec.ts
fzhao99: GitHub did something weird here with my comments, so for posterity: we were discussing mocking the API at the GitHub level here to be able to cover the backend code processing incoming data from the releases API. We opted to punt this to the followup since it'll be a lot more involved

fzhao99 (approved):

PR #1162, opened 2026-04-28
fix: XML files with UTF-8 encoding declaration can't be processed in testing workflows

### LINKED ISSUE #1102: [TASK] Account for UTF Encoding
## 🚧 What needs to be done

We need to be able to handle files with UTF Encoding

## 🤔 Why it needs to be done

Need to be able to refine UTF encoded files

## 🏗️ Acceptance Criteria

- [ ] UTF-encoded eICRs can be refined

## ⚙️ Technical Notes

- ...

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes a `500 server error` users were seeing when attempting to upload XML files containing a UTF-8 encoding declaration line. 

In those situations, this line:
```xml
<?xml version="1.0" encoding="UTF-8"?>
```

was seen at the top of the file.

### Why was this a problem?

When a user uploaded files containing that line, they would receive a 500 error and we'd see this `lxml` error in the logs:

```sh
ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.
```

This error is pretty clear. We were giving `lxml` a Python string but the string contained an encoding declaration saying to encode to UTF-8. This is technically an error on our side and `lxml` refuses to try to fix it for us.

## 🔗 Related Issue

Fixes #1102 

## 🧪 How to test

- Run the automated tests
- Try going through the test flow. Here are my modified mon mothma test files with the declaration if you'd like to see that it works: [mon-mothma-covid-influenza xml tag.zip](https://github.com/user-attachments/files/27204675/mon-mothma-covid-influenza.xml.tag.zip)


refiner/app/api/validation/file_validation.py
```diff
@@ -10,13 +11,29 @@
...
+    preserve_comments: bool = False,
+) -> str:
+    """
+    Formats XML for display purposes. Raises a 422 if the input is not valid XML.
+    """
+    try:
+        return format_xml_document_for_display(text, preserve_comments)
+    except etree.XMLSyntaxError as e:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=f"Invalid XML: {e}"
```
jakewheeler: In this situation, I feel like giving the user the exact reason for the `XMLSyntaxError` is the right way to go. I don't really see any risk of providing too much here. What do you think?

refiner/app/services/format.py
```diff
@@ -64,7 +64,8 @@ def _normalize_xml(xml: str) -> str:
     xml = SPACE_BEFORE_FIRST_ATTR.sub(r"<\1 ", xml)
 
     parser = etree.XMLParser(remove_blank_text=True)
-    root: _Element = etree.fromstring(xml, parser=parser)
+    xml_bytes = xml.encode("utf-8")
+    root: _Element = etree.fromstring(xml_bytes, parser=parser)
```
jakewheeler, line 68: There are two main ways we could go about this:
1. Convert to bytes (what I did here)
2. Remove the encoding line

Either is valid. I decided to go with converting to bytes since that appears to be the [typical way](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/services/file_io.py#L103) we go about handling XML data. Let me know if you disagree with this decision.

refiner/app/api/validation/file_validation.py
fzhao99: would there be a possibility of the XML string highlighting the chunk of the string that's erroring and thus having PII? That's the only worry that I have here

refiner/app/services/format.py
fzhao99: yeah think an encode is the right way to go. if possible, not messing with the input string seems like the least error prone way of handling things

refiner/app/api/validation/file_validation.py
jakewheeler: Technically yes. Although if a user is testing with PII that's a big problem 😅  It's a good point though and perhaps we should still err on the side of caution regardless.

Looking at the [docs](https://lxml.de/apidoc/lxml.etree.html#lxml.etree.LxmlSyntaxError), we could go for something more targeted like this instead:
```python
except etree.XMLSyntaxError as e:
    raise HTTPException(
        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
        detail=f"Invalid XML: {e.msg} (line {e.lineno}, column {e.offset})",
    )
```
`msg` can include attribute names but won't include content. Does this seem like a better alternative?

fzhao99: think this is a good guard! thanks

fzhao99 (approved):
🚢

PR #1164, opened 2026-04-29
refactor: migrate to HeadlessUI Pagination component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

Title is not entirely accurate 😄 

This PR is a custom pagination component based off of the `react-uswds` component. It functions identically but does not depend on that library or `uswds` styles at all. Unfortunately HeadlessUI doesn't provide its own pagination component out of the box but what's here should cover all of our needs, which are admittedly not very extensive at the moment.

The component's API is basically identical to the `react-uswds` component. As noted, we only use this component on the activity log page currently but this component is generic and can be used anywhere in the app.

## 📺 Demo
Kapture 2026-04-30 at 14 26 58

## 🧪 How to test

1. Log into the app, set up a configuration, and do a bunch of things that will generate events (adding/removing code sets, adding/deleting custom codes, etc.)
2. Navigate to the activity log
3. Check that the pagination component looks and works as expected


client/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+  totalPages,
+  currentPage,
+  className,
+  maxSlots = 7,
+  onClickPrevious,
+  onClickNext,
+  onClickPageNumber,
+  ...props
+}: PaginationProps) {
+  const isOnFirstPage = currentPage === 1;
```
jakewheeler, line 34: This logic is the same as the `react-uswds` component. Feel free to make suggestions on any of this if you'd like. I can't imagine we'll need to touch this much again in the future.

client/src/components/Pagination/index.test.tsx
```diff
@@ -0,0 +1,179 @@
+import { render, screen } from '@testing-library/react';
```
jakewheeler, line 1: In the beginning I used the react-uswds tests to check myself and they were all passing. These tests are based off of those but are no longer exactly the same.

client/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+}) {
+  const baseClasses = classNames(
+    'relative inline-flex items-center justify-center w-10 h-10 font-medium rounded hover:cursor-pointer underline',
+    {
+      'z-10 bg-blue-cool-80 underline text-white focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-blue-600':
+        isCurrent,
+    },
+    {
+      'text-blue-cool-60 ring-1 ring-inset ring-gray-400 hover:ring-gray-600 focus:outline-offset-0':
+        !isCurrent,
```
fzhao99, line 232: 💪

client/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+    if (currentPage !== 1) currentPageRange.unshift(1);
+    if (showNextOverflow) currentPageRange.push('overflow');
+    if (totalPages && currentPage !== totalPages)
+      currentPageRange.push(totalPages);
+  }
+
+  const prevPage = !isOnFirstPage && currentPage - 1;
+  const nextPage = !isOnLastPage && currentPage + 1;
+
+  const prevNextBase = classNames(
```
fzhao99, line 92: think it'd be more readable to pull these into a constant at the top of the file along the lines of `PREV_NEXT_BASE_STYLES`, (and something similar for the other styles in PaginationPage) and then use those constants accordingly here

fzhao99 (approved):
one suggestion, but otherwise LGTM!

PR #1165, opened 2026-04-29
fix: make application gracefully handle large files

### LINKED ISSUE #1134: [BUG] Diff Viewer fails to render files above a certain size
## 📜 Description

Diff Viewer fails to render files once they reach a certain size and can cause the app to crash. Regardless of the solution, we need to ensure:
1. The app does not crash or become unresponsive for the user
2. (If necessary) the user is informed that the file is too big for the diff viewer to render and is directed to download the file instead

## 👣 Steps to reproduce

1. Load up the app
2. Use any testing flow with a large file (20MB uncompressed in this case)

## 🌃 Environment:

- OS: Any
- AIMS Dev

## ??? Expected Result

Diff Viewer should be able to render any file regardless of size

## 🧪 Actual result

Diff Viewer fails to render above a certain size

#### Priority

- [X] **Medium priority** _Typically, medium-priority bugs do not affect customers directly and therefore can be fixed in the normal course of testing and development._
- [ ] **High priority** _High priority bugs must be addressed immediately. They often affect the app’s functionality and impact customers and their user experience. They must take priority._

## 📓 Notes

- Start by finding a reasonable return size
- Find out why it's crashing APHL's server and the app

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Changes to reflect graceful fallback of zip uploads that 

1. Will brick the browser trying to load the diff view (capped at 2MB) and suggest to the user to download the files instead. In these instances, we won't send the refined eCR across the network / save on the request size

https://github.com/user-attachments/assets/446dd8ce-c0fb-4d09-9a71-4b8e34edf27f

2. Will brick the server during processing (capped at 15MB) and notify the user about the issue.
https://github.com/user-attachments/assets/8fd7f80f-fd53-4eef-b73a-8f8cb59e6937

 
## 🔗 Related Issue

Fixes #1134 
- #1134 

## 🧪 How to test
- Set up the standard influenza / COVID configurations.
- Uploaded test files [here](https://drive.google.com/drive/folders/15dh6ZTGz5nc60D7CRONfSaz219yjw-dm?dmr=1&ec=wgc-drive-globalnav-goto) to support simulating the following scenarios:

### Local (using the sddh-covid files)
- Render fallback diff warning for files between 2MB - 15MB 
- Render error message for files above 15MB (local)

### [APHL dev](https://refiner.dev.sandbox2.aimsplatform.org/)
- Make sure the inline and independent test flows don't crash the server (generate 502 errors) by verifying the uploads of the [`test-covid-14.9-mb.zip`](https://drive.google.com/file/d/1xBXlvTawsU0vWgYlTOjx3mz5Bfz1nOaW/view?usp=drive_link) files. The browser will try to render the diff and potentially brick without the flag for conditionally rendering the diff view.


client/src/components/Diff/DiffToggleOptions.tsx
```diff
@@ -0,0 +1,131 @@
+import { Button } from '@components/Button';
```
fzhao99, line 1: Splitting this out into its own page to help with conditional rendering

client/src/pages/Testing/ReportableConditionsResults.tsx
```diff
@@ -71,7 +71,7 @@ export function ReportableConditionsResults({
             relevant only to that condition as defined in its configuration.
           </p>
         </div>
-        <div className="flex flex-col gap-4 sm:flex-row sm:gap-0">
```
fzhao99, line 74: this was causing spacing issues between the CTA's, so I flipped it since I think we meant to have the reverse spacing

fzhao99: Before:
Image

After:
Image

client/src/pages/Configurations/ConfigBuild/ConditionCodeSetListItem.tsx
```diff
@@ -37,9 +37,11 @@ export function ConditionCodeSetListItem({
   const queryClient = useQueryClient();
   const formatError = useApiErrorFormatter();
 
-  React.useEffect(() => {
+  const [prevAssociation, setPrevAssociation] = useState(condition.associated);
```
fzhao99, line 40: this was getting flagged with the useEffect hook once I delected the `React.`, so refactoring this according to [this guidance](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

client/src/pages/Testing/Success.tsx
```diff
@@ -34,28 +37,31 @@ export function Success({
 
   return (
     <div>
-      <div className="flex items-center gap-4">
+      <div className="flex place-items-center justify-between gap-4 pr-2">
```
fzhao99, line 37: some restyling here per the conversation with Tiffany [here](https://skylight-hq.slack.com/archives/C08N3EKJGRL/p1777492522053879):

client/src/components/Diff/index.tsx
```diff
@@ -59,12 +70,24 @@ export function Diff({
           </div>
           <div>
             <div className="flex flex-col items-start gap-1 py-1">
-              <Button
-                variant="tertiary"
-                onClick={() => downloadFile(refined_download_key)}
-              >
-                Download results
-              </Button>
+              {isDownloading ? (
```
fzhao99: in instances where we had larger files, the download state was taking enough time that i wanted some sort of visual feedback to the user to indicate the process was happening. Went ahead and reused the styling for our saving indicator here

https://github.com/user-attachments/assets/ee50a29e-a1e8-40b7-9854-19e62264552e

refiner/app/api/validation/file_validation.py
```diff
@@ -15,8 +15,12 @@
 from app.services.sample_file import create_sample_zip_file
 
 # File uploads
-MAX_ALLOWED_UPLOAD_FILE_SIZE = 10 * 1024 * 1024  # 10 MB
-MAX_ALLOWED_UNCOMPRESSED_FILE_SIZE = MAX_ALLOWED_UPLOAD_FILE_SIZE * 5  # 50 MB
+MEGABYTES = 1024 * 1024
+DIFF_RENDERING_MAX_MB = 2
+UNCOMPRESSED_MAX_MB = 15
```
fzhao99: there are a few different scenarios that were causing crashes in the dev env

1. The uncompressed files were causing the server to crash, which seemed to happen ~20MB uncompressed in the independent testing flow and ~17.5MB in the inline test flow. I opted just to take the lower threshold for the inline test flow with some buffer, which is how I came to the 15 number here. You can verify this by uploading the file and verifying the serve doesn't return a 502.
     - if folks think we should have separate file sizes for the independent versus inline testing flow, lmk and we can add that 
3. The file got processed fine, but the diff view was too big, causing the browser to brick. This seemed to happen around 2MB in both views (after I refactored the frontend to use the same component). Thus, added this secondary 2MB threshold

client/src/components/Diff/index.tsx
```diff
@@ -1,29 +1,39 @@
...
 type DiffProps = Pick<
   IndependentTestUploadResponse,
   'refined_download_key' | 'unrefined_eicr'
 > & {
   condition: IndependentTestUploadResponse['refined_conditions'][0];
+  renderDiff: boolean;
 };
+const loadingFragment = () => (
+  <div className="m-2 h-6 text-center">Computing diff...</div>
+);
```
jakewheeler: Can we either:
1. Update this to use the typical component style (`function Loading()`)
2. Inline the `div`

client/src/components/Diff/index.tsx
```diff
@@ -40,6 +50,7 @@ export function Diff({
       link.click();
       URL.revokeObjectURL(blobUrl);
       setDownloadError(null);
+      setIsDownloading(false);
```
jakewheeler: We need this line in both the `catch` block and `if (!response.ok)` block as well.

client/src/components/Diff/index.tsx
jakewheeler: Did this come from design? We also have `Spinner` which we use in other spots for loading currently.

client/src/pages/Testing/Success.tsx
jakewheeler: Small nitpick but I liked the old position better when full screening the page on my ultrawide monitor. It wouldn't be too bad if we had a global max width (which Ryan has suggested in the past). Probably won't impact many people beyond myself currently but just wanted to mention it.
Image

client/src/pages/Testing/ReportableConditionsResults.tsx
jakewheeler: Thanks! I noticed this as well but never got around to fixing it.

client/src/pages/Home/index.tsx
```diff
@@ -10,7 +10,7 @@ export function Home() {
         <div className="flex max-w-268 flex-col items-center gap-16 lg:flex-row">
           <div className="flex flex-col gap-8 px-10 lg:w-2/3 xl:px-0">
             <div className="flex flex-col items-start gap-4">
-              <h1 className="font-merriweather font-bold lg:!text-5xl">
+              <h1 className="font-merriweather font-bold lg:text-5xl!">
```
jakewheeler, line 13: 👍

refiner/app/api/v1/info.py
```diff
@@ -65,3 +71,33 @@ async def get_info(
             },
         }
     )
+
```
jakewheeler: There are a few ways we could go about this but I think we can skip this network request entirely if we include the `FileInfoResponse` info as part of the responses for independent testing and inline testing. That way, Orval will generate client side values we can use directly with no request needed.

I'm thinking something like this:

Add file size info to test response model
```python
@dataclass
class FileInfoResponse:
    """
    Response for file upload thresholds.
    """

    max_mb_for_diff_rendering: Literal[2] # maybe we can use our constant values here somehow instead
    max_mb_for_uncompressed: Literal[15]


@dataclass(frozen=True)
class IndependentTestUploadResponse:
    """
    Model for the response when uploading a document in the testing suite.
    """

    message: str
    conditions_without_matching_configs: list[str]
    conditions_without_active_configs: list[str]
    refined_conditions_found: int
    refined_conditions: list[Condition]
    unrefined_eicr: str
    refined_download_key: str
    render_condition_map: dict[str, bool]
    file_info_response: FileInfoResponse
```

Update testing route responses
```python
    return IndependentTestUploadResponse(
        message="Successfully processed eICR with condition-specific refinement",
        refined_conditions_found=len(conditions),
        refined_conditions=conditions,
        conditions_without_matching_configs=test_results.get_condition_names_with_no_matching_config(),
        conditions_without_active_configs=test_results.get_condition_names_with_no_active_config(),
        unrefined_eicr=format_xml_document_for_display_or_raise(
            original_xml_files.eicr, preserve_comments=True
        )
        if send_eicr_to_frontend
        else "",
        refined_download_key=output_file_name if s3_key else "",
        render_condition_map=test_results.render_condition_map,
        file_info_response=FileInfoResponse(),
    )
```
Orval generates a client side model with the values
```python
export const AppDbDemoModelFileInfoResponseValue = {
  max_mb_for_diff_rendering: 2,
  max_mb_for_uncompressed: 15,
} as const;
export type AppDbDemoModelFileInfoResponse = typeof AppDbDemoModelFileInfoResponseValue;
```

This approach seems preferable to me but I'm curious what you think. If we want to stick with the network request my suggestion would then be to move this out of `/info` since it wasn't really intended for this. I'd move the request to the test routes instead.

refiner/app/api/v1/info.py
```diff
@@ -65,3 +71,33 @@ async def get_info(
...
+
+@dataclass
+class FileInfoResponse:
+    """
+    Response for file upload thresholds.
+    """
+
+    max_mb_for_diff_rendering: int
+    max_mb_for_uncompressed: int
```
jakewheeler: We use size as a suffix in a few other places already so I might suggest we do the same here.

```suggestion
    diff_rendering_max_size_mb: int
    uncompressed_zip_max_size_mb: int
```

or something else along these lines.

refiner/app/api/validation/file_validation.py
jakewheeler: In your testing did #1 happen only in AIMS and not locally?

I think if your testing for #2 seemed good we should just go with that but keep in mind this may not be the true fix. I think it's going to come down to how many nodes need to be rendered in the DOM more so than file size directly. It's a good place to start though.

refiner/app/services/file_io.py
```diff
@@ -210,22 +209,28 @@ def _is_valid_uncompressed_size(info: list[ZipInfo]) -> bool:
     Returns:
         bool: True if the total uncompressed size is less than 50 MB; otherwise, False.
     """
-    return sum(zinfo.file_size for zinfo in info) < MAX_UNCOMPRESSED_SIZE
+
+    from app.api.validation.file_validation import UNCOMPRESSED_MAX_BYTES
```
jakewheeler, line 213: Do we run into an issue if we have this at the top of the file?

refiner/app/api/validation/file_validation.py
fzhao99: Yep, tested 1 on AIMS 

And sounds good, can keep an eye on this and see if there are performance complaints to adjust further

refiner/app/services/file_io.py
fzhao99: yeah there were circular import issues importing at the top of the file so Imported this (and the one in the other function) in the function

refiner/app/api/v1/info.py
fzhao99: hmm yeah I was trying to get this to work like this, definitely preferable to have them be Orval generated. I'll play around and see if I can't get the types to generate with the variables.

I think ideally, we have a way of the frontend to use the backend values over re-declaring them. If I can't get this to work like you suggested, I'll move them

fzhao99: figured out a way to do this! Thanks for pointing me in the right direction

client/src/components/Diff/index.tsx
fzhao99: Yeah checked with Tiffany and she preferred the inline arrow animation, so refactored some code from the build page to use it again here

client/src/components/Diff/index.tsx
fzhao99: lmk if the new inline component is better

refiner/app/api/v1/configurations/testing.py
```diff
@@ -238,9 +239,13 @@ async def run_configuration_test(
     formatted_refined_eicr = format_xml_document_for_display_or_raise(
         refined_document.refined_eicr
     )
+    render_diff = (
+        len(formatted_unrefined_eicr.encode("utf-8")) < DIFF_RENDERING_MAX_BYTES
```
jakewheeler: Can we use the existing `get_file_size_in_bytes()`?

client/src/components/Diff/index.tsx
```diff
@@ -43,9 +55,22 @@ export function Diff({
...
+  const [shouldShowSpinner, setShouldShowSpinner] = useState(false);
+  const spinnerStart = useRef<number>(0);
+
+  useEffect(() => {
+    void showSpinnerWithMinimalRenderDuration(
+      spinnerStart,
+      setShouldShowSpinner,
+      isDownloading
+    );
+  }, [isDownloading]);
```
jakewheeler: Should we convert `showSpinnerWithMinimalRenderDuration` into a hook and handle everything in there (including the `useEffect`?)

refiner/app/services/testing.py
```diff
@@ -326,6 +332,10 @@ async def independent_testing(
             trace=pipeline_trace,
         )
 
+        render_dict[trace.matching_condition.display_name] = (
+            len(result.refined_eicr.encode()) < DIFF_RENDERING_MAX_BYTES
+        )
```
jakewheeler: Instead of doing this here, can we do it as part of the `Condition` that is returned by the route?

For example, in `demo.py`:
```python
        conditions.append(
            Condition(
                code=condition.code,
                refined_rr=format_xml_document_for_display_or_raise(
                    original_xml_files.rr
                ),
                render_diff: get_file_size_in_bytes(formatted_refined_eicr) < DIFF_RENDERING_MAX_BYTES, 
                stats=[
                    f"eICR file size reduced by {
                        get_file_size_reduction_percentage(
                            unrefined_eicr=original_xml_files.eicr,
                            refined_eicr=refined_document.refined_eicr,
                        )
                    }%",
                ],
            )
        )
```
Something like this?

client/src/components/Diff/index.tsx
fzhao99: ahh yes, good suggestion. will do!

client/src/components/Diff/index.tsx
jakewheeler: That works, thanks!

refiner/app/services/testing.py
fzhao99: ahh this is much cleaner. Lmk if what I've done instead makes sense

client/src/hooks/useMinimalSpinner.tsx
```diff
@@ -0,0 +1,53 @@
...
+        spinnerStart.current = 0;
+      }
+    }
+
+    void showSpinnerWithMinimalRenderDuration();
+  }, [minimalRenderDuration, stopShowingSpinnerConditional]);
+
+  return {
+    shouldShowSpinner: shouldShowSpinner,
+    minimalRenderSpinner: (
```
fzhao99: we could simplify this down to something like 

``` 
shouldShowSpinner ? <>Spinner </> : null
```
and have the caller handle the null state with whatever we should be rendering instead, but I exposing the boolean is a bit cleaner of an interface. Lmk if we prefer the alternative option

jakewheeler: I think that is a better idea. The hook returning JSX feels a bit wrong to me since it could really just be a component at that point. 

I am thinking maybe we should create a `TimedSpinner` that uses this hook to determine what it should render.

I think the API for the `TimedSpinner` (or whatever we want to call it) could look something like this.

Case 1 (default, flip to "Saved" message):
```ts
<TimedSpinner isLoading={numSavingActions > 0} />
```

Case 2 (rendered element is different on "done"):

this
```ts
{shouldShowSpinner ? (
  minimalRenderSpinner
) : (
  <Button
    variant="tertiary"
    onClick={() => downloadFile(refined_download_key)}
  >
    Download results
  </Button>
)}
```

becomes

```ts
<TimedSpinner isLoading={numSavingActions > 0} renderDone{() => (
  <Button
    variant="tertiary"
    onClick={() => downloadFile(refined_download_key)}
  >
    Download results
  </Button>
)}
/>
```
Does that make sense? Hopefully this is clear enough but I can try to give more specifics if it'd be helpful.

refiner/app/api/validation/file_validation.py
```diff
@@ -15,8 +17,17 @@
 from app.services.sample_file import create_sample_zip_file
 
 # File uploads
-MAX_ALLOWED_UPLOAD_FILE_SIZE = 10 * 1024 * 1024  # 10 MB
-MAX_ALLOWED_UNCOMPRESSED_FILE_SIZE = MAX_ALLOWED_UPLOAD_FILE_SIZE * 5  # 50 MB
+MEGABYTES = 1024 * 1024
+
+# defining these type literals to get Orval to pick them up and codegen them to the frontend
+DiffMax = Literal[2]
+UncompressedMax = Literal[15]
```
jakewheeler, line 24: Nice. I was curious what exactly we'd need to do to satisfy mypy here.

refiner/app/api/v1/configurations/testing.py
```diff
@@ -238,9 +242,13 @@ async def run_configuration_test(
     formatted_refined_eicr = format_xml_document_for_display_or_raise(
         refined_document.refined_eicr
     )
+    render_diff = (
+        get_file_size_in_bytes(formatted_unrefined_eicr) < DIFF_RENDERING_MAX_BYTES
+    )
+    original_eicr = formatted_unrefined_eicr if render_diff else ""
```
jakewheeler: Do we need to prevent the original RR and refined documents from being sent to the client if `render_diff == False`?

fzhao99: oh don't know how I overlooked this. Changed this to conditionally return those pieces if the diff isn't being rendered

client/src/hooks/useMinimalSpinner.tsx
fzhao99: lmk if what I changed it to hits what you were thinking

client/src/components/Spinner/SpinnerWithMinimalRender.tsx
```diff
@@ -0,0 +1,79 @@
...
+type SpinnerWithMinimalRenderProps = {
+  isLoading: boolean;
+  renderWhenDone: React.ReactNode;
+  loadingMessage?: string;
+  minimalRenderDuration?: number;
+};
+
+const TWO_SECONDS_IN_MILLISECONDS = 2000;
+
+export function SpinnerWithMinimalRender({
```
jakewheeler, line 14: Do we want to change the name to match the file? I assume this name probably got changed later.

client/src/components/Spinner/SpinnerWithMinimalRender.tsx
```diff
@@ -0,0 +1,79 @@
...
+  isLoading,
+  renderWhenDone,
+  loadingMessage = 'Saving',
+  minimalRenderDuration = TWO_SECONDS_IN_MILLISECONDS,
+}: SpinnerWithMinimalRenderProps) {
+  const shouldShowSpinner = useDisplaySpinnerWithMinimalRender(
+    isLoading,
+    minimalRenderDuration
+  );
```
jakewheeler, line 23: 💯  seems to work well!

refiner/app/api/v1/configurations/testing.py
```diff
@@ -232,23 +236,34 @@ async def run_configuration_test(
...
+        original_eicr = format_xml_document_for_display_or_raise(
+            original_xml_files.eicr
+        )
+        refined_eicr = format_xml_document_for_display_or_raise(
+            refined_document.refined_eicr
+        )
+        refined_rr = format_xml_document_for_display_or_raise(
+            refined_document.refined_rr
+        )
+
```
jakewheeler: Do we need the equivalent of this in `demo.py` as well? We could make this into a function and stick it in `api/validation/file_validation.py` to share between testing routes?

client/src/hooks/useMinimalSpinner.tsx
jakewheeler: You nailed it. Great job!

refiner/app/api/v1/configurations/testing.py
fzhao99: factored this into a shared function and called in both places. lmk if it looks good

refiner/app/services/conditions.py
```diff
@@ -71,3 +75,55 @@ def create_condition_mapping_payload(
...
+
+    Args:
+        original_xml_files: Inputted files.
+        refined_document: Refined documents to potentially ship to potentially render in the diff
+
+    Returns:
+        FilesFor - with values being the strings to send to the frontend.
+
+    """
+    from app.api.validation.file_validation import (
```
fzhao99, line 105: imported here because of circle import issues on this module

jakewheeler (approved):
Thanks Bob!

PR #1166, opened 2026-04-30
chore: update orval and fix packages

fzhao99:
some misc changes that came up in the large files PR that I've split out to slim down the real changes


client/src/pages/Configurations/ConfigBuild/ConditionCodeSetListItem.tsx
```diff
@@ -37,9 +37,11 @@ export function ConditionCodeSetListItem({
   const queryClient = useQueryClient();
   const formatError = useApiErrorFormatter();
 
-  React.useEffect(() => {
+  const [prevAssociation, setPrevAssociation] = useState(condition.associated);
```
fzhao99, line 40: this got flagged when I updated the packaging / deleted the `React.` by the linter, so refactored things according to the guidance [here](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

client/src/pages/Testing/RunTest.tsx
```diff
@@ -152,16 +153,9 @@ function UploadZipFile({
...
-        <span className="flex items-center gap-2">
-          <WarningIcon role="presentation" />
-          <span>This environment is not approved to handle PHI/PII.</span>
-        </span>
-        <span className="text-center font-bold">
-          Do not upload files that contain PHI/PII.
-        </span>
-      </p>
-    </div>
+    <Warning
```
fzhao99, line 157: pulled this into a component per some new designs

client/src/components/Diff/Warning.tsx
```diff
@@ -0,0 +1,21 @@
+import { Icon } from '@trussworks/react-uswds';
+
+type Warning = {
+  heading: string;
+  message: string;
+};
+
+export function Warning({ heading, message }: Warning) {
```
jakewheeler:
```suggestion
type WarningProps = {
  heading: string;
  message: string;
};

export function Warning({ heading, message }: WarningProps) {
```

client/src/components/Diff/Warning.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <div className="text-state-error-dark bg-state-error-lighter flex w-100 items-start p-4">
+      <Icon.Warning
+        aria-label="Warning icon to visually indicate issue with the page"
+        className="-mt-0.5 h-5! w-7!"
+      ></Icon.Warning>
```
jakewheeler:
```suggestion
      />
```

client/src/components/Diff/Warning.tsx
```diff
@@ -0,0 +1,21 @@
...
+type Warning = {
+  heading: string;
+  message: string;
+};
+
+export function Warning({ heading, message }: Warning) {
+  return (
+    <div className="text-state-error-dark bg-state-error-lighter flex w-100 items-start p-4">
+      <Icon.Warning
+        aria-label="Warning icon to visually indicate issue with the page"
```
jakewheeler: It might make the most sense to hide the icon from screen readers entirely. What do you think?
```suggestion
        aria-hidden
```

client/src/components/Diff/Warning.tsx
```diff
@@ -0,0 +1,21 @@
...
+        className="-mt-0.5 h-5! w-7!"
```
jakewheeler: If we restructure this component a bit we can use flex box and avoid using margin. Double check me but this seems to look about right
```tsx
    <div className="text-state-error-dark bg-state-error-lighter flex w-100 flex-col p-4">
      <div className="flex gap-4">
        <Icon.Warning className="h-5! w-7!" aria-hidden />
        <div className="flex flex-col gap-2">
          <p className="font-bold">{heading}</p>
          <p>{message}</p>
        </div>
      </div>
    </div>
```

client/src/pages/Configurations/ConfigBuild/ConditionCodeSetListItem.tsx
jakewheeler: I ran into this recently as well and found that page helpful.

client/src/pages/Testing/RunTest.tsx
```diff
@@ -2,10 +2,12 @@ import { Button } from '@components/Button';
 import { Container, Content } from './Layout';
 import UploadSvg from '../../assets/upload.svg';
 import { ChangeEvent, useRef } from 'react';
-import { WarningIcon } from '@components/WarningIcon';
 import { useGetEnv } from '../../hooks/useGetEnv';
 import classNames from 'classnames';
 import { ExternalLink } from '@components/ExternalLink';
+import { Warning } from '@components/Diff/Warning';
+
+export const UPLOAD_FILE_THRESHOLD_MB = 5;
```
jakewheeler, line 10: Do we need to export this?

client/src/pages/Testing/RunTest.tsx
```diff
@@ -152,16 +154,9 @@ function UploadZipFile({
...
-          <span>This environment is not approved to handle PHI/PII.</span>
-        </span>
-        <span className="text-center font-bold">
-          Do not upload files that contain PHI/PII.
-        </span>
-      </p>
-    </div>
+    <Warning
+      heading="Do not upload files that contain PHI/PII"
+      message={`This environment is not approved to handle PHI/PII. The maximum file size is ${UPLOAD_FILE_THRESHOLD_MB} MB.`}
```
jakewheeler, line 159: Can we inline this value?
```suggestion
      message="This environment is not approved to handle PHI/PII. The maximum file size is 5 MB."
```

I assume there is more to come regarding this in your other PR? This line is technically inaccurate but I imagine it won't be for long once your other work is merged.

client/src/api/user/user.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Generated by orval v8.5.3 🍺
+ * Generated by orval v8.9.0 🍺
```
jakewheeler, line 2: No need to do anything. Just wanted to mention it has crossed my mind a few times to configure orval to prevent adding this comment. My hesitation is always thinking that it could come in useful one day but in practice it hasn't been useful a single time. If these files were wrong in any way the app would break which would be very obvious.

Something we could look into if we felt it was worth trying to limit some of the noise in PRs.

client/src/pages/Testing/RunTest.tsx
fzhao99: i'm using it in a forthcoming file so added the export here.

client/src/api/user/user.ts
fzhao99: yeah this is a semi-regular annoyance, so I think is worth doing 😓

client/src/pages/Testing/RunTest.tsx
fzhao99: I'm using this constant in another place in a forthcoming change, so just committed it together here. Can change it back until the other change gets in if you want though!

jakewheeler: Nope that's fine! I only wanted to confirm your intention.

jakewheeler (approved):
🚀  thanks!

PR #1183, opened 2026-05-06
refactor: migrate to HeadlessUI Select

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the `Select` components around the app to use our own custom component based off of the [HeadlessUI Select component](https://headlessui.com/react/select).

This one is pretty basic since it's mainly just providing a default style.

## 🧪 How to test

- Take a look at the various `Select` components used in the app and make sure they look & work as expected
- e2e and component tests should be passing


client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -256,24 +258,24 @@ export function ConditionCodeTable({
...
+              value={selectedCodeSystem}
+              onChange={handleCodeSystemSelect}
+            >
+              <option key="all-code-systems" value="all">
+                All code systems
               </option>
-            ))}
-          </Select>
-        </div>
+              {Object.keys(CodeSystem).map((system) => (
```
jakewheeler, line 271: Unrelated - this including `Other` seems a bit odd.

client/src/components/Select/index.tsx
```diff
@@ -0,0 +1,46 @@
+import { Select as HeadlessSelect, SelectProps } from '@headlessui/react';
+import classNames from 'classnames';
+
+interface SelectContainerProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+export function SelectContainer({ children, className }: SelectContainerProps) {
```
jakewheeler, line 9: Although we do use this everywhere currently for consistent styling it's not explicitly required when using `Select`.

client/src/components/Select/index.tsx
```diff
@@ -0,0 +1,46 @@
+import { Select as HeadlessSelect, SelectProps } from '@headlessui/react';
+import classNames from 'classnames';
+
+interface SelectContainerProps {
+  children: React.ReactNode;
+  className?: string;
+}
```
fzhao99, line 7: I think it may be worthwhile to create a base props class with these two props (and maybe others?) representing a base `ComponentProps`. A lot of the other components seem just to be using these two default ones, so it could probably factored together. 

The other specific props could just extend that base class, which looking at a lot of the components, would just be an alias for readability. Could also add things like `aria-label` or other common props to it over time

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
fzhao99: Yeah I think this was a holdover from when we were doing more naive code matching that had more flexibility over code system <> code matching. at this point, think  the rule-based refinement would make "Other" codes not really do much in the refinement, which could be confusing for folks if they have a custom Other code and not have it impact refining.

We can maybe ask Robert what this will do and whether we should remove it

fzhao99 (approved):
one suggestion for the prop declaration on this one, but feel free to either handle it here or in a follow up. thanks!

client/src/components/Select/index.tsx
jakewheeler: Is the idea to DRY things up? I'm not sure I see the benefit in doing this since the components are 100% independent of each other. For example, I don't think a change to the input of `Title` should have any impact on the input of `SelectContainer`. For components that are related I could see this being worthwhile.

PR #1184, opened 2026-05-06
fix(server):  Pretty-print output for UI comparison and download so we don't have the wrong level of indentation

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Formatting was doing a lot and fighting itself at different parts of the ecr pipeline. These changes help us get pretty printed output and smooth over some rough edges with the augmentation metadata being formatted in a weird way and also the narrative `<footnote>` getting written as one long line.

## 🔗 Related Issue

Fixes #1179 
- #1179 

## 🧪 How to test

Unit tests

3. Make sure that both `requirements.txt` and `dev-requirements.txt` are installed
4. Run the unit tests with `pytest -vv tests/` and specifically look at the updates to the format test file

or

Testing the application:

1. Run `docker compose up`
4. Look at the diff viewer and make sure things look right to you; line length is longer but bytes shouldn't be too bad


refiner/app/services/pipeline.py
```diff
@@ -258,11 +259,26 @@ def refine_for_condition(
         refined_rr = etree.tostring(rr_root, encoding="unicode")
 
         trace.refinement_outcome = "refined"
+        # * size metrics are computed against the raw serialization
```
robertmitchellv: this felt like the right change here but let me know if anyone reading this feels different

refiner/app/services/format.py
```diff
@@ -1,42 +1,51 @@
...
     """
 
-    normalized = _normalize_xml(text)
-
-    if preserve_comments:
-        return normalized
-
-    return _strip_comments(normalized)
+    if not isinstance(text, str):
+        raise ValueError(f"Expected XML as str, got {type(text).__name__!r}")
```
jakewheeler: I think I'd prefer to remove this for a couple reasons:
1. It doesn't look like we'll ever hit this exception
2. If `text` for some reason wasn't a string, I'd rather get the exception coming from lxml later in this function

refiner/app/services/pipeline.py
```diff
@@ -258,11 +259,26 @@ def refine_for_condition(
...
 
+        # * pretty-print at the pipeline boundary so every consumer of
+        # RefinementResult receives display-ready output
+        # * lxml's pretty_print=True alone does NOT indent subtrees
+        # added via SubElement after a parse without remove_blank_text
+        # (the section provenance footnotes are the visible symptom);
+        # the formatter does a remove_blank_text reparse + pretty_print
+        # to produce uniformly indented output
+        refined_eicr = format_xml_document_for_display(refined_eicr)
+        refined_rr = format_xml_document_for_display(refined_rr)
```
jakewheeler, line 301: Do we still need to format in the testing routes if we're doing this?

refiner/tests/unit/test_service_format.py
```diff
@@ -4,113 +4,385 @@
...
+        assert "  <parent>" in result
+        assert "    <child>text</child>" in result
 
-    def test_format_xml_document_for_display_or_raise_preserves_comments(self):
-        xml = "<root><!-- comment --><child/></root>"
-        result = format_xml_document_for_display_or_raise(xml, preserve_comments=True)
-        assert "<!--comment-->" in result
+    def test_preserves_element_order(self):
+        result = format_xml_document_for_display("<root><x>1</x><y>2</y></root>")
+        assert result.find("<x>1</x>") < result.find("<y>2</y>")
```
jakewheeler, line 43: 🧠 smart

refiner/tests/unit/test_service_format.py
```diff
@@ -4,113 +4,385 @@
...
+
+    def test_old_private_functions_are_gone(self):
+        """
+        The old _normalize_xml and _strip_comments were removed in the
+        format.py simplification. If they come back, something is
+        wrong.
+        """
+
+        assert not hasattr(format, "_normalize_xml")
+        assert not hasattr(format, "_strip_comments")
```
jakewheeler: I don't know that these are really necessary. Mypy + other checks in place cover tests 1 and 2. For the 3rd one I'm not seeing what the risk is / how these functions would reappear.

refiner/app/services/format.py
robertmitchellv: yeah i think that's a much better way to handle this. i'll add that

refiner/app/services/pipeline.py
robertmitchellv: i don't think we need to actually. i think at this point formatting will have already run once

refiner/tests/unit/test_service_format.py
robertmitchellv: yeah i'm fine removing these and the other test we don't need after removing the guard from the earlier thread. thanks for flagging

refiner/app/services/format.py
robertmitchellv: this is fixed now

refiner/app/api/v1/demo.py
```diff
@@ -107,16 +107,11 @@ async def _build_refined_conditions(
...
-        )
-
         conditions.append(
             Condition(
                 code=condition.code,
                 display_name=condition.display_name,
-                refined_eicr=formatted_refined_eicr,
+                refined_eicr=refined_document.refined_eicr,
                 refined_rr=format_xml_document_for_display_or_raise(
                     original_xml_files.rr
```
jakewheeler: You didn't make the change but I think this is supposed to be `refined_document.refined_rr`? 🤔

jakewheeler: I'm going to fix this in another PR 🫡

jakewheeler: #1189

refiner/app/services/pipeline.py
```diff
@@ -15,14 +15,45 @@
...
 from .ecr.reportability import get_reportable_conditions_by_jurisdiction
+from .format import format_xml_document_for_display
 from .terminology import ProcessedConfiguration
 
+# NOTE:
+# METRICS
+# =============================================================================
+
+
+def _get_size_reduction_percentage(unrefined: str, refined: str) -> int:
```
robertmitchellv, line 32: i noticed that we had some different methodologies in different places that seemed to be doing different things re calculation. seems like what we really want is bytes no matter what; bytes for s3 and for the zip that users download through the webapp. our test data makes it seem like the percentage took a small hit but it is more honest. let me know what you think

refiner/app/services/pipeline.py
robertmitchellv: so we actually do need this for the subelement issue. i've worked on trying to get pipeline to own more and share for the two processes. let me know if you have any issues with this

refiner/app/services/pipeline.py
jakewheeler: I approve 👍  This brings us closer to the way we should be handling things for sure.

refiner/app/services/testing.py
```diff
@@ -335,13 +335,21 @@ async def independent_testing(
         # has a configuration. picking the first entry in an index isn't correct but
         # we should wait to see how the testing service evolves with the routes
 
+        # we don't want to open up the RefinedDocument to a potential None so we need to check that
+        # the default value doesn't persist before packaging it up
+        if (eicr_percentage := pipeline_trace.eicr_size_reduction_percentage) is None:
+            raise ValueError(
+                "eicr_size_reduction_percentage cannot be None - this should always be set"
+            )
+
```
jakewheeler: I suppose this works for now 🤔 Another quick fix I can think of would be to update `RefinementTrace` to use this as the type & default value instead:
```python
eicr_size_reduction_percentage: int = 0
```

And then we wouldn't need this check at all. Does that seem better or worse?

refiner/app/services/testing.py
```diff
@@ -556,6 +564,13 @@ async def inline_testing(
         trace=pipeline_trace,
     )
 
+    # we don't want to open up the RefinedDocument to a potential None so we need to check that
+    # the default value doesn't persist before packaging it up
+    if (eicr_percentage := pipeline_trace.eicr_size_reduction_percentage) is None:
+        raise ValueError(
+            "eicr_size_reduction_percentage cannot be None - this should always be set"
+        )
```
jakewheeler: If we do end up keeping this for now I'd suggest we stick the code in a function and then call it like this:
```python
eicr_size_reduction_percentage=validate_eicr_size_reduction_or_raise(eicr_percentage),
```

refiner/app/services/testing.py
robertmitchellv: we could do this; the only thing is that **if** somehow we end up with an actual `0` value or if it doesn't run and is `0` we won't know the difference

refiner/app/services/testing.py
robertmitchellv: if we still use `None` as the default value instead of `0` i like this approach a lot more! then we can keep the model as `int` only. i can go in this direction if you feel ok with it based on what you suggested in the other thread. lmk

jakewheeler: Let's go with this 👍 the `0` value also isn't really a true fix so I think this is fine for now.

refiner/app/services/testing.py
jakewheeler: Let's go with the other option 🫡

refiner/app/services/testing.py
robertmitchellv: this is fixed now

refiner/app/services/testing.py
```diff
@@ -657,3 +663,25 @@ async def _map_rc_codes_to_conditions(
     # STEP 4:
     # build the final map from the file's code to its corresponding condition objects (all versions)
     return {code: rc_code_to_conditions_map[code] for code in matched_codes}
+
+
+def _require_eicr_size_reduction_or_raise(percent: int | None) -> int:
```
robertmitchellv, line 668: here's the wrapper function @jakewheeler

jakewheeler: Thanks 🫡

jakewheeler (approved):

PR #1189, opened 2026-05-07
fix: refined RR should not be shipped to the client

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

I noticed while reviewing #1184 that `refined_rr` is being shipped to the client but the client doesn't do anything with it. I've removed this from both the inline and independent routes.

## 🧪 How to test

- Both testing flows should still work as expected


refiner/tests/integration/test_ecr_refiner.py
```diff
@@ -76,15 +76,6 @@ async def test_zip_upload_covid_influenza_v1_1(
     for index, condition in enumerate(conditions):
         item_label = f"Condition #{index + 1} (Code:  {condition['code']})"
 
-        # validate RR is well-formed
```
jakewheeler, line 79: I see a very similar test happening in `test_lambda_successful_refinement` so I think we're covered but let me know if you disagree.

refiner/app/api/v1/demo.py
```diff
@@ -117,9 +117,6 @@ async def _build_refined_conditions(
                 code=condition.code,
                 display_name=condition.display_name,
                 refined_eicr=formatted_refined_eicr,
-                refined_rr=format_xml_document_for_display_or_raise(
-                    original_xml_files.rr
```
jakewheeler, line 121: This wasn't the correct data in the first place 😭

fzhao99 (approved):

PR #1191, opened 2026-05-08
fix: configuration creation always uses the latest TES version of a condition

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR addresses a few bugs related to configurations:

- When creating a new configuration draft, the associated conditions (including the primary condition) would always continue to use the same TES version going forward
  - For example, if a config was created with TES version 4.0.0 and a new draft was created, all of the conditions associated with the new draft would still be using 4.0.0 codes even if version 5.0.0 is available
- Code sets using a different TES version than the primary condition of a configuration's TES version could be associated by using the API directly
  - For example, if a config was created using TES version 4.0.0, I could use the API to associate a condition from TES 5.0.0 with that configuration   

## 🔗 Related Issue

Fixes #1192 

## 🧪 How to test

- Take a look at the new integration tests to see upgrading from older TES versions to 5.0.0 working as intended

If you want to test this manually I recommend following these steps:

1. Update your `load_tes_data.py` script's `_build_processed_conditions()` to exclude importing `5.0.0` condition data. This can be done by adding this code:
```python
    for parent in condition_groupers:
        data = ConditionData(parent, valuesets_map)
        if data.payload["version"] == "5.0.0": # <--
            continue # <--
        results.append(
            {
                "condition": data.payload,
                "context_groupers": data.context_grouper_payloads,
            }
        )
```
2. Run `just db refresh` - you should see `Total conditions to upsert: 889`
3. Log into the app and check that 4.0.0 is the latest version (http://localhost:8081/api/v1/info)
4. Create a `Chlamydia trachomatis infection` configuration
5. Add these code sets to it: `Gonorrhea`, `HIV Infection or AIDS`. Note that these two code sets + the config's primary code set are all `not expanded`
6. In your DB, take a look at the configuration's `condition_id` and `included_conditions` array. These IDs should map to the 4.0.0 conditions we're seeing in the UI. Make a note of these somewhere (`select condition_id, included_conditions from configurations`
7. Undo your `load_tes_data.py` change from earlier
8. Run the script to import 5.0.0 condition data -`python scripts/seeding/load_tes_data.py`
9. Activate the `Chlamydia trachomatis infection` configuration
10. Go to the build page and click `Draft a new version`
11. Once your new version is drafted you should see all of the code sets remain the same, however, you'll notice all of these code sets are now marked `fully complete`, meaning the draft is using the 5.0.0 data
12. Run the query again (`select condition_id, included_conditions from configurations`) and you'll see a row for each config. Notice that these configs do not make use of the same condition IDs which is what we expect

## ℹ️ Additional Information

- Is there anything else to consider?


refiner/app/api/v1/info.py
```diff
@@ -36,12 +36,21 @@ async def get_info(
...
+        {
+            "name": config.name,
+            "status": config.status,
+            "version": config.version,
+            "condition_tes_version": condition.version
+            if (
+                condition := await get_condition_by_id_db(id=config.condition_id, db=db)
+            )
+            else "",
+        }
```
jakewheeler, line 48: Figured including the TES version for a config could be useful for debugging. The details array from the `/info` response will look like this
```json
      "detail": [
        {
          "name": "Chlamydia trachomatis infection",
          "status": "active",
          "version": 1,
          "condition_tes_version": "4.0.0"
        },
        {
          "name": "Chlamydia trachomatis infection",
          "status": "draft",
          "version": 2,
          "condition_tes_version": "5.0.0"
        }
      ]
```

refiner/app/api/v1/configurations/codesets.py
```diff
@@ -86,6 +87,21 @@ async def associate_condition_codeset_with_configuration(
...
+    primary_condition = await get_condition_by_id_db(id=config.condition_id, db=db)
+
+    if not primary_condition:
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Could not get configuration's primary condition.",
+        )
+
+    # Associated condition's TES version must match the config's primary condition TES version
+    if primary_condition.version != condition.version:
```
jakewheeler, line 99: This is an additional measure to prevent mixing data from different TES versions.

refiner/tests/integration/test_configurations.py
```diff
@@ -52,6 +54,165 @@ async def test_create_configuration(
...
+        assert response.status_code == status.HTTP_200_OK
+
+        # Get the full config and ensure the app used the expected TES version
+        config_id = response.json()["id"]
+        config = await get_config_by_id(config_id)
+        assert config["condition_id"] == str(
+            await get_condition_id("Glanders", "5.0.0")
+        )
+
+    @pytest.mark.parametrize("OLD_TES_VERSION", ["4.0.0", "3.0.0", "2.0.0"])
```
jakewheeler, line 81: Running this test using a few older versions for good measure.

refiner/tests/integration/conftest.py
```diff
@@ -62,6 +62,8 @@
 TEST_JD_NAME = "Senate District Health Department"
 TEST_JD_STATE_CODE = "GC"
 
+DEFAULT_TES_VERSION = "5.0.0"
```
fzhao99, line 65: would we need to bump this constant / associated tests each time there was a TES update? As I understand it, once we ingest 6.0.0 we'll need to update the constant here and all the associated tests.

Wondering if there's a way set these up to work regardless of the latest / default TES version, but that may be too complicated to set up?

refiner/app/db/conditions/db.py
```diff
@@ -30,6 +31,128 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[str]:
...
+
+    condition_ids = [row["id"] for row in rows]
+
+    return await get_conditions_by_ids(ids=condition_ids, db=db)
+
+
+async def _get_condition_by_canonical_url_and_version_db(
+    canonical_url: str, version: str, db: AsyncDatabaseConnection
+) -> DbCondition:
+    query = """
```
fzhao99: thoughts on the tradeoffs of this function simply wrapping the `_get_conditions` function by passing in a single-item `canonical_url` array / returning the first item of the conditions array? Might be simpler to have a single function with multiple interfaces rather than maintaining multiple implementations

refiner/tests/integration/conftest.py
jakewheeler: I believe with the the way we use this currently it shouldn't matter if we're truly using the most recent version we have available or not. We don't necessarily need to bump it each time but in practice it seems like we usually do (even if not immediately upon adding new TES info).

It might make sense at some point for tests to have access to a dynamically determined `LATEST_TES_VERSION` but maybe we should hold off until we see if it's necessary.

refiner/app/db/conditions/db.py
jakewheeler: That is an excellent idea!

fzhao99 (approved):
thanks for fixing this so quickly!

PR #1193, opened 2026-05-08
fix(server): patch the plan of treatment rules so planned acts get correctly refined

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Add a new entry matching rule for the Plan of Treatment section to cover Planned Act entries--a gap where these were never matched even when their codes were configured. Planned Acts are leaf clinical statements (not wrapper templates like Problems or panels like Results), so the rule uses `preserve_whole_entry=True` and doesn't need pruning logic. The new rule was inserted as rule 4. Separately, the previous rule 4 (now rule 5) had its templateIds corrected: it had been targeting Procedure Activity Procedure (`4.14`, which belongs to the Procedures section) and Planned Act (`4.39`, an `<act>` not a `<procedure>`), so it never actually matched Planned Procedure entries. It now targets Planned Procedure (`4.41`) and the eICR Trigger Code Planned Procedure variant (`15.2.3.42`). Both fixes mean SNOMED codes added as custom codes for Planned Act or Planned Procedure entries will now refine correctly.

## 🔗 Related Issue

Fixes #
- #


refiner/app/services/ecr/specification/entry_match_rules.py
```diff
@@ -53,13 +77,13 @@
     EntryMatchRule(
         code_xpath=(
             ".//hl7:observation"
-            "[hl7:templateId[@root='2.16.840.1.113883.10.20.22.4.4']]"
+            f"[hl7:templateId[@root='{PROBLEM_OBSERVATION_V3}']]"
```
robertmitchellv, line 80: we're using the human readable name instead of the OID; this should hopefully be more clear and also reduce typos

refiner/app/services/ecr/specification/entry_match_rules.py
```diff
@@ -2,6 +2,30 @@
...
+    PROCEDURE_ACTIVITY_ACT_V2,
+    PROCEDURE_ACTIVITY_OBSERVATION_V2,
+    PROCEDURE_ACTIVITY_PROCEDURE_V2,
+    RESULT_OBSERVATION_V3,
+    TRIGGER_CODE_PLANNED_ACT,
+    TRIGGER_CODE_PLANNED_PROCEDURE,
+    TRIGGER_CODE_PROCEDURE_ACTIVITY_ACT,
+    TRIGGER_CODE_PROCEDURE_ACTIVITY_OBSERVATION,
+    TRIGGER_CODE_PROCEDURE_ACTIVITY_PROCEDURE,
+    VITAL_SIGN_OBSERVATION_V2,
```
robertmitchellv, line 27: this is a lot of imports but it seems good to explicitly name and pull in what we need so that we understand the templates we're using. open to other solutions tho

refiner/scripts/oid_catalog_builder/generate_template_oids.py
```diff
@@ -0,0 +1,457 @@
+r"""
+Generate `template_oids.py` from the eICR STU 3.1.1 IG Volume 2 markdown.
```
robertmitchellv, line 2: this probably seems insane but there wasn't a better way to get the human readable names and i probably could've just copy + pasted again but that didn't totally work out for me last time. instead we read the markdown version of the IG and create the script that `entry_match_rules.py` imports

refiner/app/services/ecr/specification/entry_match_rules.py
fzhao99: This is a super welcome change!

refiner/scripts/oid_catalog_builder/generate_template_oids.py
fzhao99: what's the expectation of these getting updated / changing over time? Am a little worried about how maintainable this script / the process for taking the (PDF?) IG documents and turning them into the file of constants. Think this is fine to have if the IG is relatively static, but may want to slowly build this to be more robust in the event it's changing at some regular interval

Also, is there a particular script you're using to turn the PDF's into the Markdown format we're expecting? We probably want to commit that too since we're making regex checks on the content

refiner/scripts/oid_catalog_builder/generate_template_oids.py
```diff
@@ -0,0 +1,457 @@
...
+
+    lines = glue_wrapped_identifier_lines(md_path.read_text().splitlines())
+    templates: list[Template] = []
+
+    for i, line in enumerate(lines):
+        heading = _HEADING_RE.match(line)
+        if not heading:
+            continue
+
+        identifier = _find_next_identifier(lines, i + 1, max_lookahead=20)
```
fzhao99, line 195: how are these lookahead values determined? Are they just the length of the expected pattern for the relevant values or do they represent something else?

refiner/scripts/oid_catalog_builder/generate_template_oids.py
jakewheeler: I have similar thoughts to Bob. The automation is not a bad thing but this PR getting merged will make this a core part of the app.

Like Bob mentioned, the IG being relatively static and the generated file being checked in limits my concern. I agree including more info about how exactly we generate the markdown is a good thing to add.

I don't think it needs to happen in this PR but I think a doc explaining all of the steps required to update to a new IG (and linking to it from the main README) would be a useful addition.

refiner/app/services/ecr/specification/template_oids.py
```diff
@@ -0,0 +1,614 @@
+"""Template OIDs for the eCR Refiner entry matching engine.
+
+Generated from the eICR STU 3.1.1 IG Volume 2 markdown by
+`scripts/generate_template_oids.py`. DO NOT EDIT BY HAND - change the
```
jakewheeler: Any way we could move this to the very top of the file so it stands out a bit more?
```python
GENERATED FILE - DO NOT EDIT BY HAND
GENERATED BY RUNNING: `scripts/oid_catalog_builder/generate_template_oids.sh`
```

Something like this would probably be obvious enough.

refiner/scripts/oid_catalog_builder/generate_template_oids.py
```diff
@@ -0,0 +1,457 @@
...
+         hops. If any reachable parent is a section, group under it.
+      3. If no path leads to a section, group as "Cross-cutting".
+
+    First section reached wins (Contained By order is the IG drafters'
+    primary-container hint).
+    """
+
+    by_name = {t.display_name: t for t in templates}
+    groups: dict[str, list[Template]] = {}
+
```
fzhao99, line 313: do we want to use `defaultdict` here since we're using `set_default` throughout this function?

refiner/scripts/oid_catalog_builder/generate_template_oids.py
robertmitchellv: i can add the command that i used with `marker` to produce the markdown in the readme--that's not a problem. the other option is one i didn't have enough time to explore but i might try to ask Sarah about it is the Trifolia workbench tool--they may have some data set that we can use instead. my thinking is that this is more or less a band aid--what i want is to get some kind of csv or data set with all of these pairs so we can delete this script. and instead generate the file from a dataset

refiner/app/services/ecr/specification/template_oids.py
robertmitchellv: yeah we can do that

refiner/scripts/oid_catalog_builder/generate_template_oids.py
robertmitchellv: i added the command to the readme

refiner/scripts/oid_catalog_builder/generate_template_oids.py
robertmitchellv: yeah we can switch to `defaultdict(list)` — it cleans up the three `setdefault(..., []).append(...)` calls and reads better.

one heads-up: it'll shift the output ordering slightly. right now i explicitly insert the cross-cutting bucket right after the section-2 groups, which controls where that banner lands in the rendered file. with a naive `defaultdict` swap, cross-cutting gets inserted whenever the first entry-3 template with no section path hits it, which is a different position.

not a big deal either way; the ordering is pretty arbitrary. happy to either let it land wherever the IG order puts it, or keep cross-cutting in its current spot with a one-liner that touches the key early. lmk which you'd prefer

refiner/scripts/oid_catalog_builder/generate_template_oids.py
robertmitchellv: they're just "don't scan forever past the heading" values. after a heading like `# 3.41 Problem Observation (V3)`, the identifier code fence is always within a couple lines and the Contained By table is within ~15. i picked 20 and 80 as comfortable upper bounds so a future IG revision that adds an extra paragraph or two doesn't silently break parsing.

the helpers also hard-stop at the next heading, so the line value is really just a backup on top of that.

refiner/app/services/ecr/specification/template_oids.py
robertmitchellv: i folded this in; thanks for the suggestion; i like the way this reads a lot more

refiner/scripts/oid_catalog_builder/generate_template_oids.py
fzhao99: yeah not too hung up on ordering it first, so wherever the IG has it is fine

refiner/app/services/ecr/specification/template_oids.py
robertmitchellv: i had to move things around a bit to make ruff happy. and using the ruff command in the sh file so hopefully this will work a little easier. but it's still moved to the top with a statement above and a space because ruff wants that

jakewheeler: Thank you 🙌

refiner/scripts/oid_catalog_builder/generate_template_oids.py
robertmitchellv: folded this in; it actually didn't end up changing the ordering at all so that works out well

fzhao99 (approved):

PR #1196, opened 2026-05-11
chore: update `configurations.included_conditions` to be of type uuid[]

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR changes the type of `configurations.included_conditions` from `jsonb` to `uuid[]`. There were a few reasons for this change:

- More accurate typing
- Reduce query complexity
  - Can use simpler array functions instead of jsonb functions
  - No need to type cast  
- Eliminate models and replace them with `UUID`
- No mapping required when working with `included_conditions` in code

## 🧪 How to test

Run `just migrate local` and then test that the app functions correctly. `just migrate local down` should swap the DB back to its original state appropriately.

- All automated tests should be passing
- App should function as it did previously


fzhao99 (approved):
💪

PR #1216, opened 2026-05-12
feat: Version Select in Independent Testing Flow

### LINKED ISSUE #987: [USER STORY] Version Select in Independent Testing Flow
## 👤💭 User Story

As a user, I can select what version of a configuration to apply in the independent testing flow, so that I can see the results for that version

## 🏗️ Acceptance Criteria

- [ ] Ability to select version for each condition to apply in refinement
- [ ] Remove checks/errors for conditions needing to be active, but keep errors for configurations needing to be created (i.e. can't test a condition that doesn't even have a draft config)

## 🎨 Design Notes
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=9351-48828&t=T7WfiQLMRfFQqrTN-1

## Additional Context
It will be impossible for JDs to test representative outputs for their files without activating conditions they may not want to

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR modifies the way that the independent test flow works so that configurations no longer need to be `active` in order to be used in testing. Configurations that are `draft`, `inactive`, and `active` can all be used for testing.

Additionally, users now have a choice of which configuration to use for testing on a per reportable condition basis. If an `active` configuration exists for a reportable condition, that will be the default selection. Users can uncheck the checkbox next to the condition name to not use that condition's configuration for refining at all.

At least one condition must be selected. If no conditions are selected then the `Refine eCR` button will be disabled until the user selects one or more conditions.

### Shadow RR

The shadow RR should be produced during the independent test flow when:
1. There are reportable conditions that don't have an associated config
2. There are unused conditions (unchecked in the UI)
3. Both 1 and 2 at the same time

## 🔗 Related Issue

Fixes #987 

## 📺 Demo

Kapture 2026-05-15 at 10 28 26

## 🧪 How to test

Manual:

1. Run `just db refresh` to start with a clean database
4. Navigate to the `Testing` tab and use the local file option
5. You'll see a warning that no configurations are available (this isn't new)
6. Set up a COVID-19 and/or Influenza configuration. Go through the flow again and you'll see that you can now make selections
7. Try out the various options and then hit the `Refine eCR` button. Refinement should happen according to your selected options
8. Ensure the zip file package looks correct according to your choices

Let me know if you have suggestions for other automated tests.


client/src/pages/Testing/index.tsx
```diff
@@ -20,16 +20,27 @@ export function Testing() {
...
+    resetConfigurations,
+    errorMessage: configurationsErrorMessage,
+  } = useDiscoveredConfigurations();
+
+  const {
+    refinementResponse,
+    runRefinement,
+    resetRefinement,
+    errorMessage: refinementErrorMessage,
+  } = useRunRefinement();
```
fzhao99, line 34: nice, these new hooks are super helpful

refiner/tests/integration/test_demo.py
```diff
@@ -44,10 +77,17 @@ async def test_demo_upload_smoke(
         )
     assert response.status_code == status.HTTP_200_OK
     data = response.json()
-    assert "refined_conditions" in data
-    assert "conditions_without_matching_configs" in data
-    assert "unrefined_eicr" in data
-    assert "refined_download_key" in data
+
+    expected_keys = {
```
fzhao99, line 81: 💪

refiner/tests/integration/test_demo.py
```diff
@@ -79,3 +120,87 @@ async def test_demo_upload_smoke(
...
+    covid_config = await create_config(covid_condition_id)
+
+    # draft config
+    flu_condition_id = await get_condition_id("Influenza")
+
+    uploaded_file = covid_influenza_v1_1_zip_path
+
+    # helper function to get the actual ID
+    def resolve(fixture):
+        return [str(flu_condition_id)] if fixture == "flu" else fixture
```
fzhao99:
```suggestion
        return [str(flu_condition_id)] if fixture == "flu" else []
```

was a little hard for me to grok that this helper was returning the empty list  in the "unconfigured / unused" cases. Think it's more readable to just return the empty list explicitly

refiner/app/services/testing.py
```diff
@@ -329,57 +255,42 @@ async def independent_testing(
...
+        )
 
     return IndependentTestingResult(
-        original_eicr_doc_id=first_original_eicr_doc_id
-        if first_original_eicr_doc_id
-        else "",
-        refined_documents=refined_documents,
-        no_matching_configuration_for_conditions=no_matching_configurations,
-        no_active_configuration_for_conditions=no_active_configurations,
+        original_eicr_doc_id=first_original_eicr_doc_id,
```
fzhao99, line 292: glad we could clean up these function calls

refiner/app/services/testing.py
```diff
@@ -125,198 +64,185 @@ class InlineTestingTrace:
...
+    Model to represent individual discovered configurations.
+    """
+
+    id: UUID
+    version: int
+    status: DbConfigurationStatus
+
+
+@dataclass
+class DiscoveredConfigurationGroup:
```
fzhao99: Naming suggestion: any chance we could use something like `DiscoveredConfigurationHistory` or `DiscoveredConfigurationSet` instead of using group here? Think my brain is having a namespace collision against "condition groupers" that's making some of the related code here hard to understand at a glance 😅

client/src/pages/Testing/ReportableConditionsResults.tsx
```diff
@@ -1,30 +1,85 @@
...
+  const [groupSelections, setGroupSelections] = useState<
+    Map<string, { checked: boolean; selectedId: string }>
+  >(
+    () =>
+      new Map(
+        matchedGroupsWithConfig.map((cg) => [
+          cg.name,
+          {
+            checked: true,
+            // try to set "active" as the default, otherwise fall back to whatever is first
```
fzhao99:
```suggestion
            // try to set "active" as the default, otherwise fall back to whatever is first since the backend reverse sorts the response based on version number
```

refiner/app/services/testing.py
jakewheeler: That's true 💀 I like `Set`. I'll make the switch.

fzhao99 (approved):

refiner/app/services/testing.py
fzhao99: appreciate you changing this despite the annoyance!

PR #1218, opened 2026-05-13
fix: error should display MB instead of bytes

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

I noticed that this error message was displaying incorrectly while doing some manual testing.

Before:
image

After:
image

## 🧪 How to test

- Try uploading a file in the independent test flow that is larger than the max uncompressed size


fzhao99 (approved):
😅

PR #1219, opened 2026-05-13
fix(server): add augmentation to remainder RR

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This minimally adds the augmentation of the remainder RR (FKA "shaddow RR" 🙏🏻 ) and brings it's construction closer in line with the pipeline service. There was also some refactoring to try and move away from the `AugmentationContext` to the `AugmentationRun` shape. There is still a great deal of work that should be done to get both the `testing.py` and `lambda_function.py` implementations of `pipeline.py` to be better aligned and there are a handful of TODOs that have been added to call those out in different places. We can use this PR to shape the discussion of what that should look like.

## 🔗 Related Issue

Fixes #1190 
- #1190 

## 🧪 How to test

Unit tests

3. Make sure that both `requirements.txt` and `dev-requirements.txt` are installed
4. Run the unit tests with `pytest -vv tests/unit` (specifically pipeline and augmentation)
5. Additionally you can run the test in `refiner/app/lambda`

or

Testing the application:

1. Run `docker compose up`
4. You should be able to add just one configuration (covid of flu) and then run the independent testing
5. Download the zip from the app and then check the remainder RR


refiner/app/lambda/lambda_function.py
```diff
@@ -445,6 +457,14 @@ def run_refinement(input: RefinementInput) -> RefinementOutput:
         operation="discovered_reportability",
     )
 
+    # TODO:
+    # * planning-object: the planning object should own this
+    # one AugmentationRun per input pair, shared across every
+    # refine_for_condition and produce_remainder_rr_for_jurisdiction
+    # call in this invocation so all augmented outputs of the session
+    # share an effectiveTime
+    run = create_augmentation_run_from_xml_files(input.xml_files)
```
fzhao99, line 466: maybe this will be clear once the planning object gets implemented, but is there a functional difference for how the run is going to be instantiated (different inputs, inputs from a different place, etc.), or is it just going to get moved / abstracted into a different place?

refiner/app/api/v1/demo.py
```diff
@@ -211,11 +211,13 @@ async def demo_upload(
...
 
-    # Add shadow RR (for inactive conditions) to zip
-    if test_results.shadow_rr:
+    # add remainder RR: carries reportability for conditions that did not have an active
+    # configuration for the jurisdiction; see pipeline.produce_remainder_rr_for_jurisdiction
+    if test_results.remainder_rr is not None:
         zip_package.add(
             ZipFileItem(
-                file_name="CDA_RR_unrefined_rr.xml", file_content=test_results.shadow_rr
+                file_name="CDA_RR_remainder.xml",
```
jakewheeler: I don't think we should do this. It no longer emulates the lambda naming which is what we were going for.

fzhao99 (approved):
one question about future work but otherwise lgtm!

refiner/app/services/testing.py
```diff
@@ -79,14 +82,19 @@ class IndependentTestingResult:
...
+           skipped conditions so each condition's reportability
+           appears exactly once across the full output package.
     """
 
     original_eicr_doc_id: str
     refined_documents: list[RefinedDocument]
     no_matching_configuration_for_conditions: list[NoMatchEntry]
     no_active_configuration_for_conditions: list[NoMatchEntry]
-    shadow_rr: str | None
+    remainder_rr: RemainderRRResult | None
```
jakewheeler: I think this should remain as `str | None`. We don't actually use other `RemainderRRResult` properties in the test flow.

refiner/app/services/pipeline.py
```diff
@@ -54,6 +69,58 @@ def _get_size_reduction_percentage(unrefined: str, refined: str) -> int:
...
 
 
+# NOTE:
+# SESSION CONSTRUCTION
+# =============================================================================
+
+
+def create_augmentation_run_from_xml_files(
+    xml_files: XMLFiles,
+    augmentation_time: str | None = None,
```
jakewheeler: Can we remove this since it's never used?

refiner/app/services/pipeline.py
```diff
@@ -54,6 +69,58 @@ def _get_size_reduction_percentage(unrefined: str, refined: str) -> int:
...
+        eicr_root = xml_files.parse_eicr()
+    except etree.XMLSyntaxError as e:
+        raise XMLValidationError(
+            message="Failed to parse eICR document",
+            details={"error": str(e)},
+        )
+
+    return create_augmentation_run(
+        eicr_root=eicr_root,
+        augmentation_time=augmentation_time,
```
jakewheeler: Actually same for this function. `augmentation_time` is never passed in.

refiner/app/services/ecr/augment.py
```diff
@@ -71,14 +78,24 @@
...
+# * occupies the same seed slot as the condition grouper UUID for the
+# per-condition pair case; the literal value distinguishes the two
+# families of output within a jurisdiction. cannot collide with any
+# real grouper UUID because UUIDs have a fixed 36-character hyphenated
+# shape that the literal does not satisfy
+REMAINDER_SCOPE: Final[str] = "remainder"
 
-def _extract_uuid_from_canonical_url(canonical_url: str) -> str:
+
+def extract_uuid_from_canonical_url(canonical_url: str) -> str:
```
jakewheeler: This function is basically a duplicate of `_extract_uuid_from_canonical_url()` in `aws/s3_keys.py`.

refiner/app/services/ecr/augment.py
```diff
@@ -115,21 +132,23 @@ def _derive_augmented_eicr_id(
 def _derive_augmented_rr_id(
     original_rr_id_root: str,
     jurisdiction_id: str,
-    condition_grouper_uuid: str,
+    scope: str,
```
jakewheeler: Can we narrow this type? It looks like the acceptable values are only `scope: UUID | Literal["remainder"]`?

refiner/app/lambda/lambda_function.py
robertmitchellv: mostly just moved. the inputs stay the same (eICR + RR), and the cross output timestamp sharing behavior is already in effect-; both callers construct the run at the session boundary today and thread it through. what changes with the planning object is who owns the construction: the plan builder runs first, captures the run as part of the plan, and consumer code receives the plan instead of constructing pieces of it. functionally equivalent, structurally a cleaner seam. the TODO is calling that out so the reason for the deferral is on the record

refiner/app/api/v1/demo.py
robertmitchellv: i can change this back but would like a todo about what we should name it to once we have a PR for that maybe? what do you think?

refiner/app/services/ecr/augment.py
robertmitchellv: do you think a function that we stick in `core/` in a `utils.py` script would be appropriate? or should i just import the one from `s3_keys.py`? the only difference (which is connected to another question you raised) is that the one in `s3_keys.py` returns a `UUID` instead of `str`. because the augmentation functions want to use an `str` in creating the uuidv5 i didn't want to return a UUID because i'd have to convert it to `str` anyway. maybe there's a good reason to do this from a validation pov? i knew there were two but i kept the one that returns a `str` for augmentation service. i'm open to using a more strict function just let me know what you think

refiner/app/services/pipeline.py
robertmitchellv: this and the comment about `create_augmentation_run_from_xml_files` are sort of connected. i agree that we can remove them for now. if the planning object ends up wanting to inject a captured timestamp we can adjust as needed. was kind of getting ahead of myself because i started on the planning object and then reeled myself back in 🎣

refiner/app/services/ecr/augment.py
robertmitchellv: if we decide to go with the stricter function from `aws_keys.py` then i think this makes sense

refiner/app/api/v1/demo.py
robertmitchellv: this is wrapped up now

refiner/app/services/pipeline.py
robertmitchellv: this is wrapped up now

refiner/app/services/ecr/augment.py
robertmitchellv: we talked offline about importing the s3_keys function as is; and i added a todo about potentially moving it to core or doing something else

refiner/app/services/ecr/augment.py
robertmitchellv: this is done now

refiner/app/services/testing.py
robertmitchellv: it's getting returned in pipeline; do you still think it should be cut?

refiner/app/services/pipeline.py
robertmitchellv: this is folded in now

refiner/app/services/testing.py
jakewheeler: Yes. I think it makes sense to only return the text (if it exists) since that's all we expect to use in the independent test route.

(in demo file) current:
```python
    if test_results.remainder_rr is not None:
        zip_package.add(
            ZipFileItem(
                file_name="CDA_RR_unrefined_rr.xml",
                file_content=test_results.remainder_rr.remainder_rr,
            )
        )
```

(in demo file) my suggestion
```python
    if test_results.remainder_rr is not None:
        zip_package.add(
            ZipFileItem(
                file_name="CDA_RR_unrefined_rr.xml",
                file_content=test_results.remainder_rr, # <-- just the text
            )
        )
```

refiner/app/services/pipeline.py
```diff
@@ -5,13 +5,26 @@
...
+# TODO:
+# **core-util** _extract_uuid_from_canonical_url currently lives in
+# aws/s3_keys.py and is imported across package boundaries here
+# * the augment.py-side duplicate was removed in favor of this single
+# implementation, which validates by constructing a UUID (a malformed
+# canonical_url fails fast at construction)
+# * follow-up ticket: relocate this to a shared core util and give it
+# a public name so neither aws/s3_keys.py nor this module owns a cross-package
+# private
+from .aws.s3_keys import _extract_uuid_from_canonical_url
```
jakewheeler, line 18: thank you 👍

refiner/app/services/testing.py
robertmitchellv: ahhhh ok i get you now. yeah that sounds good to me

robertmitchellv: if you want to take a quick peek this is wrapped now

PR #1223, opened 2026-05-15
chore: refactor app banner business logic

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Some backend cleanup for the app update banner effort that uses better idioms / fallback methods for the mapping layer

## 🧪 How to test
- Run a db refresh using `just db refresh`
- Notice app update banner renders. Click either the dismissal or the "view updates" button to take you to the app updates page
- Navigate back to the homepage and notice the banner disappears and the date stamp for when you actioned the notification gets written to the database


refiner/app/api/auth/handlers.py
```diff
@@ -223,14 +225,38 @@ class NotificationInfo(BaseModel):
...
+    Notification information needed to render frontend banners.
     """
 
     most_recent_app_update: NotificationInfo
 
 
+def _map_to_aware_dt(val: str | datetime) -> datetime:
+    """Ensures value is a datetime, mapping to UTC timezone."""
+    dt = datetime.fromisoformat(val) if isinstance(val, str) else val
+    return dt.astimezone(UTC)
```
fzhao99: the GitHub API returns timezone-aware objects, so layering in UTC defaults / conversions to compare apples to apples

refiner/app/api/auth/handlers.py
```diff
@@ -223,14 +225,38 @@ class NotificationInfo(BaseModel):
     date_acknowledged: str | None = None
```
jakewheeler: Any issue with excluding `date_acknowledged` in the client response? Doesn't look like we use it at all.

refiner/app/api/auth/handlers.py
```diff
@@ -223,14 +225,38 @@ class NotificationInfo(BaseModel):
...
+class NotificationResponse:
     """
-    User notification state.
+    Notification information needed to render frontend banners.
     """
 
     most_recent_app_update: NotificationInfo
 
 
+def _map_to_aware_dt(val: str | datetime) -> datetime:
```
jakewheeler: Could this ever be `datetime`?

refiner/app/api/v1/app_notifications.py
```diff
@@ -40,4 +40,4 @@ async def update_user_notifications(
         db=db,
     )
 
-    return build_user_response(updated_user)
+    return UserResponse.from_db_user(updated_user)
```
jakewheeler, line 45: We don't necessarily need to tackle this in this PR but I don't think it makes sense for the client to send `name` and `date_acknowledged` to the server. There's no validation around either of these so they can be any values.

If I were going to fix this, I'd probably hardcode the name server-side for now (since we only have 1 possibility) and then let the server determine the `date_acknowledged` value.

fzhao99: I think I figured something out to do what you're suggesting: used our old Orval-generates-enum trick to ship a package of fixed values for the frontend to select from, and then had the backend handle the date acknowledgement as suggested. Think this is a good compromise between extensibility and validation, since we can just extend the enum with forthcoming notification needs

Lmk if what I did makes sense

refiner/app/api/auth/handlers.py
fzhao99: the thought was to use this for the default `datetime.min` values as well for when the releases API returns nothing, as well as when the user object has nothing. 

I folded that into both checks now so this helper method's typing is more helpful.

refiner/app/api/v1/app_notifications.py
```diff
@@ -33,11 +34,12 @@ async def update_user_notifications(
     """
     Updates notification acknowledgement state for the current user.
     """
+
     updated_user = await update_user_notifications_db(
         user_id=user.id,
-        name=request.name,
-        date_acknowledged=request.date_acknowledged,
+        name=request.key,
+        date_acknowledged=datetime.now(UTC).isoformat(),
```
jakewheeler, line 41: 👍

refiner/app/api/auth/handlers.py
```diff
@@ -214,21 +216,29 @@ async def auth_callback(
...
+
+class NotificationKeys(StrEnum):
     """
-    Information for a single notification.
+    Enum class to type the values of the notifications possible for actioning on the frontend.
     """
 
-    should_show: bool
-    date_acknowledged: str | None = None
+    MOST_RECENT_APP_UPDATE = "most_recent_app_update"
```
jakewheeler, line 232: This should work! Looks like FastAPI/Pydantic will automatically sure the request uses a value defined here.

refiner/app/api/v1/app_notifications.py
jakewheeler: Yep I like it 🎉

client/src/pages/Configurations/index.tsx
```diff
@@ -125,23 +129,22 @@ export function Configurations({ user, setUser }: ConfigurationsProps) {
...
+  if (!should_show) {
     return null;
   }
 
   async function dismissNotification() {
     try {
       const resp = await updateUserNotifications({
-        name: 'most_recent_app_update',
-        date_acknowledged: new Date().toISOString(),
+        key: NotificationKeys.most_recent_app_update,
```
jakewheeler: Since we're looking at it, would you mind swapping this to use `mutateAsync` from the hook instead?
```ts
const { mutateAsync } = useUpdateUserNotifications();
...

  async function dismissNotification() {
    try {
      const resp = await mutateAsync({
        data: {
          key: NotificationKeys.most_recent_app_update,
        },
      });
      setUser(resp.data);
    } catch (error) {
      console.error('Failed to update user notifications', error);
    }
  }
```

client/src/pages/Configurations/index.tsx
```diff
@@ -125,23 +129,22 @@ export function Configurations({ user, setUser }: ConfigurationsProps) {
 }
 
 function AppUpdateBanner({
-  appUpdateInfo,
+  should_show,
   setUser,
 }: {
-  appUpdateInfo: NotificationInfo;
+  should_show: boolean;
```
jakewheeler: Maybe we rename this to be `isVisible`?

client/e2e/db/index.ts
```diff
@@ -16,3 +16,7 @@ export const db = new Pool({
 export async function deleteAllConfigurations(): Promise<void> {
   await db.query('DELETE FROM configurations');
 }
+
+export async function resetUserNotificationState(): Promise<void> {
```
jakewheeler, line 20: Whoops I did something similar in my PR since I was experiencing some flakiness and needed the tests to pass 😆 We can get this merged first and I'll handle any conflicts in mine.

jakewheeler (approved):
Thank you 🫡

PR #1224, opened 2026-05-15
refactor: update useLogin to use react query hook

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

> [!NOTE]
> Let's merge #1223 before we merge this one.

This PR primarily refactors the `useLogin` hook used at the top-level of the app to check if there is an active user session. This refactor makes use of the generated hook `useGetUser` instead of handling all of the data and states manually.

Using the generated hook also allows us to make use of query invalidation. `UseLogin` now also offers `refreshUser` as part of its API, which can be called to invalidate the user and perform a refetch.

## 🧪 How to test

- App should work as it did before


fzhao99 (approved):
thanks!

PR #1242, opened 2026-05-19
fix: independent testing not matching on some configurations

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the configuration discovery service in the independent test flow to find configurations for reportable conditions regardless of their primary condition's TES version.

This was happening because the matching was overly strict (see my comment).

Fixes #1239 

## 🧪 How to test

- Take a look at the new automated test

If you'd like to see this manually, the easiest way is this:
1. Go to the `testing.spec.ts` e2e test
2. Comment out line 10 (config deletion)
3. Replace `test` with `test.only` on line 45 (`test.only('Should be able...)`)
4. Run the test, you should have 21 COVID configs in your local environment
5. In `app/db/conditions/db.py` replace line 122 with this: `latest_version = '3.0.0'`
6. Create new COVID draft(s) in the UI. Any new drafts you create will use the TES 3.0.0 COVID-19 condition

At this point you have a mix of TES 3.0.0 and 5.0.0 COVID-19 configurations. You should see all of these configurations appear in the independent test flow dropdown when running the test file.

## ℹ️ Additional Information

- Any other testing to add?

### DIFF
--- a/refiner/app/services/testing.py
+++ b/refiner/app/services/testing.py
@@ -142,11 +142,11 @@ async def discover_configurations_for_conditions(
         jurisdiction_id=jurisdiction_id, db=db
     )
 
-    conditions_grouped_by_url = _group_conditions_by_url(rc_to_conditions_map)
-
     configured_primary_condition_ids = {
         config.condition_id for config in all_jurisdiction_configs
     }
+
+    conditions_grouped_by_url = _group_conditions_by_url(rc_to_conditions_map)
 
     condition_sets: list[DiscoveredConfigurationSet] = []
     for all_versions in conditions_grouped_by_url.values():
--- a/refiner/app/services/testing.py
+++ b/refiner/app/services/testing.py
@@ -152,7 +152,7 @@ async def discover_configurations_for_conditions(
     for all_versions in conditions_grouped_by_url.values():
         all_condition_ids = {c.id for c in all_versions}
 
-        representative_condition = next(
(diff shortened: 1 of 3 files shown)


refiner/tests/localstack/seed.py
```diff
@@ -41,11 +41,17 @@ def seed_localstack(s3_client):
     """
     BUCKET = "local-config-bucket"
 
+    response = s3_client.list_objects_v2(Bucket=BUCKET)
```
jakewheeler, line 44: The new test is working with COVID-19 and Influenza which we use all over the place, including for Lambda testing. I updated this script so that the bucket is always cleared out upon seeding which will prevent anything unexpected from ending up here. This will allow integration tests to activate configs without impacting Lambda testing.

refiner/app/services/testing.py
```diff
@@ -159,16 +161,16 @@ async def discover_configurations_for_conditions(
             [
                 c
                 for c in all_jurisdiction_configs
-                if c.condition_id == representative_condition.id
+                if c.condition_id in all_condition_ids
```
jakewheeler, line 172: This was the main issue. If `representative_condition` was version `5.0.0` but the configuration uses a `condition_id` of the version `4.0.0` condition no match would happen, for example.

fzhao99 (approved):
great find!

PR #1243, opened 2026-05-19
feat(api,ui): Add Narrative Only awareness

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

While this PR attempts to make sections aware of Narrative Only sections such as
`10154-3`, `29299-5`, `10164-2`, and `10187-3`. There are some UI changes in the
provided design image on #1182 that don't coincide with the current designs of
these sections.

1. 💬 feat(api,ui): Add Narrative Only awareness 🏷️ 7cb811d1
1. 💬 fix: Improve matching just to a single row 🏷️ e3e96d31
1. 💬 chore: Format file(s) 🏷️ 68ffad6b
1. 💬 test(api): Verifies "refine" action preservation 🏷️ ef21fd4f
1. 💬 fix: Refactor refine to match test cases 🏷️ 7e65c071

## 🔗 Related Issue

- Closes #1182
- Closes #1188

## ✅ Acceptance Criteria

- [x] Sections without coded data (narrative only) do not have a coded data/data
      handling approach toggle and say "Not applicable for this section" instead
- [x] Include/exclude checkbox remains unchanged and toggle or dropdown for
      narrative is limited to keep/remove options
- [x] Need to be able to handle these sections appropriately in refining algorithm
- [x] Affected sections can be found in Robert's doc on the topic (Chief Complaint,
      Reason for Visit, History of Present Illness, Review of Systems)

## 🧪 How to test

1. Run the server with `just dev up` and reset the database if necessary with
   `just db refresh`.
1. Navigate to a Configuration's Section page under `/build` and verify that
   four sections now have a _Data handling approach_ of _Not applicable for this
   section_ rather than a switch to _Refine & optimize_.
1. Test the API response by attempting to toggle the _Narrative_ off to reveal
   an API response of `400 Bad Request` with the message similar to _"The section
   Section Name is narrative-only and cannot be refined."_
1. Run server tests `just server test`
1. Run e2e tests `just client run e2e`


client/src/pages/Configurations/ConfigBuild/Sections/index.tsx
```diff
@@ -28,6 +28,12 @@ import { Tooltip } from '@components/Tooltip';
  */
 
 const disabledSections = new Set(['88085-6', '83910-0']);
+const narrativeOnlySections = new Set([
+  '10154-3', // Chief Complaint
```
fzhao99: We should grab these values on the server and have Orval ship them to the frontend using something like what we're doing in [this](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1165/changes#diff-26fb372745b95513a0bb663d39d59d7d930b45e6194cad51e0b5fef817eacf5cR24) [PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1165/changes#diff-fda83d133d1f6cb2361f78a0366c983dba090d6514dd6388e0cee8a923705b9eR25-R31).

refiner/app/services/ecr/refine.py
```diff
@@ -291,12 +291,23 @@ def create_eicr_refinement_plan(
...
+
+        # NORMALIZATION: Narrative-only sections cannot be refined.
+        # Treat "refine" as "retain" to prevent accidental stubbing.
+        spec_entry = specification.sections.get(code)
+        if (
+            spec_entry
+            and not spec_entry.has_match_rules
+            and instructions.action == "refine"
+        ):
+            instructions = dataclasses.replace(instructions, action="retain")
```
fzhao99: Could you say a bit more about why we're replacing refine with retain here rather than throwing like we do in the other file? Unsure what the context for "accidental stubbing" here is, but a little worried that if somehow we get to this point we're doing a silent override

refiner/app/services/ecr/refine.py
```diff
@@ -329,6 +340,7 @@ def create_eicr_refinement_plan(
 def _interpret_run_result(
     section_rules: DbConfigurationSectionInstructions,
```
fzhao99: it looks like we can remove this parameter because it isn't getting used

refiner/app/services/ecr/refine.py
```diff
@@ -362,6 +374,9 @@ def _interpret_run_result(
             the outcome — but reserved for future policy variations
             that need to consult the configuration.
         run_result: What the matching engine reported about the run.
+        has_match_rules: Whether it's not a narrative-only section or is.
```
fzhao99:
```suggestion
        has_match_rules: Whether it's not a narrative-only section.
```

refiner/app/services/ecr/refine.py
```diff
@@ -370,7 +385,7 @@ def _interpret_run_result(
     # policy override: no matches always produces a stub, regardless
     # of what the narrative configuration said. see the docstring
     # above for the rationale.
-    if not run_result.matches_found:
+    if has_match_rules and not run_result.matches_found:
```
fzhao99: I think we may need to rework the outcome of this function, since in cases of narrative-only sections, I'd assume we'd want to return some variant of `NARRATIVE_RETAINED` or `NARRATIVE_REMOVED` that don't currently exist rather than the current options that suggest something's changed in the coded data. 

Would want to double check with Robert on this, but we might want to change `has_match_rules` to something like `is_narrative_only` and do some conditional gating accordingly

refiner/app/services/ecr/refine.py
rogeruiz: Sure. From what I understand from `refiner/DIBBs-eICR-Narrative-Only-Sections.md` is that narrative-only sections are designed to have no match rules and are processed on their narrative. So the accidental stubbing here is about how the refiner works where a section is configured for refine but wit the matching engine finding no matching entries the section is stubbed regardless of the narrative setting. By replacing the `refine` with `retain` for these specific sections during plan creation, it bypasses the refinement logic all together. This means that these sections will be handled by the retain logic which preserves the narrative or removes it if selected without risking the silent override/stubbing by the matching policy.

Specifically this section stood out to me to create this new designation for narrative-only sections.

> The "refine" action depends on entry-level code matching (EntryMatchRule evaluation) to selectively retain condition-relevant entries. For sections with no IG-defined entries, there is nothing to match against — the only meaningful content is the narrative block, which is indivisible from the refiner's perspective. Offering "refine" for these sections would be misleading: it implies entry-level filtering that cannot occur.

I think I'd want @robertmitchellv to double-check my understanding here when he's got the time.

client/src/api/schemas/sectionMetadata.ts
```diff
@@ -0,0 +1,21 @@
+
+/**
+ * Utility class to help Orval ship these values to the frontend.
+
+These sets of LOINC codes drive UI behavior in the eICR Sections
+table and are sourced from refiner policy + the eICR specification.
+Shipping them as literal-typed defaults means the frontend imports
+real values rather than re-declaring them.
+ */
+export interface SectionMetadata {
```
fzhao99: Hmm, think we want to try to avoid having to redeclare these values on the frontend. Think instead of this, it's better / more readable to use enums and then include those in a section response to have Orval ship them to the frontend. I put together a sketch of what I meant here: https://github.com/CDCgov/dibbs-ecr-refiner/pull/1274

Feel free to approve / merge if you think what I'm suggesting makes sense

rogeruiz: Yeah let me take a look. Thanks @fzhao99

client/shared/openapi.json
fzhao99, line 1: we generally try to keep this file minified / on the one line, so you might want to have Orval regen this and then save without formatting

refiner/app/services/ecr/policy.py
```diff
@@ -12,23 +12,53 @@
...
+    every catalog section with has_match_rules=False is listed here).
+    """
+
+    CHIEF_COMPLAINT = "10154-3"
+    REASON_FOR_VISIT = "29299-5"
+    HISTORY_OF_PRESENT_ILLNESS = "10164-2"
+    REVIEW_OF_SYSTEMS = "10187-3"
+
+
+NARRATIVE_ONLY_SECTIONS = [section.value for section in NarrativeOnlySection]
```
fzhao99, line 80: noticed in my rework that we aren't using this anywhere. Is that intentional?

refiner/app/services/ecr/policy.py
```diff
@@ -12,23 +12,53 @@
...
+    """
+
+    EMERGENCY_OUTBREAK = "83910-0"
+    REPORTABILITY_RESPONSE = "88085-6"
+
+
+SECTION_PROCESSING_SKIP = [section.value for section in DisabledSection]
+
+
+class NarrativeOnlySection(StrEnum):
```
robertmitchellv, line 60: this is great; it really explains _why_ refine isn't offered (nothing to match against), points to where the normalization actually lives (`refine.py`), and commits to a sync test guarding against catalog/enum drift 🫡

refiner/app/services/ecr/policy.py
```diff
@@ -12,23 +12,53 @@
...
+    untouched in the refined output. In the future we may decide to implement new ways to handle these sections
+    but for now skipping them is easier and produces valid (Schematron-valid)
+    output. The Literal-typed tuple alias below is the single source of truth: it is
+    used both at runtime (to derive SECTION_PROCESSING_SKIP) and at the API
+    boundary (so Orval ships the concrete LOINC codes to the frontend as
+    const values rather than as plain `string[]`).
+    """
+
+    EMERGENCY_OUTBREAK = "83910-0"
+    REPORTABILITY_RESPONSE = "88085-6"
```
robertmitchellv, line 54: after reading what i initially wrote and what you've added to the docstring (which i think is a much better💡 imo), i think it would be better to be more precise about these two sections; if you want to use this or rework it in any way feel free:

```suggestion
class DisabledSection(StrEnum):
    """
    These sections are preserved intact in every refined document.
    
    They sit outside the normal refinement workflow not because the
    refiner needs to protect them, but because we don't yet have a
    real-world authoring contract to refine against:

    - Reportability Response Information (88085-6) is defined by the
      eICR STU 3.1.1 IG as a section the PHA populates after eICR
      receipt, as part of internal data integration -- not something
      the HCO authors at eICR generation time. It's sketched in the
      IG but is not part of eICRs flowing through AIMS today.

    - Emergency Outbreak Information (83910-0) has a defined section
      template and a deliberately generic Observation structure
      ("unknown until the time of the outbreak," per the IG). There
      is no settled EHR implementation pattern, and the next outbreak
      will likely shape how this section actually appears in
      production.

    Skipping is the easier path for now and produces Schematron-valid
    output. We can revisit -- per section, independently -- if and
    when either becomes something we actually see in real documents.

    The Enum is the single source of truth: used at runtime to derive
    SECTION_PROCESSING_SKIP, and at the API boundary so Orval ships
    the concrete LOINC codes to the frontend as const values rather
    than as plain `string[]`.
    """

    EMERGENCY_OUTBREAK = "83910-0"
    REPORTABILITY_RESPONSE = "88085-6"
```

refiner/app/services/ecr/refine.py
robertmitchellv: your read is right roger but i think we're patching this in maybe the wrong spot. ideally we'd want the catalog to drive section defaults--it does this via `get_default_sections` in `services/configurations.py`:

```
action="retain" if loinc_code in SECTION_PROCESSING_SKIP else "refine"
```

since we were just setting defaults for all sections it's kind of dumb on purpose. i think that the ideal situation tho would be that we send the right data ahead of time to the configuration so you don't have to patch anything or check for things in `refine.py`--it'll already be in the configuration. maybe something like:

```
action=(
    "retain"
    if loinc_code in SECTION_PROCESSING_SKIP or not section_spec.has_match_rules
    else "refine"
)
```

there's still some space for things to bite us tho in `clone_section_processing_instructions` where after this PR lands it could clone the wrong instructions and things could blow up so this will probably need to be adjusted. hopefully this isn't too painful a suggestion.

i would say that essentially for narrative only sections we should really only offer "include" as a true/false and that's it since i'm not sure what "don't include section" vs "include section but don't include the narrative" means as an end result for a narrative only seciton. things will be similarly shaped with different stubbed out words and a different footnote but i don't want to overcomplicate things!

refiner/app/services/ecr/policy.py
rogeruiz: Good catch! I missed this. It should be cleared up now.

refiner/app/api/v1/configurations/model.py
```diff
@@ -13,9 +13,28 @@
...
+    These sets of LOINC codes drive UI behavior in the eICR Sections
+    table and are sourced from refiner policy + the eICR specification.
+    Shipping them as literal-typed defaults means the frontend imports
+    real values rather than re-declaring them.
+    """
+
+    disabled_sections = DisabledSection
+    narrative_only_sections = NarrativeOnlySection
+
+
```
fzhao99:
```suggestion
```

don't think we need this guy anymore

refiner/app/services/ecr/refine.py
```diff
@@ -462,12 +475,33 @@ def refine_eicr(
...
+            # RETAINED_NARRATIVE_REMOVED below because this branch
+            # reflects the spec's structural reality, not a
+            # configuration choice the jurisdiction made.
+            if not section_rules.narrative:
+                replace_narrative_with_removal_notice(
+                    section=section, namespaces=HL7_NS
+                )
+                outcome = SectionOutcome.NARRATIVE_ONLY_REMOVED
+            else:
+                outcome = SectionOutcome.NARRATIVE_ONLY_RETAINED
```
fzhao99, line 495: 💪

client/shared/openapi.json
fzhao99, line 1: would you mind minifying / putting this file back like you did for the other one?

fzhao99 (approved):
two minor things but otherwise lgtm! Giving an "approve with changes"

client/shared/openapi.json
rogeruiz: sure. I'm not sure what happened here. But it should be good now.

PR #1244, opened 2026-05-20
refactor: inline testing cleanup

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR does some minor refactoring on the inline testing service.

- Service function no longer requires that all associated config conditions be provided
  - Updated route helper function to no longer pull these values from the DB
- Remove intermediate `InlineTestingTrace` object in favor of using provided data directly
- Added inline testing result logging helper function

## 🧪 How to test

- Automated tests and app functionality should be unaffected


refiner/app/services/testing.py
```diff
@@ -467,42 +437,37 @@ async def inline_testing(
 
     # if no matching codes, then the eICR/RR pair was not suitable for testing the configuration
     if not matched_codes:
-        logger.warning(
```
jakewheeler, line 470: The new helper function uses `info` for both logs. I didn't think `warning` was a great fit here since folks potentially uploading a file without reportable conditions being identified is a completely valid and anticipated path.

Let me know if you disagree.

refiner/app/api/v1/configurations/testing.py
```diff
@@ -44,28 +44,22 @@ def _get_upload_zip() -> Callable[[DbUser, io.BytesIO, str, Logger], Awaitable[s
...
     )
+
     if primary_condition is None:
         raise HTTPException(
-            status_code=404,
+            status_code=status.HTTP_404_NOT_FOUND,
             detail="Primary condition not found for configuration.",
         )
 
-    if len(configuration.included_conditions) <= 1:
```
fzhao99, line 61: being able to remove this is a great sign for the way we've abstracted refinement-related info into the config object itself!

fzhao99 (approved):

PR #1246, opened 2026-05-20
chore: add migration and seeding for systems table

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Part one of the systems table work, split out for ease of review, that adds

1. The migration / system table seeding
2. The core services and routes

Actual refactor coming in a forthcoming PR

## 🔗 Related Issue

Part 1 of #1229 
- #1229

## 🧪 How to test
No functionality has been changed yet, so check main flows to make sure there aren't any app regressions

## ✅ Acceptance Criteria

- [x]  Systems information is populated on storage


refiner/app/services/ecr/specification/constants.py
```diff
@@ -30,3 +30,12 @@
 LOINC_OID: Final[str] = "2.16.840.1.113883.6.1"
 RXNORM_OID: Final[str] = "2.16.840.1.113883.6.88"
 SNOMED_OID: Final[str] = "2.16.840.1.113883.6.96"
+
+
+CODE_SYSTEM_LABELS: Final[dict[str, str]] = {
```
fzhao99, line 35: moved this into this file to avoid a circular import issue. Going to look into removing it in the forthcoming PR

refiner/scripts/seeding/seed_db.py
```diff
@@ -40,6 +42,7 @@ def seed_database(db_url: str, db_password: str) -> None:
         raise
 
     load_tes_data(db_url=db_url, db_password=db_password)
+    load_system_data(db_url=db_url, db_password=db_password)
```
fzhao99: @jakewheeler know you mentioned wanting to put all the db seeding work into one script, so I opted to put it here. Lmk if we would rather put it inside `load_tes_data` instead

jakewheeler: The ops container doesn't ever run this script so it'll need to go in `load_tes_data`.

refiner/scripts/seeding/load_system_data.py
```diff
@@ -0,0 +1,66 @@
+from config import logger
+from lib import (
+    get_db_connection,
+)
+
+CODE_SYSTEM_DATA = {
+    "snomed": {"oid": "2.16.840.1.113883.6.96", "display_name": "SNOMED"},
+    "loinc": {"oid": "2.16.840.1.113883.6.1", "display_name": "LOINC"},
+    "icd-10": {"oid": "2.16.840.1.113883.6.90", "display_name": "ICD-10"},
```
jakewheeler: I like the idea of using `icd10` as the key for simplicity. Would this be a pain at all to switch?

refiner/migrations/20260520185510_add_system_table.sql
```diff
@@ -0,0 +1,11 @@
+-- migrate:up
+CREATE TABLE systems (
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+    key text NOT NULL,
+    display_name text NOT NULL,
```
jakewheeler: We seem to sometimes use `name` and other times use `display_name`. I think your choice to go with the latter makes sense since we intend to show this value to users.

Nothing to do! Just thinking about this as I'm looking at it.

refiner/scripts/seeding/load_system_data.py
fzhao99: nope, happy to change it back

refiner/app/db/code_systems/db.py
```diff
@@ -0,0 +1,231 @@
...
+        async with conn.cursor(row_factory=class_row(DbCodeSystem)) as cur:
+            await cur.execute(query, params)
+            row = await cur.fetchall()
+            if len(row) == 0:
+                return None
+
+            if len(row) > 1:
+                logger.warning(
+                    f"Found multiple matches for code system when querying by oid {oid}. Returning first match"
+                )
```
jakewheeler: Is this possible? We have the unique constraint on `oid`.

refiner/migrations/20260520185510_add_system_table.sql
```diff
@@ -0,0 +1,11 @@
...
+    oid text UNIQUE
+);
+
```
jakewheeler, line 10: Is there any work we should be doing on existing custom codes in the DB as part of this migration?

Is it useful to attempt to add `key` or modify other data as a part of the custom code?

This is currently what a custom code object looks like in the DB for reference:
```json
  {
    "code": "22",
    "name": "1",
    "system": "ICD-10"
  }
```

refiner/app/db/code_systems/db.py
```diff
@@ -0,0 +1,231 @@
...
+    string_to_search = name.lower()
+    try:
+        by_name = await get_code_system_by_display_name_or_raise_db(
+            db=db, logger=logger, name=string_to_search
+        )
+        return by_name
+    except ValueError:
+        # fall back to search by key
+        by_key = await get_code_system_by_key_or_raise_db(db=db, key=string_to_search)
+        return by_key
```
jakewheeler: Not sure if [this](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1246/changes#r3276641787) would help tighten this up?

refiner/app/db/code_systems/db.py
```diff
@@ -0,0 +1,231 @@
...
+
+            if len(row) > 1:
+                logger.warning(
+                    f"Found multiple matches for code system when querying by display_name {name}. Returning first match"
+                )
+
+            return row[0]
+
+
+async def get_code_system_by_display_name_or_raise_db(
```
jakewheeler: Looks like some of these functions are intended to be private (aka `_get_code_system_by_display_name_or_raise_db` in Python world 😮‍💨)

refiner/app/db/code_systems/db.py
```diff
@@ -0,0 +1,231 @@
...
+        logger: Logger: The system logger.
+
+    Returns:
+        DbCodeSystem: Values from the systems table with matching display name.
+
+    Raises:
+        ValueError: if no code system is found
+    """
+
+    display_name = await get_code_system_by_display_name_db(
```
jakewheeler: I think this variable name is outdated. Looks like it should be `system` or something like that.

refiner/app/services/code_systems.py
```diff
@@ -0,0 +1,55 @@
+from app.db.code_systems.db import (
+    CodeSystemKey,
+    DbCodeSystem,
+    get_all_code_systems_db,
+)
+from app.db.pool import AsyncDatabaseConnection
+
+
+async def get_allowed_code_system_display_names(
```
jakewheeler: This appears to be unused.

jakewheeler: Actually looks like the others are as well 🤔 what am I missing?

fzhao99: This is point two of my message [here](https://skylight-hq.slack.com/archives/C08LC62CVNE/p1779298840323169) around putting in the db service next to the migration, but I'll pull this out since I think the rest of it is a digestible chunk without this file

refiner/migrations/20260520185510_add_system_table.sql
fzhao99: I think we could do one of two things:

1. Leave existing custom codes as is and layer in `system_key` vs `display_name` at the backend service level. This is kinda what I'm starting to do [here](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1160/changes), where I'm defining a new `ConfigurationCustomCode` object that has the extra key property that also renames the system information
2. Write some sort of mapping to translate / modify the custom code information.
     - Could either do this in part (ie when there are new code updates) or in full (migrate the whole column somehow) 

2 is the technically correct option but might be difficult given the timing of when this gets in / the reality of us being able to touch the prod AWS RDS. It may also not be much of an issue given we'll be refactoring / reworking the tables soon anyways

1 is the more pragmatic option, but does introduce the possibility of drift in the meantime between the strings we're storing and the key's were using to reference the data. 

Thoughts on the tradeoffs?

refiner/app/services/code_systems.py
jakewheeler: Ah my bad, I misunderstood but I see what you're getting at now. Yeah, I think this chunk being fully "workable" makes sense to me. It's a bit difficult for me to look at unused code and feel confident giving it a ✅ Not the biggest deal either way but that's how I was thinking about it.

refiner/scripts/seeding/load_static_data.py
```diff
@@ -239,6 +239,69 @@ def _build_condition_groupers(
...
 
 
+CODE_SYSTEM_DATA = {
+    "snomed": {"oid": "2.16.840.1.113883.6.96", "display_name": "SNOMED"},
+    "loinc": {"oid": "2.16.840.1.113883.6.1", "display_name": "LOINC"},
+    "icd10": {"oid": "2.16.840.1.113883.6.90", "display_name": "ICD-10"},
+    "rxnorm": {"oid": "2.16.840.1.113883.6.88", "display_name": "RxNorm"},
+    "cvx": {"oid": "2.16.840.1.113883.12.292", "display_name": "CVX"},
+    "other": {"oid": None, "display_name": "Other"},
+}
```
rogeruiz, line 237: I wonder if it may make sense to maintain this via a config or an external system in the future? 🤔

refiner/scripts/seeding/load_static_data.py
```diff
@@ -272,6 +335,21 @@ def load_tes_data(db_url: str, db_password: str) -> None:
     logger.info("🏁 Done!")
 
 
+def load_static_data(db_url: str, db_password: str) -> None:
```
rogeruiz, line 315: good orchestration here 👍🏼

refiner/migrations/20260520185510_add_system_table.sql
rogeruiz: Jumping in here with my 2¢. I think option 1 is the better for now given the fast-follow of the rest of the work. I'd say we should document the tech debt as much as possible (e.g. `TODO` comments, tests, etc) if we're gonna take more than one sprint to implement the rest of the refactor.

refiner/scripts/seeding/load_static_data.py
fzhao99: yeah if there's ever a need, it's possible we can look into getting this data from an external source. The TES seeding gives us some good precedence for seeding the data via a static file / script, so building it out this way is helpful for future extensibility

refiner/app/services/code_systems.py
fzhao99: pulled out most of this file and will include it in the followup

refiner/migrations/20260520185510_add_system_table.sql
jakewheeler: I was thinking about this some more and realized the timing doesn't work out for what I was suggesting. When the migration runs the `systems` table is going to be empty. If we wanted to do any additional work we'd need to work from whatever system info is available in existing custom codes.

I'm good with doing whatever we think is most straightforward and then switching to custom codes being associated with a system ID as soon as we can make that happen.

refiner/scripts/seeding/load_static_data.py
```diff
@@ -239,6 +239,69 @@ def _build_condition_groupers(
...
+    db_password: str,
+):
+    """
+    Function to upsert system information.
+    """
+    try:
+        with get_db_connection(db_url, db_password) as conn, conn.cursor() as cursor:
+            logger.info("⏳ Upserting system data...")
+
+            system_upsert_query = """
```
jakewheeler: I'm noticing that this query will keep inserting `other` each run. A couple of thoughts this gave me:
1. Maybe we should enforce that `key` and `display_name` in `systems` also be unique
2. Can we add `created_at` and `updated_at` to `systems` for audit purposes?

Here's what I'm seeing after running the script a few times:
Image

refiner/scripts/seeding/load_static_data.py
```diff
@@ -239,6 +239,69 @@ def _build_condition_groupers(
...
+                key,
+                display_name,
+                oid
+            )
+            VALUES (
+                %(key)s,
+                %(display_name)s,
+                %(oid)s
+            )
+            ON CONFLICT (oid)
```
jakewheeler: We'll need to update this to handle other potential conflicts. I'm getting an error on `key` when re-running the script after the migration update.
```sh
INFO:config:⏳ Upserting system data...
ERROR:config:❌ A critical error occured during upsert of system data
  File "/Users/skylight/Documents/dev/cdc/dibbs-ecr-refiner/refiner/scripts/seeding/load_static_data.py", line 293, in load_system_data
    cursor.executemany(system_upsert_query, params)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/skylight/Documents/dev/cdc/dibbs-ecr-refiner/refiner/venv/lib/python3.13/site-packages/psycopg/cursor.py", line 148, in executemany
    raise ex.with_traceback(None)
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "systems_key_key"
DETAIL:  Key (key)=(other) already exists.
INFO:config:⏱️  Static data loaded in 4.760 seconds
```

refiner/scripts/seeding/load_static_data.py
```diff
@@ -239,6 +239,69 @@ def _build_condition_groupers(
...
+
+def load_system_data(
+    db_url: str,
+    db_password: str,
+):
+    """
+    Function to upsert system information.
+    """
+    try:
+        with get_db_connection(db_url, db_password) as conn, conn.cursor() as cursor:
```
jakewheeler: I'm kind of thinking we should load condition data and system data all in the same transaction. Even though it shouldn't be an issue to re-run the script, I think I'd feel better knowing that if either action fails nothing gets stored. What do you think?

refiner/migrations/20260520185510_add_system_table.sql
```diff
@@ -0,0 +1,13 @@
+-- migrate:up
+CREATE TABLE systems (
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
+    key text UNIQUE NOT NULL,
+    display_name text UNIQUE NOT NULL,
+    oid text UNIQUE,
+    created_at timestamptz NOT NULL DEFAULT now(),
+    updated_at timestamptz NOT NULL DEFAULT now()
```
jakewheeler, line 8: We'll need to add a trigger for `updated_at` as well. There should be examples of this in a few of the existing migration files.

Here's one from `20260402215326_add_coverage_status.sql` for reference:
```sql
CREATE TRIGGER update_conditions_context_groupers_updated_at
BEFORE UPDATE ON conditions_context_groupers
FOR EACH ROW
EXECUTE FUNCTION set_updated_at();
```

fzhao99: thanks for catching! added the trigger / drop statement to the up and down migrations

refiner/scripts/seeding/load_static_data.py
fzhao99: updated this so that it'll skip on key and OID (since those get written to S3 and have semantic meaning in the refinement run and may affect matching in the future when we hook the plan construction up with the DB) but will update display name. Lmk if that makes sense

refiner/scripts/seeding/load_static_data.py
fzhao99: Good suggestion. Refactored things so that the steps share a connection so that they're transacted together

refiner/scripts/seeding/load_static_data.py
fzhao99: good catch, should be fixed now!

refiner/scripts/seeding/load_static_data.py
jakewheeler: Yep that makes sense to me!

refiner/scripts/seeding/load_static_data.py
```diff
@@ -0,0 +1,349 @@
...
+
+    New rows are inserted. Existing rows with the same oid, key, or
+    are updated only when relevant fields have changed. This is done in a single transaction to systems data to ensure the insert either succeeds or fails all at once
+
+    Args:
+       cursor: A DB cursor
+    """
+    logger.info("⏳ Upserting system data...")
+
+    system_upsert_query = """
```
jakewheeler, line 255: This query seems to cause `updated_at` to change every run when things aren't changing. Are you seeing the same thing?

fzhao99: oh yeah looks like since MERGE AT does a match check on each row, Postgres will treat that as a trigger match that will always fire the `set_update_at()`. 

Updated the trigger so that it only fires if there's an update to display name. This does introduce the possibility of things drifting over the course of time if the conflict criteria changes, so I add some comments to hopefully cover that case. I think that's slightly more preferable than doing something like manually updating `updated_at` in the query in the event that there's ever any other script that updates that table, since it aligns that column's behavior with the other instances of `updated_at`. 

Lmk if there's an issue I'm not seeing / another way to handle this

jakewheeler: This seems to be working the way I'd expect now. I don't see any issues with the way you've handled it. I would expect a future migration to make an update to the trigger if it's needed at that point in time so I think this is good.

Thanks!

jakewheeler (approved):
Feels like a good spot to call this part of the work. Thanks Bob!

PR #1250, opened 2026-05-22
feat(server): Update TES Condition Groupers to version 6.0.0

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the assets used to seed the refiner's backend database. In order to avoid GitHub's file limit we're now sharding the files--this didn't affect the seeding process as you'll notice that code isn't updated. The changes were mostly made in the gather or the files from the FHIR ValueSets we pull from the TES API. Some adjustments were made to the integration tests since we've updated the versions available (and therefore what is considered 'default' now).

## 🔗 Related Issue

Fixes #1194 
- #1194

PR #1253, opened 2026-05-22
fix: Display loading states when activating and deactivating configurations

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR addresses an issue we've begun to see as a result of #1250, which is that configuration activation and deactivation can now take longer to complete. Loading states have been added to give the user some feedback and show them that their action is working.

Previously, configuration activation and deactivation was so snappy the vast majority of the time that having no explicit loading state provided an acceptable user experience. However, now that configuration sizes have the possibly to grow much larger much more easily, it now makes sense to give the user some feedback that something is happening while activation or deactivation is in progress.

Here's an example of what these loading states look like, which demonstrates **activation**, **deactivation**, and **switching**:
Kapture 2026-05-26 at 11 17 24

## ✅ Acceptance Criteria

- App loading states work as shown in GIF
- Automated testing continues to pass


client/src/pages/Configurations/ConfigActivate/ActivationButtons.tsx
```diff
@@ -22,41 +24,38 @@ export function ActivationButtons({
...
-            heading: 'Error activating condition',
-            body: errorDetail,
-          });
-        },
-      }
-    );
+  async function handleActivation() {
+    setIsLoading(true);
+    try {
+      await activate({ configurationId: configurationData.id });
```
jakewheeler, line 38: `onSuccess` would complete before `invalidateQueries()` finished and the UI was updated, which resulted in the UI briefly flashing the buttons again in their normal state prior to the modal closing. Using the async mutation functions and handling the state manually seemed like the most straightforward solution.

fzhao99 (approved):

PR #1276, opened 2026-05-27
fix: improve config builder performance by virtualizing code set table

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR pulls in and makes use of `@tanstack/react-virtual` for the code set tables. This greatly improves client performance by preventing the browser from rendering all code rows (potentially 10s of thousands) at once. Instead, it only renders codes that are in view.

### Server-side solution?

The main issue this PR is solving is a rendering bottleneck. While it may be preferable to move code related functionality to the server at some point in the future, it is tough right now due to codes being stored as jsonb. Functionality like searching and pagination will be much simpler to implement in the future when codes are normalized and stored in their own table.

## 🔗 Related Issue

Fixes #1275

## 🧪 How to test

1. Pull down this branch and install npm dependencies
2. Navigate to the config builder page, add large code sets such as COVID-19 or Influenza
3. Click around. You should notice performance is significantly improved while codes are still scrollable, filterable, and searchable
4. Check that the builder page looks and functions about the same as before otherwise


client/src/pages/Configurations/ConfigBuild/index.test.tsx
```diff
@@ -20,6 +20,27 @@ import {
   useValidateCustomCodeFromConfiguration,
 } from '../../../api/configurations/configurations';
 
+vi.mock('@tanstack/react-virtual', () => ({
```
jakewheeler, line 23: This needs to be mocked since there is no browser in a testing environment. The tests fail when this isn't mocked because only 1 item ends up being rendered.

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -288,65 +300,84 @@ export function ConditionCodeTable({
...
+
+            <div role="rowgroup" className="contents">
+              <div
+                className="col-span-full"
+                style={{ height: `${topSpacer}px` }}
+              />
+
+              {virtualItems.map((virtualRow) => {
+                const code = visibleCodes[virtualRow.index];
                 const matchingResult = results.find(
```
lina-roth: Thoughts on updating this to help performance. only virtual rows render, but if `results` is huge using a map i think would be cleaner here

```
               const resultsByCode = useMemo(() => {
  return new Map(
    results.map((result) => [
      `${result.item.system}-${result.item.code}-${result.item.description}`,
      result,
    ])
  );
```

```
const matchingResult = resultsByCode.get(
  `${code.system}-${code.code}-${code.description}`
);
```

lina-roth (commented):
I had just one suggestion - curious on your thoughts. Otherwise looks good overall

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
jakewheeler: Updated this!

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -206,29 +205,43 @@ export function ConditionCodeTable({
...
+
+  /**
+   * TODO: Known issue
+   * See: https://github.com/TanStack/virtual/issues/1119
+   */
+  // eslint-disable-next-line react-hooks/incompatible-library
+  const virtualizer = useVirtualizer({
+    count: visibleCodes.length,
+    getScrollElement: () => parentRef.current,
+    estimateSize: () => 48,
```
fzhao99, line 241: just to check my understanding:

- 48 is the height of each of the rows here for the virtualizer to guess how many rows to render. Are we pulling this from CSS somehow, and if so, would it be useful to pull it in a common var?
- After we filter on search string, system, etc., count is the number of rows in the window of the virtualizer that it needs to manage, which presumably the library uses to calculate the window of values to return / overscan, etc.

jakewheeler: - Correct, this is truly just an estimate of the height of each row for it to use internally. I ended up at this value by manually inspecting the height of the rows in the rendered document and this seemed about right for various code sets I tested. We don't explicitly set the row heights so I'm not sure if pulling it out is useful but am open to suggestions if you have an idea
- 👍

fzhao99: think it's fine either way. think it'll be pretty obvious in the event we change the styling on the table in a way that breaks the virtualization

fzhao99 (approved):

PR #1284, opened 2026-05-28
style: config builder menu item counts use commas when large

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the config builder list item counts to use commas when needed (>= thousands).

For example:
- `1234` becomes `1,234`
- `12345` becomes `12,345`

image

## 🧪 How to test

1. Load up the app and navigate to the configure builder page
2. Add a large code set, such as COVID-19 or Influenza
3. Notice code counts will include a comma


rogeruiz (approved):
lgtm 🌈

PR #1286, opened 2026-05-29
test(server): Scenario based testing to mirror Tim's tests

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Adds a scenarios test suite at `tests/integration/scenarios/` that exercises the production refinement pipeline--the same path lambda runs--against committed eICR/RR fixtures. Each scenario authors its configuration through the live API (`create → customize → activate → read the activation payload back from localstack`), then asserts in two layers: validation (CDA R2 XSD + schematron) and snapshot comparison against committed expected files.

Configurations are built from declarative recipes rather than committed JSON: a `Scenario` names the condition plus the custom codes, section-processing overrides, and associated conditions to layer on, and a shared `build_scenario_configuration` fixture interprets the recipe against the running app. This gives a single authoring path that uses the exact serialization production uses, with no committed `active.json` to drift--updating a scenario means editing its recipe, not hand-editing or re-downloading JSON.

The suite contains 8 snapshot scenarios on the `all_sections_COVID_INFLUENZA` fixture, 8 explicit named-assertion tests, and 3 smoke tests pinning the harness itself (19 total). Every refined document is XSD- and schematron-validated on every test run. An auto-generated `REPORT.md` summarizes the suite for stakeholder review.

Because the suite lives under `tests/integration/`, it runs as part of the existing integration CI job automatically--no separate CI wiring needed, and these tests passed in CI on this PR.

All six issues from the original testing roll-up sheet are directly covered: entry matching across OID mismatch, custom codes in nested locations (`entryRelationship/value` and `substanceAdministration`), procedure retention via unrelated `entryRelationship` codes, vital sign panel pruning, code-set bloat, and schematron conformance of refined output. The previously informal "Tim runs through a spreadsheet manually" workflow is now a CI-runnable suite.

## 🔗 Related Issue

Fixes #1240
- #1240

## ✅ Acceptance Criteria

- [x] A new test layer exercises the production refinement pipeline end-to-end against committed fixtures, with configurations authored through the live API so there is a single authoring path and no committed configuration JSON to drift.
- [x] Snapshots pin refinement behavior (entry matching, identifier derivation, section pruning) such that regressions surface as snapshot diffs rather than silent drift.
- [x] All six issues from the original roll-up sheet are directly covered, with five additionally backed by named explicit assertions that fail with a behavior-level message rather than an opaque snapshot diff.
- [x] Every refined document is validated against CDA R2 XSD and schematron on every test run; errors and fatal severity fail the test.
- [x] A stakeholder-facing report (`REPORT.md`) is generated deterministically from the committed snapshots.
- [x] Developer workflow for authoring new scenarios and fixtures is documented.
- [x] The suite runs as part of integration CI.

## 🧪 How to test

1. From the project root, `cd refiner/`.
2. Activate a virtual environment.
3. Install `requirements.txt` and `dev-requirements.txt`.
4. Ensure the integration stack is available (docker compose / localstack), same as the rest of `tests/integration/`.
5. Run `pytest tests/integration/scenarios/`.
6. You should see 19 tests pass.
7. Inspect `tests/integration/scenarios/REPORT.md` for the stakeholder-facing summary of what the suite pins.
8. Browse `tests/integration/scenarios/snapshots/<fixture>/<scenario>/` to see committed expected output per scenario (`expected_trace.json` is the most legible starting point).
9. Read `tests/integration/scenarios/README.md` for the full developer workflow: authoring new scenarios, adding new fixtures, regenerating the report.

To verify the snapshot regeneration workflow:

```bash
pytest tests/integration/scenarios/test_all_sections_covid_influenza.py --update-snapshots
python tests/integration/scenarios/authoring/build_report.py
git diff
```

The diff should be empty if the committed state is current; a non-empty diff indicates drift between committed snapshots and current refinement output, or between snapshots and the report.

## ℹ️ Additional Information

**Design choices worth flagging for review:**

- **Configurations are built through the live API from declarative recipes**, then read back from localstack in the production activation-format JSON. This keeps the suite on the same serialization and code path production uses, with a single authoring path and no committed `active.json` to maintain.
- **Refinement timestamp is fixed** (`20260101000000+0000`) so deterministic UUIDv5 identifier derivation produces stable snapshots run-to-run. Without this, snapshots would never match between runs.
- **Augmentation is seeded by the activating jurisdiction** (the integration test user's, `SDDH`), not the RR's reportable-to jurisdiction. The fixture RR may be reportable elsewhere; that is intentionally not consulted, so arbitrary test data can be reused without standing up matching fake jurisdictions.
- **The activation payload is fetched by discovering the active version from `current.json`** rather than assuming version 1, so tests that build more than one configuration for the same condition resolve each correctly.
- **Three snapshot artifacts per scenario** (trace JSON + refined eICR + refined RR) so regressions surface at the right granularity: top-line outcome changes in the trace, structural changes in the XML.
- **Explicit assertions in `test_explicit_assertions.py`** complement the snapshot scenarios. They build configs through the same recipe path and pin specific named behaviors with precondition guards that fail diagnostically if the fixture or configuration drifts in a way that makes the test no longer exercise its named concern.

**Deferred to follow-up PRs (small, well-scoped):**

- CI step that runs `python tests/integration/scenarios/authoring/build_report.py` and fails on `git diff --exit-code REPORT.md` to prevent the report from going stale relative to snapshots. A `TODO` at the top of `build_report.py` flags this.
- Lift `normalize_xml` (currently imported from `tests/unit/conftest.py`) to a shared `tests/_helpers/` module. The scenarios suite still reaches into the unit suite for this one helper; functional but architecturally awkward. (The integration validation helpers are now inherited/imported from the parent integration conftest, which is the suite's natural parent.)

**Runtime:** 19 tests; the suite now runs on the integration path (docker compose / localstack) rather than infra-free, since configs are authored through the live API. Each scenario does a full create → activate → fetch cycle; if the suite grows substantially, a session-scoped cache of built configurations keyed by scenario name is the natural mitigation.


refiner/scripts/exports/export-groupers-report.py
```diff
@@ -0,0 +1,426 @@
+import csv
```
jakewheeler: Did you mean to check this (and the .csv) in with this work? I am confused exactly how it's related / what I should be looking at here.

refiner/tests/integration/scenarios/test_scenario_wiring.py
```diff
@@ -0,0 +1,232 @@
...
+    assert result.trace.configuration_resolved is True
+    assert result.trace.configuration_version == CONFIGURATION_VERSION
+    assert result.trace.canonical_url == COVID_CANONICAL_URL
+    assert result.trace.skip_reason is None
+    assert result.trace.error_detail is None
+
+    # size reduction must be populated and positive: refining for COVID
+    # against a multi-condition eICR drops the non-COVID content.
+    assert result.trace.eicr_size_reduction_percentage is not None
+    assert result.trace.eicr_size_reduction_percentage > 0
```
jakewheeler: Is it worth making these checks? Are we able to make it this far if a lot of these values are wrong or missing?

refiner/scripts/exports/export-groupers-report.py
robertmitchellv: this was something that kyle asked me to produce and i wanted to add it in case i had to rerun it again. it can/should be it's own if you want. it's just sort of coming along for the ride. this is just to produce a specific kind of output (the csv)

refiner/tests/integration/scenarios/test_scenario_wiring.py
robertmitchellv: you're not wrong; this was the first thing that i worked on to make sure i had everything up and running so i can test. i decided to leave it in as more or less a smoke test but i would also not be opposed to removing it

refiner/scripts/exports/export-groupers-report.py
jakewheeler: Is there anything additional that needs to be documented about it? Or anything I should be checking?

refiner/tests/integration/scenarios/test_scenario_wiring.py
jakewheeler: I was going back and forth on this. The thing that is a bit odd about `trace` is that it kind of acts as both a refining input but also a logging tool.

If refining can't go on without the values being correct I would say it's fine to remove.

refiner/scripts/exports/export-groupers-report.py
robertmitchellv: i think i'm going to remove this from the pr and just add a local ignore for things that i usually end up running as reports. it's probably not necessary to check in everything i produce that becomes some kind of artifact for understanding the data in the tes

robertmitchellv: this is all wrapped up now

refiner/tests/integration/scenarios/test_scenario_wiring.py
robertmitchellv: i whittled this down so let me know if it works for you

refiner/tests/integration/scenarios/test_all_sections_covid_influenza.py
```diff
@@ -0,0 +1,166 @@
+import json
+from pathlib import Path
+
+import pytest
+
+from ...unit.conftest import normalize_xml
```
jakewheeler, line 6: I'm not too concerned about it as a part of this PR but I wonder if it'd make sense to have fixtures like this in a `tests/conftest.py` to share across our various test packages.

refiner/tests/integration/scenarios/conftest.py
```diff
@@ -0,0 +1,433 @@
...
+        assert response.status_code == status.HTTP_200_OK, (
+            f"Could not fetch active.json at {url} (status {response.status_code})"
+        )
+        return ProcessedConfiguration.from_dict(response.json())
+
+    return _get
+
+
+@pytest_asyncio.fixture
+async def build_scenario_configuration(
```
jakewheeler, line 368: This is cool 👍

refiner/tests/integration/scenarios/authoring/build_report.py
```diff
@@ -0,0 +1,543 @@
...
+def _render_header() -> str:
+    return (
+        dedent("""
+            # eCR Refiner — Scenarios Report
+
+            *Auto-generated. To regenerate:*
+
+            ```
+            pytest tests/scenarios/ --update-snapshots
+            python tests/scenarios/authoring/build_report.py
```
jakewheeler:
```suggestion
            pytest tests/integration/scenarios/ --update-snapshots
            python tests/integration/scenarios/authoring/build_report.py
```

refiner/tests/integration/scenarios/authoring/build_report.py
```diff
@@ -0,0 +1,543 @@
...
+            ```
+
+            This report summarizes the behaviors pinned by the scenarios test suite at `tests/scenarios/`. Each scenario refines a committed eICR/RR pair against a committed configuration JSON and asserts in two layers: validation (well-formedness, CDA R2 XSD, schematron) and snapshot comparison against committed expected files.
```
jakewheeler:
```suggestion
            This report summarizes the behaviors pinned by the scenarios test suite at `tests/integration/scenarios/`. Each scenario refines a committed eICR/RR pair against a committed configuration JSON and asserts in two layers: validation (well-formedness, CDA R2 XSD, schematron) and snapshot comparison against committed expected files.
```

refiner/tests/integration/scenarios/authoring/build_report.py
```diff
@@ -0,0 +1,543 @@
...
+def _render_appendix() -> str:
+    return (
+        dedent("""
+            ## Appendix — running the suite
+
+            ```
+            pytest tests/scenarios/                                  # run all scenarios + smoke tests
+            pytest tests/scenarios/test_<fixture>.py -k <scenario>   # one scenario
+            pytest tests/scenarios/ --update-snapshots               # regenerate after intentional changes
+            python tests/scenarios/authoring/build_report.py        # regenerate this report
```
jakewheeler: Maybe `tests/integration/scenarios` should go in a variable? I left a couple corrections up above but we're referencing the path quite a bit in this file. Might be easier to keep everything synced up with a variable.

refiner/tests/integration/test_configurations.py
```diff
@@ -362,32 +363,25 @@ def require_section_by_code(sections, code):
...
+            narrative=False,
         )
-        assert response.status_code == status.HTTP_200_OK
-        update_response = response.json()
         assert update_response["section_updated_code"] == admission_diagnosis_code
 
         # Get the updated admission diagnosis section
         response = await authed_client.get(f"/api/v1/configurations/{draft_id}")
-        response.status_code == status.HTTP_200_OK
+        assert response.status_code == status.HTTP_200_OK
```
jakewheeler, line 377: Thank you 🤦

refiner/tests/integration/scenarios/test_scenario_wiring.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+
+    return XMLFiles(
+        eicr=load_fixture_str("eicr_v3_1_1/multi-condition-multi-covid-CDA_eICR.xml"),
+        rr=load_fixture_str("eicr_v3_1_1/multi-condition-multi-covid-CDA_RR.xml"),
+    )
+
+
+@pytest.fixture
+def covid_processed_configuration() -> ProcessedConfiguration:
```
jakewheeler: Are we still intending to use this? I'm not sure I understand what this is for.

refiner/tests/integration/scenarios/test_all_sections_covid_influenza.py
robertmitchellv: yeah i think a test cleanup sort of thing or general `tests/conftest.py` is a good idea

refiner/tests/integration/scenarios/test_scenario_wiring.py
robertmitchellv: i removed parts of it but i think fundamentally this was supposed to prove that the harness works and then the other file is for the particular fixture. but we _could_ remove it because if the harness doesn't work the test will fail anyway. if you'd prefer to keep things more lean since these tests add more time i'm ok with that

jakewheeler: I think it'd be better to remove it. I agree with what you're saying + Lambda's test config being hard-coded isn't great so I wouldn't mind not depending on it here.

robertmitchellv: sounds good; just pushed up the change that removes it

refiner/tests/integration/scenarios/conftest.py
```diff
@@ -0,0 +1,433 @@
...
+    """
+
+    current_code: str
+    include: bool | None = None
+    narrative: bool | None = None
+    action: str | None = None  # DbSectionAction; kept as str to avoid the import
+
+
+@dataclass(frozen=True)
+class CustomCode:
```
fzhao99, line 78: feel free to handle in a follow up: we should probably swap this out for `DbConfigurationCustomCode` since they're basically the same object. It seems we use that object anyways in the API script?

fzhao99 (approved):

jakewheeler (approved):
🫡

PR #1287, opened 2026-05-29
fix: improve configurations list performance

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR attempts to address a performance issue with the `GET /configurations` route that becomes more noticeable as the number of configurations grows within a jurisdiction. 

- Remove somewhat complex Python logic in favor of letting the DB handle the work
  - Query pulls a more limited data set than before
  - Query handles "ranking" instead of Python, which ends up being more straightforward
  - Query handles the final sorting 
- Reduce number of DB queries required to build response

## 🧪 How to test

There is a measurable improvement even when the total number of configurations is low.

My setup (3 total configs):

- Acanthamoeba (`active`)
- COVID-19 (`draft`)
- Influenza (`inactive`)

There is some variance when running this locally, however I am seeing roughly these numbers when loading `GET /configurations`:

- this branch: ~9ms
- main: ~100ms

**NOTE:** this change should have no other impact on how the application functions.

### DIFF
--- a/refiner/app/api/v1/configurations/base.py
+++ b/refiner/app/api/v1/configurations/base.py
@@ -9,19 +9,15 @@ from app.db.conditions.db import (
     get_condition_by_id_db,
     get_conditions_by_version_db,
     get_primary_condition_db,
-    get_primary_conditions_for_configurations_db,
 )
 from app.db.configurations.db import (
     get_configuration_by_id_db,
     get_configuration_versions_db,
-    get_configurations_db,
+    get_configurations_summary_db,
     get_latest_config_db,
     get_total_condition_code_counts_by_configuration_db,
     insert_configuration_db,
     is_config_valid_to_insert_db,
-)
-from app.db.configurations.model import (
-    DbConfiguration,
 )
 from app.db.pool import AsyncDatabaseConnection, get_db
 from app.db.users.db import get_user_by_id_db
@@ -29,7 +25,6 @@ from app.db.users.model import DbUser
 from app.services.configuration_locks import ConfigurationLock
 from app.services.configurations import (
     format_section_naming,
-    get_canonical_url_to_highest_inactive_version_map,
 )
 from app.services.logger import get_logger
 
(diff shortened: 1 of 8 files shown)


refiner/app/db/configurations/db.py
```diff
@@ -1202,6 +1203,48 @@ async def get_configuration_versions_db(
...
+                c.name,
+                c.status,
+                cond.canonical_url,
+                ROW_NUMBER() OVER (
+                    PARTITION BY cond.canonical_url
+                    ORDER BY
+                        CASE c.status
+                            WHEN 'active' THEN 1
+                            WHEN 'inactive' THEN 2
+                            WHEN 'draft' THEN 3
```
fzhao99: can you say more about ranking inactive conditions over draft ones? my first impression is that in cases where we have all inactive and a singular draft, the draft one is the one most often usable by the frontend since it's the editable one.

lmk what I might be missing though

refiner/app/api/v1/configurations/base.py
```diff
@@ -62,81 +57,12 @@ async def get_configurations(
...
-    }
-
-    # inactive config with the highest version by condition
-    highest_version_inactive_configs_map: dict[str, DbConfiguration] = (
-        get_canonical_url_to_highest_inactive_version_map(
-            configs=all_configs, config_id_to_canonical_url=config_id_to_canonical_url
+    return [
+        GetConfigurationsResponse(id=c.id, name=c.name, status=c.status)
+        for c in await get_configurations_summary_db(
+            jurisdiction_id=user.jurisdiction_id, db=db
```
fzhao99, line 63: crazy we can get better performance and a simplified get operation here. Yay for Postgres optimizations!

refiner/app/db/configurations/db.py
jakewheeler: Great catch, that's my bad. You're right that draft should be ranked above inactive. I'll add a test for this.

fzhao99 (approved):
🚀

PR #1291, opened 2026-05-29
fix: improve activity log request performance

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes an issue where the activity log page can take a long time to load as the number of events within a jurisdiction grows. 

This PR does two main things to address the performance issue:
1. Uses `asyncio.gather` to allow tasks that don't depend on previous tasks to run concurrently
2. Let the DB handle filter option building with `get_event_filter_options_db()`

## 🧪 How to test

The activity log functionality itself should remain unchanged.

The performance can be tested by:
1. Performing actions to add events to the activity log
2. Loading the activity log page and checking how long it took
3. Do the same on `main`, compare the time

For example, I added 13 events to the activity log. On this branch it took `12ms` to load the data. On `main`, it took `97ms`.


refiner/app/api/v1/events.py
```diff
@@ -77,8 +76,18 @@ async def get_events(
...
+        ),
+        get_events_by_jd_db(
+            jurisdiction_id=jd,
+            page=page,
+            page_size=PAGE_SIZE,
+            canonical_url=canonical_url,
+            db=db,
+        ),
+        get_event_filter_options_db(jurisdiction_id=jd, db=db),
     )
```
rogeruiz: Hmm, should we add error handling here? This should help with debugging and makes the UX a bit more forgiving as well. I'm thinking we could use  `return_exceptions=True` as the last argument for `asyncio.gather()`. I can do a fast-follow PR if this make sense to add. But right now I'm thinking my changes may be out of scope for this PR. 😅

refiner/app/db/events/db.py
```diff
@@ -90,6 +101,33 @@ async def is_event_valid(
...
+            c.id,
+            c.name,
+            cond.canonical_url
+        FROM configurations c
+        JOIN configurations_conditions cc ON cc.configuration_id = c.id AND cc.is_primary = true
+        JOIN conditions cond ON cond.id = cc.condition_id
+        WHERE c.jurisdiction_id = %s
+        ORDER BY cond.canonical_url, c.name
+    ) sub
+    ORDER BY LOWER(name)
```
rogeruiz: Is the double `ORDER BY` here redundant?

rogeruiz (approved):
lgtm 🌈 

Great work on this @jakewheeler. 

> [NOTE!]
> Non-blocking suggestions or questions.

refiner/app/api/v1/events.py
jakewheeler: Yeah I think so! I'll add it here. I think I'm going to avoid `return_exceptions=True` so that it'll error out as soon as one query fails.

refiner/app/db/events/db.py
jakewheeler: The sub query's `ORDER BY` is meant to keep the read deterministic but I think I have it slightly wrong. I believe it should actually be:
```
ORDER BY cond.canonical_url, c.version DESC
```

This way it'll select the first distinct row it sees which should be the one with highest config version since that's the most relevant.

Hopefully it won't matter a whole lot in practice (assuming the condition name isn't changing) but I'll update this just to be sure we're showing the latest data if it does.

PR #1293, opened 2026-06-01
refactor: add normalized code tables

### LINKED ISSUE #1111: [TASK] Serve codes from the new code storage model
Implementation ticket based on the outcome of #1110

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Adds the proposed normalized codes and custom codes table, as well as a condition <> codes join representing RSG SNOMED codes and associated updates to the local seeding script.

These are purely the backend model / script updates: code to refactor RSG ingestion and display coming in a future PR.

## 🔗 Related Issue

Fixes #1111
- #1111

## ✅ Acceptance Criteria

- [ ] Custom codes and codes normalization table exists
- [ ] Condition<>codes join table representing RSG child snomeds exists
- [ ] Seeding script is updated to seed the relevant data into codes / join table

## 🧪 How to test
Verification queries: 

```sql
SELECT SUM(total_elements) FROM (
    SELECT cardinality(c.child_rsg_snomed_codes) AS total_elements FROM conditions as c
) AS subquery;
```

```sql
SELECT r.condition_id, r.code_id, c.display_name, c.version, cardinality(c.child_rsg_snomed_codes) AS total_elements FROM conditions as c 
FULL OUTER JOIN condition_child_rsg_codes as r ON c.id = r.condition_id
ORDER BY  c.display_name, c.version;
```
```sql
SELECT r.condition_id, r.code_id, c.display_name, c.version, cardinality(c.child_rsg_snomed_codes) AS total_elements FROM conditions as c 
FULL OUTER JOIN condition_child_rsg_codes as r ON c.id = r.condition_id WHERE c.display_name = 'COVID-19'
ORDER BY  c.display_name, c.version;
```

- Start on a cut of main with seeding complete. Create a couple configurations, and add whatever custom interactions you'd like.
- Use the first `total_elements` query to note the number of RSG codes we're expecting in the new table after seeding (should be 1614)
- Switch to this branch and run migrations and seed. Notice they work correctly.
- Run the next two queries to verify 
    1. the seed added all the RSG codes we'd expect into the table (values match what was on main in query 1)
    2. Seeding added an additional code for COVID between 5.0.0 and 6.0.0 due to an update in the TES
- Test the down migration back onto main

## ℹ️ Additional Information
Anything else we should test?


refiner/scripts/seeding/load_static_data.py
```diff
@@ -104,6 +198,7 @@ def _upsert_conditions_and_groupers(
     condition_upsert_query = """
         WITH upsert_condition AS (
             INSERT INTO conditions (
+                id,
```
fzhao99: adding the ID insert here so that Python can generate / keep track of the condition ID inserts for the join relationships

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,53 @@
...
+    created_at timestamptz NOT NULL DEFAULT now(),
+    updated_at timestamptz NOT NULL DEFAULT now(),    
+    UNIQUE (system_id, version, value),
+    CONSTRAINT fk_codes_system_id_fkey 
+        FOREIGN KEY (system_id) 
+        REFERENCES systems (id) 
+        ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS condition_child_rsg_codes (
```
jakewheeler: ~~Should we call this `conditions_codes`?~~ Never mind, I think that's kind of confusing 🤔

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,53 @@
...
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
```
jakewheeler: Can we get away without `id`? Always referencing the record by `condition_id`/`code_id` seems right to me.

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,53 @@
+-- migrate:up
+CREATE TABLE IF NOT EXISTS custom_codes(
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    name TEXT NOT NULL,
+    value TEXT NOT NULL,
```
jakewheeler, line 5: Nothing we need to do in this PR but it would be great to put some limits in place around custom codes in the near-ish future.

refiner/migrations/20260604134336_add_codeset_tables.sql
fzhao99: yeah I think down the road we'd also have a join between the configuration and code tables that would hold the non-RSG, codeset codes. Think the idea is to have separate joins between condition and code that are somewhat buisness-logic aware (ie one for RSG codes, another for condition codeset codes, maybe also for ACG codes?) or otherwise to expand the codes table with columns that allow for that.

very much open to feedback on how to do it differently though.

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,91 @@
...
+            FOREIGN KEY (system_id) 
+            REFERENCES systems (id) 
+            ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS codes(
+    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+    name TEXT NOT NULL,
+    value TEXT NOT NULL,
+    version TEXT NOT NULL,
```
jakewheeler, line 20: Should codes have a `version`? Should this be determined by the link to the condition since those already have a version?

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,91 @@
...
+
+ALTER TABLE configurations_conditions 
+    DROP CONSTRAINT configurations_conditions_condition_id_fkey;
+
+ALTER TABLE configurations_conditions 
+    ADD CONSTRAINT configurations_conditions_condition_id_fkey 
+    FOREIGN KEY (condition_id) 
+    REFERENCES conditions (id) 
+    ON UPDATE CASCADE 
+    ON DELETE CASCADE;
```
jakewheeler: It might make sense to use `RESTRICT` here instead so that we'd be notified about (probably accidental) config modification rather than it happening silently. We don't and likely won't delete conditions but maybe worth including just in case?

I don't think this causes any teardown issues with ou

```suggestion
    ON DELETE RESTRICT;
```

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,91 @@
...
+
+ALTER TABLE conditions_context_groupers 
+    DROP CONSTRAINT conditions_context_groupers_condition_id_fkey;
+
+ALTER TABLE conditions_context_groupers 
+    ADD CONSTRAINT conditions_context_groupers_condition_id_fkey 
+    FOREIGN KEY (condition_id) 
+    REFERENCES conditions (id) 
+    ON UPDATE CASCADE 
+    ON DELETE CASCADE;
```
jakewheeler: Maybe `RESTRICT` here as well? 🤔 
```suggestion
    ON DELETE RESTRICT;
```

refiner/migrations/20260604134336_add_codeset_tables.sql
fzhao99: hmm good point! we are duplicating that data as is.

think the main complication removing version is changing `UNIQUE(system_id, version, value)` to `UNIQUE(system_id, value)`. Assuming we want codes to keep the constraint as well as remaining many-to-many / not include a foreign key condition column, any codes that show up in multiple valuesets across condition versions would need to be tracked / managed accordingly during seeding to respect the uniqueness constraint. Certainly doable, but may complicate the seeding script a bit to maintain the cross-references rather than just inserting the codes per condition version. 

Thoughts on the tradeoffs here? Worth mentioning that we'll probably will probably have to do this anyways once we get to the full codes refactor, since there are almost definitely codes within the condition valuesets that show up across multiple conditions.

refiner/migrations/20260604134336_add_codeset_tables.sql
fzhao99: Think this is the right delete pattern! 

Just to confirm:
- it looks like `conditions_context_groupers` has `ON DELETE CASCADE` as is, so I'll change it to `RESTRICT` here but leave `CASCADE` on the down migration for the sake of provenance
- For `configurations_conditions` the foreign key for conditions didn't have a delete action, so I'll just drop and remake the foreign key reference in the down migration.

refiner/migrations/20260604134336_add_codeset_tables.sql
jakewheeler: That's my bad. I was thinking about this wrong. You're right that codes should not be linked to a condition. The concept that has a `version` is actually the TES.

It would be a bit more work on top of what you've already done, but it might make sense to have a `tes` table that has a row for each TES version. The `codes` and `conditions` would then have an associated `tes_id`. 

What do you think? We don't necessarily need to do this in this PR if we don't want to.

fzhao99: yep, think we should add a separate table for it. Will definitely be useful to have for all the forthcoming TES acceptance related work. 

Might be a fair chunk of work to update things in all the places, so I'll make a ticket for it and we can address it in a separate PR to change that column/related references to a TES foreign key

fzhao99: Ticket made here: https://app.zenhub.com/workspaces/dibbs-ecr-refiner-67ddd053d70b9f000ffbb542/issues/gh/cdcgov/dibbs-ecr-refiner/1347

jakewheeler: Thanks!

refiner/migrations/20260604134336_add_codeset_tables.sql
jakewheeler: I'd be interested in @robertmitchellv's thoughts on whether or not a different approach makes sense right now but I am questioning my original comment.

I think if we stick with the current implementation then the original `conditions_codes` I initially suggested actually does make sense since that best describes what this table is doing. We don't currently differentiate "child RSGs" from "codes" in any way at the data level.

refiner/migrations/20260604134336_add_codeset_tables.sql
```diff
@@ -0,0 +1,89 @@
...
+    UNIQUE (system_id, version, value),
+    CONSTRAINT fk_codes_system_id_fkey 
+        FOREIGN KEY (system_id) 
+        REFERENCES systems (id) 
+        ON DELETE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS condition_child_rsg_codes (
+    code_id UUID REFERENCES codes(id) ON DELETE CASCADE, 
+    condition_id UUID REFERENCES conditions(id) ON DELETE CASCADE,
```
jakewheeler, line 32: Not a big deal either way but I'd like `condition_id` to be above `code_id` since the table name starts with `conditions_` 😅 if you wouldn't mind making this adjustment.

refiner/migrations/20260604134336_add_codeset_tables.sql
fzhao99: Just to understand further, would the idea be to store all the code joins in one table and be able to differentiate "valueset codes" vs "RSG codes" somehow? Not immediately clear to me as is how we'd be able to tell them apart when going to query the data, but we could do something like add another column to the join table along the lines of "is RSG" or something like that.

Or would we make a separate non-RSG join table and name it something else?

jakewheeler: > Just to understand further, would the idea be to store all the code joins in one table and be able to differentiate "valueset codes" vs "RSG codes" somehow?

No, I'm thinking that these would be separate. I think I'm getting confused by the naming of these tables / what we're referring to as "codes" 😖  I appreciate you mentioning "valueset" vs. "RSG" codes explicitly.

If this PR will only handle RSG codes and not do anything with valueset codes, I think it'd make sense to rename these tables:
- `codes` -> `rsg_codes`
- `condition_child_rsg_codes` -> `conditions_rsg_codes`

I would expect in the future that valueset codes would go into their own table, similar to custom codes. Let me know if this makes more sense / if I'm still missing something. And apologies for my confusion 😸

refiner/scripts/seeding/load_static_data.py
```diff
@@ -65,6 +75,76 @@ class ProcessedCondition(TypedDict):
...
+
+
+type ConditionUniqueIndex = tuple[VsCanonicalUrl, VsVersion]
+
+type ConditionToCodeRelationshipIndex = dict[
+    ConditionUniqueIndex, ConditionToCodeRelationshipTrace
+]
+
+
+def _build_rsg_codes(
```
jakewheeler, line 101: Is it intentional that `codes.id` and `conditions.id` changes for all codes and conditions every time we run the data loading script? That behavior doesn't seem ideal to me. If the entity already exists in the DB I'd expect its ID to never change.

fzhao99: updated things to follow the condition upsert pattern / do a follow up fetch for ID when upserted to keep track of the relationships

refiner/migrations/20260604134336_add_codeset_tables.sql
fzhao99: Kinda building this join on the fly, so no worries about the confusion! I don't feel too strongly about the below opinion, but think I'm slightly leaning towards leaving codes the same and modeling differences in their contexts (RSG vs valueset vs anything else) in separate join tables. 

Just to make sure we're talking about the same thing, think our options are

1. one codes table, two joins (one represent valueset relationships, one representing RSG relationships) to conditions - which what the PR is setting up currently
1. two codes tables (one for valueset, one for RSG), two joins to conditions - which is what I think you're proposing
1. one codes table, one join (with a column for valueset or RSG in the join table) - a hypothetical third option

The main worry I have if we split the tables out by context in 2) is if we ever make changes to the metadata we're tracking from the TES / how we enrich them from other sources, splitting them out `codes` into multiple tables would require us to keep the schemas in sync across the different tables. From a data perspective, since all the codes are coming from the TES / thus have the same shape, there I don't envision there being a difference at the data level between the RSG code table and the valueset code table (though maybe you have a different perspective on this). Unlike for custom codes, where the shape of codes / their relationship to the parent entity is different, there's no obvious difference to me at the data level that necessitates splitting out the tables.

Since any differences would be at the relationship level between conditions and the codes in question, I'm leaning slightly towards storing the codes in one table and modeling the differences in their usage at the join table level. 

We can also talk about this at eng time since I'd want to have @robertmitchellv weigh in on this too

fzhao99: Decision in eng time was to go with option 1) and revisit if needed.

jakewheeler (approved):
I think we're good to call this work for now per our team discussion. Thanks for your work on this 🎉

PR #1295, opened 2026-06-01
fix: `created_by` is typed as `UUID` but should be `str`

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR addresses an issue where a model's attribute was typed as `UUID` but is actually being used as a string.

Here's an example of the warning we were seeing from Pydantic
```sh
UserWarning: Pydantic serializer warnings:
server-1      |   PydanticSerializationUnexpectedValue(Expected `uuid` - serialized value may not be as expected [field_name='created_by', input_value='refiner', input_type=str])
```

## 🧪 How to test

You should no longer be seeing this warning coming from the app when navigating to any of the configuration pages (build/test/activate).


rogeruiz (approved):
lgtm 🌈

PR #1314, opened 2026-06-02
feat(ui): Add LOINC to sections & headers sticky

### LINKED ISSUE #1277: [TASK] Section table improvements
## 🚧 What needs to be done

Need to freeze table headers for sections table and add section LOINCs 

## 🤔 Why it needs to be done

It's difficult to track what the columns are without headers, and sections can be named differently by different EHRs, so having a LOINC is necessary to identify which section is which.

## 🏗️ Acceptance Criteria

- [ ] Freeze section table headers (they should still show when scrolling down)
- [ ] Add LOINCs for each section - [design with LOINCs in each section row](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=10487-30695&t=DoGowzckeXsnQK3s-1) 

## ⚙️ Technical Notes

- ...

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Makes the table headers for sections sticky by adding a scrollable wrapper. Also
the section code display is now showing on all sections not just custom ones.

## 🔗 Related Issue

- Closes #1277

## ✅ Acceptance Criteria

- [ ] Freeze section table headers (they should still show when scrolling down)
- [ ] Add LOINCs for each section - [design with LOINCs in each section
      row](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=10487-30695&t=DoGowzckeXsnQK3s-1)

## 🧪 How to test

Load up the application locally. Navigate to a configuration's Sections view and
scroll to observe the sticky headers and the LOINC codes should now be visible
for all sections not just custom ones.


client/src/pages/Configurations/ConfigBuild/Sections/index.tsx
```diff
@@ -97,90 +97,92 @@ export function Sections({
...
-          {sectionProcessing.map((section) => (
-            <tr key={section.code} className="text-gray-cool-60">
-              <td>
-                <div className="flex justify-center p-8">
-                  <IncludeCheckbox
-                    configurationId={configurationId}
-                    currentSection={section}
-                    sections={sectionProcessing}
-                    disabled={disabled || isDisabledSection(section.code)}
+      <div className="max-h-[calc(100vh-400px)] overflow-y-auto">
```
fzhao99: Doing things this way doesn't use the full height at smaller screen sizes.

https://github.com/user-attachments/assets/0f79e97e-438b-4a40-993a-319fb21fd26d

An alternative is to remove the div and instead make the individual `th` elements sticky and set the relative positioning using something like `-top-6 z-10 bg-white`

client/src/pages/Configurations/ConfigBuild/Sections/index.tsx
```diff
@@ -130,57 +130,61 @@ export function Sections({
...
-                  setSelectedSection={() => onSelectedSection(section)}
-                />
-              </td>
-              <td>
-                {section.include ? (
-                  <div className="flex justify-center">
-                    <RefineSwitch
+      </table>
+      <div className="min-h-0 flex-1 overflow-y-scroll">
+        <table className="w-full table-fixed">
```
fzhao99, line 102: breaking things into two tables like this will cause accessibility issues, since the bottom table doesn't have any semantic labeling and will get lost for screenreader users without associated header columns. 

Is there a concern for keeping things in one semantic table and using `sticky` here like we were previously? Otherwise, this will come back to bite us when we need to submit accessibility scans at the next CDC review

rogeruiz: Yeah fair. I did think about this but wanted to see if this tradeoff would work well. We shouldn't degrade a11y for the sake of design. I have another option that should be more accessible using CSS grid or possibly using a different technique to render a tooltip. The breaking apart was done to allow the tooltip to escape the `overflow-hidden` on the table.

client/src/components/Tooltip/index.tsx
```diff
@@ -26,12 +21,57 @@ const arrowClasses: Record<TooltipPosition, string> = {
...
 
+function getTooltipStyle(
+  position: TooltipPosition,
+  rect: DOMRect | null
+): React.CSSProperties {
+  if (!rect) return { position: 'fixed', visibility: 'hidden' };
+
+  const gap = 8;
+
+  switch (position) {
```
fzhao99: think we can make this code more readable if we 

1. Tweak this interface slightly to do the client rect calculation off the triggering button element inside the style function (and thus getting rid of extra trigger ref that I don't think we need)
1. Split out the procedural style calculations away from the declarative code selecting the actual styles

I put up some suggestions [here](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1342) for you to review. Feel free to merge if you agree

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -299,7 +299,7 @@ function Builder({
                     aria-current={tableView === 'custom' ? 'true' : undefined}
                   >
                     <span>Custom codes</span>
-                    <span>{custom_codes.codes.length.toLocaleString()}</span>
+                    <span>{custom_codes.codes.length?.toLocaleString()}</span>
```
fzhao99, line 302: don't know if we need these optional chainers since Orval is enforcing the type contract for us? 

If I'm missing something though, we should add a fallback string in the event that the value isn't defined

fzhao99 (commented):
left a couple suggestions on how to make the style calculation code more readable.

happy to give a fast approval if the suggestions look good to you

client/src/pages/Configurations/ConfigBuild/index.tsx
rogeruiz: Yeah I was getting a console error on first load without these.

client/src/components/Tooltip/index.tsx
rogeruiz: I took a look at #1342 and think that there is a performance regression in your PR, @fzhao99. Specifically, the `getBoundingClientRect()` is now called during the render cycle via `getTooltipStyle` on line 60 of `client/src/components/Tooltip/index.tsx`. I appreciate the clean up of the switch statement to make the code a bit DRYer but I don't quite agree with the changes in their current state. Do you mind if I add some changes to #1342 to get it in a good state? You can review it before I merge it into this one.

fzhao99 (approved):

PR #1316, opened 2026-06-03
test: Pydantic warnings are errors when running the app locally

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the app to throw errors (local only) whenever Pydantic `UserWarning`s are reported, which in our case typically amount to `PydanticSerializationUnexpectedValue` warnings. There is no real reason we'd ever want these warnings to slip through during local dev / when testing. These should be much more obvious now since they'll break the app 😅 

## 🧪 How to test

A simple way to see this in action is to update [this line](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/api/auth/handlers.py#L270) to be:
```python
to_render={"most_recent_app_update": should_show_app_update}
```

We're using the string representation of the enum here, which Pydantic will warn about. This should cause an error on this branch.

### DIFF
--- a/refiner/app/main.py
+++ b/refiner/app/main.py
@@ -31,6 +31,7 @@ from .core.config import ENVIRONMENT
 from .db.pool import AsyncDatabaseConnection, get_db
 from .services.logger import get_logger, set_request_id
 
+# Pydantic warnings will be reported as errors in local dev / testing
 if ENVIRONMENT["ENV"] == "local":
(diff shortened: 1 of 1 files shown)


fzhao99 (approved):
🫡

And 🫡🫡🫡 to whoever's the lucky soul to fix all the errors

Issue #1016, opened 2026-03-30
[EPIC] Reconstructed narratives


kplaneaux:
## 👤💭 User Story

As a user, I can choose whether to replace the narrative with one reconstructed from refined coded data, so that I can control what kind of information comes through in the narrative

## 🏗️ Sections to Implement

First four:
- [ ] Results
- [ ] Problems
- [ ] Medications Administered
- [ ] Immunizations

Next:
- [ ] Admission Diagnosis
- [ ] Admission Medications
- [ ] Discharge Diagnosis
- [ ] Encounters
- [ ] Medications
- [ ] Past Medical History
- [ ] Plan of Treatment
- [ ] Pregnancy
- [ ] Procedures
- [ ] Social History
- [ ] Vital Signs
- [ ] Custom sections

Not applicable:
- Emergency Outbreak Information
- Reportability Response Information
- Chief Complaint
- Reason for Visit
- History of Present Illness
- Review of Systems

PR #1318, opened 2026-06-03
feat(ui): Convert Narrative `<Switch/>` to `<Dropdown/>`

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch updates the narrative data toggle switch to a dropdown select
component with two options ("Keep original" and "Remove"). This only addresses
the first part of the narrative reconstruction dropdown feature. The change
replaces the `<NarrativeSwitch/>` component with a new `<NarrativeSelect/>`
component that uses the existing `<Select/>` component. I'm also preserving the
blanking pattern for excluded rows; if `include: false` then no controls are shown.

> ❗ **Important**
>
> This patch is only starting the narrative reconstruction dropdown feature. It
> does not include the new "Reconstruct" option, cross-field validation logic,
> nor backend enum migration. Those will be addressed in subsequent patches.

## 🔗 Related Issue

- #1016
- #1195

## ✅ Acceptance Criteria

- [x] Convert narrative data options to a drop-down
- [ ] ~~Reconstruct option appears in the drop-down when available for that
      section~~
- [ ] ~~Reconstruction option grayed out unless "refine" is selected for coded
      data~~
- [x] ~~Updated table column names (coded data, narrative data)~~
- [x] ~~Updated coded data options copy (keep original, refine)~~
- [x] ~~Updated tooltip copy~~
- [ ] ~~If user has "reconstruct" for the narrative and tries to change coded
      data to "keep original", error appears under the toggle~~
- [x] Default settings are "refine" for coded data and "keep original" for
      narrative
- [x] Should retain pattern of blanking out excluded rows (not showing options
      for coded data or refined data)

## 🧪 How to test

1. Run `just dev up` to bring up the application locally
1. Navigate to the configuration's Build step and inspect the Sections table.
   Confirm that the narrative column now displays a dropdown instead of a toggle
   switch.
1. Verify that "Keep original" and "Remove" are both working as expected.
1. Run automated checks with `just client run lint fmt check test e2e`

## ℹ️ Additional Information

> [!NOTE]
> The work related to #1195 will be applied in four different patches/PRs
> continuing with this one. Each patch will be as independent as possible and rely
> only on prior patches as strictly necessary. This patch relies on the changes
> introduced in #1317.

### DIFF
--- a/client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
+++ b/client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
@@ -158,12 +158,7 @@ describe('Configuration sections', () => {
     const disabledSectionName = screen.getByText('Disabled section');
     const row = disabledSectionName.closest('tr');
     expect(row).not.toBeNull();
-
-    if (!row) {
-      throw new Error('Could not find table row for "Disabled section"');
-    }
-
-    const select = within(row).queryByRole('combobox');
+    const select = row && within(row).queryByRole('combobox');
     expect(select).toBeInTheDocument();
     expect(select).toBeDisabled();
   });
--- a/client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
+++ b/client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
@@ -45,6 +45,19 @@ const sections: DbConfigurationSectionProcessing[] = [
   },
 ];
 
+const sectionsWithDisabled: DbConfigurationSectionProcessing[] = [
+  ...sections,
+  {
+    name: 'Disabled section',
+    action: 'retain',
+    include: true,
+    code: '83910-0',
+    narrative: true,
+    versions: ['3.1'],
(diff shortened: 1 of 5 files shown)


client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
```diff
@@ -66,43 +78,94 @@ describe('Configuration sections', () => {
...
+
+    const disabledSectionName = screen.getByText('Disabled section');
+    const row = disabledSectionName.closest('tr');
+    expect(row).not.toBeNull();
+
+    if (!row) {
+      throw new Error('Could not find table row for "Disabled section"');
+    }
+
+    const select = within(row).queryByRole('combobox');
```
fzhao99: think we can just check for existence here in the assignment 

```suggestion
    const select = row && within(row).queryByRole('combobox');
```

fzhao99 (approved):
one minor suggestion in a test file, but otherwise lgtm!

PR #1321, opened 2026-06-04
feat: Code set improvements

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR:

- Updates the code set table's header from `Condition` to `Display name`
- Updates the configuration CSV export
  - Removes the restriction that prevents users from downloading a configuration export when config is locked by another user
  - Adds a `Condition` column that includes the condition grouper's name (unless it's a custom code, then it's blank)
  - Fixed an issue where the `system_key` value for custom codes was being added to the document instead of system name
  - Refactoring / cleanup (add functions, remove unneeded comments, etc.)
- Additions / improvements to tests 

## 🔗 Related Issue

Fixes #1278 
- #1278 

## 🧪 How to test

- Check that you now see `Display name` instead of `Condition` when looking at the code set table
- Verify that the configuration export CSV downloads and appears as expected

### DIFF
--- a/client/e2e/configuration.spec.ts
+++ b/client/e2e/configuration.spec.ts
@@ -10,6 +10,23 @@ test.describe('Configuration detail flow', () => {
     await deleteAllConfigurations();
   });
 
+  test('Validate code set table appearance', async ({
+    page,
+    configurationsPage,
+    configurationPage,
+  }) => {
+    const condition = 'Anotia';
+    await configurationsPage.createConfiguration(condition);
+    await configurationPage.goToBuildTab();
+    await page.getByLabel('View TES code set information for Anotia').click();
+
+    await expect(page.getByRole('columnheader')).toHaveText([
+      'Code',
+      'Code system',
+      'Display name',
+    ]);
+  });
+
   test('Check code set status and individual grouper statuses', async ({
     page,
     configurationsPage,
--- a/refiner/app/api/v1/configurations/exports.py
+++ b/refiner/app/api/v1/configurations/exports.py
@@ -65,6 +65,7 @@ async def get_configuration_export(
         writer.writerow(
             [
                 "Code Type",
+                "Condition",
                 "Code System",
                 "Code",
                 "Display Name",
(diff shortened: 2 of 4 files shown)


refiner/app/api/v1/configurations/exports.py
```diff
@@ -33,77 +35,76 @@ async def get_configuration_export(
...
-                "Code System",
-                "Code",
-                "Display Name",
-            ]
+            ["Code Type", "Condition", "Code System", "Code", "Display Name"]
         )
-        for cond in included_conditions:
+
+        for cond in conditions:
             codes = await get_condition_codes_by_condition_id_db(id=cond.id, db=db)
```
jakewheeler, line 85: I considered doing something about this here but realistically you'd have to add a ton of code sets for it to have a significantly negative impact on performance. It's not ideal but doesn't seem likely to cause any immediate issues with the way things work today.

client/e2e/configuration.spec.ts
```diff
@@ -10,6 +10,23 @@ test.describe('Configuration detail flow', () => {
     await deleteAllConfigurations();
   });
 
+  test('Validate code set table appearance', async ({
```
rogeruiz, line 13: Thanks for this!

refiner/tests/integration/test_export.py
```diff
@@ -30,10 +38,145 @@ async def test_export_returns_csv_with_correct_headers(
...
+        assert code_systems_in_csv <= expected_systems
+
+    async def test_export_custom_codes_have_blank_code_system(
+        self,
+        setup,
+        authed_client,
+        get_condition_id,
+        create_config,
+        add_custom_code,
+        db_pool,
```
rogeruiz: We're not using this `db_pool` here? Do we want to remove it or leave it here to avoid changing call signatures?

refiner/app/api/v1/configurations/exports.py
```diff
@@ -33,77 +35,76 @@ async def get_configuration_export(
...
+
+        for cc in config.custom_codes or []:
             writer.writerow(
                 [
                     "Custom code",
-                    custom.system_key or "",
-                    custom.code or "",
-                    custom.name or "",
+                    "",
+                    code_systems[cc.system_key].display_name,
```
rogeruiz: Should we be safer about this or just allow for the 500 here? I can't imagine that the code system wouldn't have a display name since we default to Other. But maybe some changes in the future could cause this to trip us up? Could do something like:

```py
cs = code_systems.get(cc.system_key)
cs_display = cs.display_name if cs else cc.system_key
```

rogeruiz (approved):
lgtm 🌈 

Minor nit here and there but over all this looks great. Thanks for adding a new E2E test too.

refiner/app/api/v1/configurations/exports.py
rogeruiz: Maybe we can add a `TODO:` comment here to check this in the future? Maybe something like:

```py
# TODO: This loop issues a separate DB query per condition to fetch codes
# (N+1). If configs start including many many more code sets, consolidate into
# a single query that fetches codes for all condition IDs at once.
```

refiner/tests/integration/test_export.py
jakewheeler: Good catch 🫡  I think it just stuck around accidentally when I was iterating on the tests. I'll delete it.

refiner/app/api/v1/configurations/exports.py
jakewheeler: That's a good point 🤔 I think I'll modify this a bit.

PR #1320, opened 2026-06-04
 feat(api,ui): Narrative ENUM migration (`retain`, `remove`, `reconstruct`) 

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch migrates the narrative field from a boolean type to a three-value
enum across the API and client. This prepares the API and the client for the
work being done for CDCgov/dibbs-ecr-refiner#1195. This work is independent of
CDCgov/dibbs-ecr-refiner#1317 and CDCgov/dibbs-ecr-refiner#1318 but closely
related to them. The latter needs to be safely integrated with this patch to
ensure the changes don't conflict with one another.

> ❗ **Important**
>
> This patch is continuing the work from #1317 and #1318. It's as independent as
> possible from those changes but should be carefully integrated when #1318 is
> approved and merged.

## 🔗 Related Issue

- #1016
- #1195

## ✅ Acceptance Criteria

- [x] ~~Convert narrative data options to a drop-down~~
- [ ] ~~Reconstruct option appears in the drop-down when available for that
      section~~
- [ ] ~~Reconstruction option grayed out unless "refine" is selected for coded
      data~~
- [x] ~~Updated table column names (coded data, narrative data)~~
- [x] ~~Updated coded data options copy (keep original, refine)~~
- [x] ~~Updated tooltip copy~~
- [ ] ~~If user has "reconstruct" for the narrative and tries to change coded
      data to "keep original", error appears under the toggle~~
- [x] ~~Default settings are "refine" for coded data and "keep original" for
      narrative~~
- [x] ~~Should retain pattern of blanking out excluded rows (not showing options
      for coded data or refined data)~~

## 🧪 How to test

1. Start the dev server: `just dev up`
1. Manually test the changes in the application
   - Open the app and navigate to a configuration's section settings
   - Toggle the narrative switch for various sections
   - Verify the API requests send narrative: "retain" or narrative: "remove" (not true/false)
   - Check that the UI correctly reflects the current narrative state
   - Test creating a new configuration and verify default narrative values
1. Migrate the database: `just migrate local`
1. Verify the narrative column has a type of section_narative (enum) instead of
   boolean: `just db query "\d configurations_sections"`
1. Run server tests: `just server test`
1. Run client tests: `just client test`

## ℹ️ Additional Information

> [!NOTE]
> The work related to #1195 will be applied in four different patches/PRs
> continuing with this one. Each patch will be as independent as possible and rely
> only on prior patches as strictly necessary. This patch should be independent
> of the others.

PR #1322, opened 2026-06-04
feat: Add reconstruct option to narrative dropdown with cross-field validation

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch adds "Reconstruct" as a third narrative option to the dropdown built
in #1318 and the updated API from #1320. This option is conditionally visible
for sections with coded entry match rules and disabled when coded data is set to
"Keep original". There's also cross-field validation that prevents switching
coded data to "Keep original" while narrative is set to "Reconstruct", with
an error state lifted to the parent component so errors can clear across all
rows on any user interaction. The `NarrativeSelect` component now uses
string enum values (`retain`/`remove`/`refine`) directly.

> This patch is continuing the work from #1317, #1318, and #1320. And should be
> merged into #1320 to ensure no conflicts happen in latter PR merges.

## 🔗 Related Issue

- #1016
- #1195

## ✅ Acceptance Criteria

- [x] ~~Convert narrative data options to a drop-down~~
- [x] Reconstruct option appears in the drop-down when available for that
      section
- [x] Reconstruction option grayed out unless "refine" is selected for coded
      data
- [x] ~~Updated table column names (coded data, narrative data)~~
- [x] ~~Updated coded data options copy (keep original, refine)~~
- [x] ~~Updated tooltip copy~~
- [x] If user has "reconstruct" for the narrative and tries to change coded
      data to "keep original", error appears under the toggle
- [x] ~~Default settings are "refine" for coded data and "keep original" for
      narrative~~
- [x] ~~Should retain pattern of blanking out excluded rows (not showing options
      for coded data or refined data)~~

## 🧪 How to test

1. Start the dev server: `just dev up`
1. Manually test the changes in the application
   - Open the app and navigate to a configuration's section settings
   - Confirm "Reconstruct" appears in the Narrative data dropdown for coded sections
   - Confirm "Reconstruct" does not appear for narrative-only sections (e.g.
     Chief Compliant)
   - Toggle coded data to "Keep original" and confirm "Reconstruct" is grayed out
   - Select "Reconstruct" for a section, then try toggling coded data to "Keep
     original" and confirm the error message appears
   - Interact with a different row and confirm the error clears
1. Run server tests: `just server test`
1. Run client tests: `just client test`
1. Run e2e tests: `just client run e2e`

## ℹ️ Additional Information

> [!NOTE]
> The work related to #1195 will be applied in four different patches/PRs
> continuing with this one. Each patch will be as independent as possible and rely
> only on prior patches as strictly necessary. This patch builds upon #1320 and
> #1318.


refiner/schema.sql
```diff
@@ -270,20 +270,6 @@ CREATE TABLE public.custom_codes (
...
-CREATE TABLE public.custom_codes (
-    id uuid DEFAULT gen_random_uuid() NOT NULL,
-    name text NOT NULL,
-    value text NOT NULL,
-    system_id uuid NOT NULL,
-    created_at timestamp with time zone DEFAULT now() NOT NULL,
-    updated_at timestamp with time zone DEFAULT now() NOT NULL
-);
-
-
```
rogeruiz, line 286: I thought this was weird. Looks like on `main` we have duplicated table creations so this is me removing it. Here's a screenshot of the change on `main` as of the time of this comment.

<details open>
<summary>📸  Screenshot(s)</summary>

Image

</details>

cc: @fzhao99

client/src/pages/Configurations/ConfigBuild/Sections/index.tsx
```diff
@@ -37,6 +37,10 @@ export function Sections({
   const [selectedSection, setSelectedSection] =
     useState<DbConfigurationSectionProcessing | null>(null);
   const [isOpen, setIsOpen] = useState(false);
+  const [errorSectionCode, setErrorSectionCode] = useState<string | null>(null);
+
+  const clearError = () => setErrorSectionCode(null);
+  const setError = (code: string) => setErrorSectionCode(code);
```
jakewheeler: This could be a really good candidate for a context so that we can manage the error state at the top level and make use of it inside any of the children. We could make use of a hook instead of passing the props along to all of the children that need them.

I'm testing this out locally which seems to work well
```tsx
// ConfigBuild/Sections/SectionErrorContext.tsx
import { createContext, useContext, useState } from 'react';

interface SectionErrorContextValue {
  errorSectionCode: string | null;
  setError: (code: string) => void;
  clearError: () => void;
}

const SectionErrorContext = createContext<SectionErrorContextValue | null>(
  null
);

export function SectionErrorProvider({
  children,
}: {
  children: React.ReactNode;
}) {
  const [errorSectionCode, setErrorSectionCode] = useState<string | null>(null);

  return (
    <SectionErrorContext.Provider
      value={{
        errorSectionCode,
        setError: (code) => setErrorSectionCode(code),
        clearError: () => setErrorSectionCode(null),
      }}
    >
      {children}
    </SectionErrorContext.Provider>
  );
}

export function useSectionError() {
  const ctx = useContext(SectionErrorContext);
  if (!ctx)
    throw new Error(
      'useSectionError must be used within a SectionErrorProvider'
    );
  return ctx;
}
```

In `ConfigBuild/Sections/index.tsx` I wrapped the top-level section component:
```tsx
  return (
    <SectionErrorProvider>
      <section className="flex min-h-0 w-full flex-1 flex-col gap-6">
        <div className="flex flex-col gap-2">
       ...
    </SectionErrorProvider>
```

And then I deleted the error props on each component and can do something like this using the context hook instead:
```tsx
function RefineSwitch({
  currentSection,
  configurationId,
  disabled,
  isNarrativeOnly,
}: SelectionToggleProps & {
  isNarrativeOnly: boolean;
}) {
  const updateSection = useSectionUpdater(configurationId);
  const { clearError, setError, errorSectionCode } = useSectionError();

  const showError = errorSectionCode === currentSection.code;
  ...

  const handleSwitchChange = (checked: boolean) => {
    // TODO: This validation should eventually be enforced by backend API as well
    if (!checked && currentSection.narrative === 'reconstruct') {
      setError(currentSection.code);
      return;
    }
    clearError();
    updateSection(currentSection, {
      action: checked ? DbSectionAction.refine : DbSectionAction.retain,
    });
  };
...
...
```
Hopefully that example helps. Let me know what you think!

refiner/schema.sql
jakewheeler: Was this from a conflict? If you regenerate `schema.sql` does it have this listed twice?

client/src/pages/Configurations/ConfigBuild/Sections/Sections.test.tsx
```diff
@@ -58,6 +58,40 @@ const sectionsWithDisabled: DbConfigurationSectionProcessing[] = [
   },
```
jakewheeler, line 58: Could be worth adding to [this e2e test ](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/client/e2e/configuration.spec.ts#L319)as well to account for the new narrative box selections we have available.

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+  test('should manage "Reconstruct" option availability and state', async ({
+    page,
+  }) => {
+    // 1. Standard Section - 3 options
+    const standardRow = page
+      .locator('tr')
+      .filter({ hasText: 'Admission Diagnosis' });
+    const standardSelect = standardRow.getByRole('combobox');
+
+    const standardOptions = standardSelect.locator('option');
```
jakewheeler:
```suggestion
    const standardOptions = standardSelect.getByRole('option');
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    await expect(
+      standardOptions.filter({ hasText: 'Reconstruct' })
+    ).toBeAttached();
+
+    // 2. Narrative-Only Section - 2 options
+    const narrativeOnlyRow = page
+      .locator('tr')
+      .filter({ hasText: 'Chief Complaint' });
+    const narrativeOnlySelect = narrativeOnlyRow.getByRole('combobox');
+    const narrativeOnlyOptions = narrativeOnlySelect.locator('option');
```
jakewheeler:
```suggestion
    const narrativeOnlyOptions = narrativeOnlySelect.getByRole('option');
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+      .getByRole('combobox')
+      .locator('option')
+      .filter({ hasText: 'Reconstruct' });
+    await expect(reconstructOption).toBeDisabled();
+
+    // 4. Enabled when Coded Data is 'Refine'
+    await codedDataSwitch.click(); // Set to 'Refine' (ON)
+    const reconstructOptionEnabled = sectionRow
+      .getByRole('combobox')
+      .locator('option')
```
jakewheeler:
```suggestion
      .getByRole('option')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+      await codedDataSwitch.click();
+      await codedDataSwitch.click();
+    }
+    await expect(errorAlert).toBeVisible();
+
+    // 4. Persistence via Internal Click
+    await page
+      .locator('[data-error-trigger]')
+      .first()
+      .click({ position: { x: 2, y: 2 } });
```
jakewheeler: I am not having any issue getting the test to pass without `position`.
```suggestion
      .click();
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    await page.getByRole('heading', { level: 1 }).first().click();
+    await expect(errorAlert).not.toBeVisible();
+
+    // 3. Re-trigger Error
+    if (await codedDataSwitch.isChecked()) {
+      await codedDataSwitch.click();
+    } else {
+      await codedDataSwitch.click();
+      await codedDataSwitch.click();
+    }
```
jakewheeler: Does this work instead?

```suggestion
    await expect(codedDataSwitch).toBeChecked();
    await codedDataSwitch.click();
    await expect(errorAlert).toBeVisible();
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    const codedDataSwitch = sectionRow.getByRole('switch');
+
+    // Setup: Set Narrative to 'Reconstruct'
+    await narrativeSelect.selectOption('reconstruct');
+    await expect(narrativeSelect).toHaveValue('reconstruct');
+
+    // 1. Trigger Error: Switch to 'Keep original'
+    if (await codedDataSwitch.isChecked()) {
+      await codedDataSwitch.click();
+    }
```
jakewheeler:
```suggestion
    await expect(codedDataSwitch).toBeChecked();
    await codedDataSwitch.click();
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
     });
   });
 });
+
+test.describe('Sections Validation and Error Lifecycle', () => {
```
jakewheeler, line 578: Nice tests 👍

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    const reconstructOptionEnabled = sectionRow
+      .getByRole('combobox')
+      .locator('option')
+      .filter({ hasText: 'Reconstruct' });
+    await expect(reconstructOptionEnabled).toBeEnabled();
+  });
+
+  test('should handle the validation error lifecycle', async ({ page }) => {
+    const sectionRow = page
+      .locator('tr')
```
jakewheeler:
```suggestion
      .getByRole('row')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+      .filter({ hasText: 'Admission Diagnosis' });
+    const codedDataSwitch = sectionRow.getByRole('switch');
+
+    if (await codedDataSwitch.isChecked()) {
+      await codedDataSwitch.click();
+    }
+
+    const reconstructOption = sectionRow
+      .getByRole('combobox')
+      .locator('option')
```
jakewheeler:
```suggestion
      .getByRole('option')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    const narrativeOnlySelect = narrativeOnlyRow.getByRole('combobox');
+    const narrativeOnlyOptions = narrativeOnlySelect.locator('option');
+    await expect(narrativeOnlyOptions).toHaveCount(2);
+    await expect(
+      narrativeOnlyOptions.filter({ hasText: 'Reconstruct' })
+    ).not.toBeAttached();
+
+    // 3. Disabled when Coded Data is 'Keep original'
+    const sectionRow = page
+      .locator('tr')
```
jakewheeler:
```suggestion
      .getByRole('row')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+
+    const standardOptions = standardSelect.locator('option');
+    await expect(standardOptions).toHaveCount(3);
+    await expect(
+      standardOptions.filter({ hasText: 'Reconstruct' })
+    ).toBeAttached();
+
+    // 2. Narrative-Only Section - 2 options
+    const narrativeOnlyRow = page
+      .locator('tr')
```
jakewheeler:
```suggestion
      .getByRole('row')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+    await configurationsPage.createConfiguration(condition);
+    await page.getByRole('button', { name: 'Sections' }).click();
+  });
+
+  test('should manage "Reconstruct" option availability and state', async ({
+    page,
+  }) => {
+    // 1. Standard Section - 3 options
+    const standardRow = page
+      .locator('tr')
```
jakewheeler:
```suggestion
      .getByRole('row')
```

client/e2e/configuration.spec.ts
```diff
@@ -572,3 +572,110 @@ test.describe('Configuration detail flow', () => {
...
+
+    // 3. Disabled when Coded Data is 'Keep original'
+    const sectionRow = page
+      .locator('tr')
+      .filter({ hasText: 'Admission Diagnosis' });
+    const codedDataSwitch = sectionRow.getByRole('switch');
+
+    if (await codedDataSwitch.isChecked()) {
+      await codedDataSwitch.click();
+    }
```
jakewheeler:
```suggestion
    await expect(codedDataSwitch).toBeChecked();
    await codedDataSwitch.click();
```

client/src/pages/Configurations/ConfigBuild/Sections/SectionErrorContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+        setError: (code) => setErrorSectionCode(code),
+        clearError: () => setErrorSectionCode(null),
+      }}
+    >
+      {children}
+    </SectionErrorContext.Provider>
+  );
+}
+
+export function useSectionError() {
```
jakewheeler: Can we add this above the line? I noticed ESLint is flagging us here.

```suggestion
// eslint-disable-next-line react-refresh/only-export-components
export function useSectionError() {
```

client/e2e/configuration.spec.ts
jakewheeler: Can we add a few a11y checks throughout these? After the error message becomes visible especially would be a good place to add a check.

```ts
```

client/e2e/configuration.spec.ts
rogeruiz: hmm. Changing these from `locator` to `getRoleBy` gives me an error where the `getRoleBy` method is not found. I think it's because I'm not mapping to ARIA roles for targeting these elements. According to the Playwright documentation on Locators (https://playwright.dev/docs/locators), getByRole is specifically for finding elements by their ARIA role (https://www.w3.org/TR/wai-aria-1.2/), whereas locator() is a general-purpose tool for CSS and XPath selectors.

jakewheeler: Can you double check you're using `getByRole` and not `getRoleBy` as you mentioned in your message? 😅

client/e2e/configuration.spec.ts
rogeruiz: The position seems to need to be bigger than the row's width and height. I have this test here so that I can make sure things don't get cleared when the user clicks in the same row as the error. Having it set to zero would just test to see if anyone clicks the containing div in general.

client/e2e/configuration.spec.ts
rogeruiz: Yeah sorry. that was a typo, but i'm still getting an error regarding these changes with `getByRole` as well.

<details open>
<summary>📸  Screenshot(s)</summary>

Captura de pantalla 2026-06-16 a la(s) 1 12 02 p m

</details>

refiner/schema.sql
rogeruiz: No it doesn't on regenerate. This was probably a Git merge that just added both of them when merging things into `main`.

client/e2e/configuration.spec.ts
jakewheeler: Did you intend to change other `locator()` calls outside of the specific instances I mentioned in my comments? 

This image makes it look like this line (for example):
```ts
const savedCodeTableRows = page.locator('table tbody tr');
```

changed to this:
```ts
const savedCodeTableRows = page.getByRole('table tbody tr');
```

Which does cause an error since it's not valid.

client/e2e/configuration.spec.ts
jakewheeler: Oh, I understand now. I didn't initially understand what the comment was getting at.

jakewheeler (approved):
🫡

PR #1324, opened 2026-06-04
fix: e2e a11y checks are too limited

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR:

- Updates the rulesets the Playwright Axe checks use to be closer to the default used by the Axe browser extension
- Excludes the `react-diff-viewer` from a11y checks since it gets flagged and we cannot fix it (noted in a comment)
- Small update to `toHaveNoAxeViolations()`
- Increased a11y checks throughout e2e tests
- Fixed an a11y issue related to the config builder menu

## 🧪 How to test

- Tests and app should work as they did before
- See an [example](https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/26978095217/job/79610146307) of failing a11y issues in the job report

### DIFF
--- a/client/e2e/activity-log.spec.ts
+++ b/client/e2e/activity-log.spec.ts
@@ -10,7 +10,11 @@ test.describe('Activity log', () => {
     await deleteAllConfigurations();
   });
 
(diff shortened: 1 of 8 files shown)


fzhao99 (approved):
thanks!

PR #1325, opened 2026-06-05
test: use `eslint-plugin-jsx-a11y` for improved a11y checking during development

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR aims to improve the dev team's ability to detect potential a11y issues during development. `eslint-plugin-jsx-a11y` has been added to the project and can be run locally with `npm run lint`. It will also run automatically in CI on all PRs/updates.

- Added [`eslint-plugin-jsx-a11y`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y)
  - Opted into recommended ruleset 
- Fixed a variety of issues discovered by the plugin
  - Added exceptions for a few non-issues (primarily using `autoFocus` within modals, which is valid)
- Fixed a deprecation notice about not using `focus-trap-react`'s default export
- Improved implementation and usability of condition code set drawer
- Updated tests as needed

## How to test

Main thing to check is the drawer. The `Add` and `Remove` buttons are now the only interactive elements within the list. This should work in a much more polished and expected manner than the previous implementation. It should also eliminate a11y issues related to this component.

Both mouse controls and keyboard controls should work in a very natural way.

Example:
Kapture 2026-06-05 at 11 39 17

> [!IMPORTANT]
> Tiffany [approves](https://skylight-hq.slack.com/archives/C08QQ1SCDLY/p1780674840295139?thread_ts=1780674305.236949&cid=C08QQ1SCDLY) of the design changes 👍

### DIFF
--- a/client/src/pages/Simulator/RunSimulation.tsx
+++ b/client/src/pages/Simulator/RunSimulation.tsx
@@ -46,7 +46,7 @@ export function RunSimulation({
       <div className="flex flex-col gap-6 xl:flex-row">
         <Container className="flex-1" color="white">
           <Content className="flex items-center gap-6">
-            <img className="px-3 py-1" src={UploadSvg} role="presentation" />
+            <img className="px-3 py-1" src={UploadSvg} alt="" />
             <div className="flex flex-col items-center gap-10">
               <p className="flex flex-col items-center gap-2 text-black">
                 <span className="font-bold">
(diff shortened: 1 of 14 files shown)


client/src/pages/Simulator/RunSimulation.tsx
```diff
@@ -46,7 +46,7 @@ export function RunSimulation({
       <div className="flex flex-col gap-6 xl:flex-row">
         <Container className="flex-1" color="white">
           <Content className="flex items-center gap-6">
-            <img className="px-3 py-1" src={UploadSvg} role="presentation" />
+            <img className="px-3 py-1" src={UploadSvg} alt="" />
```
fzhao99, line 49: did we mean to add an alt description here?

jakewheeler: Yes.  [The empty `alt` value is preferred over `role="presentation"`](https://www.w3.org/WAI/tutorials/images/decorative/). 

I updated this based on this message from the linter:

> 49:13  error  Prefer alt="" over a presentational role. First rule of aria is to not use aria if it can be achieved via native HTML  jsx-a11y/alt-text

fzhao99 (approved):

PR #1328, opened 2026-06-06
docs(rfc): Spike for reconstructing the narrative

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

Adds RFC proposing how the refiner will rebuild a section's human-readable 
`<text>` from the entries that survive pruning, starting with Results, Problems, 
Medications, and Immunizations.

This is documentation only--no application code changes. The RFC records the
architecture decision and the reasoning behind it; implementation follows in a
later PR once the approach is reviewed.

In short, the chosen approach ("Option C") splits the work into three layers:
shared mechanical primitives written once (a typed-value renderer, a field
extractor, a table builder); per-section **field maps as data** (keyed by
`template_id`); and short per-section **join functions as code** that handle the
structural reach into ancestors/siblings. The RFC explains why this beats both a
fully-hardcoded approach and a fully-data-driven one, and how the field maps
become a down payment on the future template-aware matching engine without
building it now.

A runnable companion script (`reconstruction_toy.py`) demonstrates the design
end-to-end on a small eICR fragment.

Please review the RFC itself for the full context, options comparison, and
deferred questions:
- `docs/rfc/0010-narrative-reconstruction.md`

## 🔗 Related Issue

Fixes #1241
- #1241

## 🧪 How to test

You can run the toy example folded into the RFC if you'd like to test things out.


jakewheeler (approved):
Great work Robert 🎉

PR #1329, opened 2026-06-06
fix(test): Fix the `check-seeding` script

robertmitchellv:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #
- #


robertmitchellv:
an alternate solution is proposed on #1794

PR #1342, opened 2026-06-08
separate out declarative and config code

fzhao99:


client/src/components/Tooltip/index.tsx
```diff
@@ -21,57 +21,59 @@ const arrowClasses: Record<TooltipPosition, string> = {
...
+const POSITION_MAPPING: Record<TooltipPosition, PositionConfig> = {
+  top: {
+    left: (r) => r.left + r.width / 2,
+    top: (r) => r.top - BOUNDING_GAP,
+    transform: 'translate(-50%, -100%)',
+  },
+  bottom: {
+    left: (r) => r.left + r.width / 2,
+    top: (r) => r.bottom + BOUNDING_GAP,
+    transform: 'translateX(-50%)',
```
fzhao99, line 41: think this translation needed to be different from the translation for top and left

rogeruiz: I'm a bit confused. Could you expand on this comment a bit so I can understand it better?

fzhao99: oh think this was a copy-paste issue on my end. feel free to disregard

PR #1349, opened 2026-06-09
chore: prevent testing report from going stale

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

> [!IMPORTANT]
> `run-tests/run-python-tests` is set as an expected check but no longer exists. I will fix this at the repo level prior to this PR being merged.

This PR primarily does two things:

- Updates the CI job to allow unit/integration/scenario tests in parallel
- Updates `scenarios` to produce a new `REPORT.md` file automatically when `--update-snapshots` is used

This differs a bit from the idea we originally discussed where CI would be responsible for generating a new `REPORT.md` as needed. The reason for this is because a new document only needs to be produced when the snapshots change. If the snapshots change, that means a developer made a change to the scenario setup and is responsible for making any updates.

However, the changes in this PR means the developer only needs to run `--update-snapshots` and a report will also be generated. This should help reduce the burden needed to create the report and will also prevent the report from going stale.

## 🔗 Related Issue

Fixes #1311
- #1311

## 🧪 How to test

- Automated tests run and pass in CI
- Run `pytest tests/integration/scenarios --update-snapshots -s` locally. You'll see `Wrote integration/scenarios/REPORT.md` printed once the tests successfully complete. You can delete `REPORT.md` before running this command if you'd like to ensure `REPORT.md` is being written


.github/workflows/tests.yml
```diff
@@ -30,9 +30,45 @@ jobs:
         run: |
           pip install -U pip
           pip install -r dev-requirements.txt -r requirements.txt
-      - name: Run tests
+      - name: Run integration tests
```
rogeruiz: Should we have this be for unit tests instead?


```suggestion
      - name: Run unit tests
```

.github/workflows/tests.yml
```diff
@@ -30,9 +30,45 @@ jobs:
...
+        uses: actions/checkout@v6
+      - name: Setup python ${{env.PYTHON_RUNNER_VERSION}}
+        uses: actions/setup-python@v6
+        with:
+          python-version: ${{env.PYTHON_RUNNER_VERSION}}
+          cache: pip
+      - name: Install dependencies
+        run: |
+          pip install -U pip
+          pip install -r dev-requirements.txt -r requirements.txt
```
rogeruiz, line 50: I like that things are split up into multiple jobs, but it also means that we'll have multiple calls to `pip install *` as well. I'm not sure if the parallelism will make that a moot point if it's still faster. 

> [!NOTE]
> Just an observation not a suggestion.

refiner/tests/integration/scenarios/build_report.py
```diff
@@ -11,7 +11,7 @@
 # PATHS
 # =============================================================================
 
-SCRIPT_DIR: Path = Path(__file__).resolve().parent
+SCRIPT_DIR: Path = Path(__file__).resolve()
```
rogeruiz: Without the `.parent` this is now the file itself, should we rename this to `SCRIPT_FILE` to make that more clear?


```suggestion
SCRIPT_FILE: Path = Path(__file__).resolve()
```

refiner/tests/integration/scenarios/build_report.py
```diff
@@ -11,7 +11,7 @@
...
 SCENARIOS_DIR: Path = SCRIPT_DIR.parent
```
rogeruiz:
```suggestion
SCENARIOS_DIR: Path = SCRIPT_FILE.parent
```

If the other suggestion is approved.

refiner/tests/integration/scenarios/build_report.py
```diff
@@ -480,7 +480,7 @@ def _render_appendix() -> str:
     )
 
 
-def build_report(scenarios: list[ScenarioSnapshot]) -> str:
+def _compose_report(scenarios: list[ScenarioSnapshot]) -> str:
```
rogeruiz, line 483: Thanks for this! Much easier to understand now.

refiner/tests/integration/scenarios/build_report.py
rogeruiz: Do we need to update these comments in this file?


```suggestion
            python tests/integration/scenarios/build_report.py
```

rogeruiz: Do we need to update these comments in this file?


```suggestion
            python tests/integration/scenarios/build_report.py         # regenerate this report
```

refiner/tests/integration/scenarios/conftest.py
```diff
@@ -47,6 +48,20 @@ def update_snapshots(request: pytest.FixtureRequest) -> bool:
...
 
 
+# This hook runs automatically at the end of the test session
+def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
+    """
+    Generate a new report at the end of the scenario run.
+
+    All scenarios must pass and the update snapshot command must be used.
+    """
+    if exitstatus != 0:
```
rogeruiz, line 58: Should we have this check here? If snapshots get updated in a PR but an unrelated test fails, the `REPORT.md` file won't be updated. I guess we'd be able to avoid the drift by just not approving those kinds of changes. So maybe it's not a big deal, but I figured I'd bring up my thoughts here.

rogeruiz (changes requested):
❌ This PR needs changes ...

Good work on this @jakewheeler. I have some small suggestions here.

.github/workflows/tests.yml
jakewheeler: Thank you 🤦

.github/workflows/tests.yml
jakewheeler: Yeah, it's a good thought. The [jobs](https://github.com/CDCgov/dibbs-ecr-refiner/actions/runs/27363137522/job/80855553066) do seem to be using the cached dependencies when available.

I'm eyeballing it against `main` but it looks like it's around ~20-30 seconds faster in total. The integration tests job is the bottleneck.

refiner/tests/integration/scenarios/conftest.py
jakewheeler: I think my idea was that there's no point in running a rebuild since failures would mean the tests are still in a working state and the PR wouldn't be mergeable at this point anyway. I think you're right it doesn't matter a whole lot either way. It just seemed unnecessary to me at the time to attempt a rebuild when the scenarios aren't successful.

rogeruiz (approved):
lgtm 🌈 

Looks good based on the changes that were accepted. 🚀

PR #1350, opened 2026-06-09
refactor: remove code array from store and read of activated configuration

lina-roth:
# 🔀 PULL REQUEST

### local s3 see codes no longer listed 
Screenshot 2026-06-10 at 3 11 59 PM

## 💡 Summary

- activated configs will no longer store codes
- instead will concat codes array from code set system data upon read 

## 🔗 Related Issue

Fixes #1251 
- #1251 

## ✅ Acceptance Criteria

- [ ]  codes array is no longer written to the active.json file upon activation
 - [ ] Check if there is other data that we can avoid writing to active.json
 - [ ] Remove it as a part of this work if so
 - [ ] Refining works exactly as it did previously

## 🧪 How to test

1. Run `docker compose up`
4. Create COVID-19 config, activate, go here http://localhost:4566/local-config-bucket/configurations/SDDH/07221093-b8a1-4b1d-8678-259277bfba64/1/active.json 
5. see codes no longer in active.json in s3 (see screenshot above)


refiner/app/lambda/test_lambda.py
```diff
@@ -231,9 +231,17 @@ def test_lambda_one_active(
 
     # Create activation for COVID
     covid_activation = {
-        "codes": ["101289-7"],
         "sections": [],
         "included_condition_rsg_codes": ["840539006"],
+        "code_system_sets": {
```
fzhao99, line 237: we have a fixture [here](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/tests/fixtures/lambda/active.json) that we'll need to clean up similarly as well

fzhao99 (approved):

PR #1351, opened 2026-06-09
test(tes): add update to check for acg <-> cg completion flag mismtaches

robertmitchellv:
# 🔀 PULL REQUEST

## 🔗 Related Issue

Fixes #
- #


robertmitchellv:
an alternate solution is proposed on #1794

PR #1355, opened 2026-06-11
feat: cg to snomed mapping

### LINKED ISSUE #1099: [USER STORY] CG to SNOMED Mapping - Configuration Detail Page
## 👤💭 User Story

As a user, I can see which SNOMED condition codes map to my configuration, so that I can understand which eICRs my configurations will affect.

## 🏗️ Acceptance Criteria

- [ ] Icon next to condition name
- [ ] Modal to show which SNOMED codes map to the condition grouper

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=8983-49624&t=ulpJNGvyN2PCi8E4-1

## 📜 Additional context

Our "reportable conditions" within the app are actually aggregates of one or more SNOMED condition codes that would actually show up in the RR. We aren't currently mapping that out anywhere so there's no way for the users to know what reportable condition SNOMED codes in the RR map to each configuration.

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Frontend for the CG to SNOMED Mapping work

https://github.com/user-attachments/assets/9d06735a-49b2-428f-8063-e90ff1e2ef45

## 🔗 Related Issue
Fixes #1099
- #1099

## ✅ Acceptance Criteria

- [ ] Icon next to condition name
- [ ] Modal to show which SNOMED codes map to the condition grouper

## 🧪 How to test
- Make a configuration
- Click the icon next to the name. Verify CG information renders

## ℹ️ Additional Information
Any organizational work that we should do in this PR as well?


refiner/app/api/v1/configurations/base.py
```diff
@@ -297,6 +301,7 @@ async def get_configuration(
             [format_section_naming(section) for section in config.section_processing],
             key=lambda r: r.name.lower(),
         ),
+        rsg_codes=rsg_codes,
```
fzhao99, line 306: know that the ConfigurationResponse is getting to be a bit of a catch-all object, so open to suggestions on to where / whether to split this up

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -549,3 +570,69 @@ function OptionsListContainer({ children }: { children: React.ReactNode }) {
...
+    >
+      <ModalHeader className="pb-1!">
+        <ModalTitle className="font-public-sans">
+          {primaryConditionDisplayName}
+        </ModalTitle>
+      </ModalHeader>
+      <ModalBody>
+        <p className="wrap-break-word">
+          Applies to eCR documents reportable for reportable condition SNOMED
+          codes {rsgCodesString}. Only one output will be produced per condition
```
fzhao99: @kplaneaux slightly tweaked the copy [here](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=8983-48846&t=d7yGflG43JFF2yFT-4) to include the extra phrase "condition SNOMED codes" before the code string because I though things read awkwardly without it.

refiner/app/db/codes/db.py
```diff
@@ -0,0 +1,27 @@
...
+    configuration_primary_condition_id: UUID, db: AsyncDatabaseConnection
+) -> list[DbCode]:
+    """
+    Function to get all RSG code objects for a configuration.
+    """
+
+    query = """
+        SELECT c.name, c.value, c.version, c.system_id
+        FROM conditions_rsg_codes as rsg
+        LEFT JOIN codes c on c.id = rsg.code_id
```
fzhao99, line 19: lmk if we want to organize this function elsewhere. Figured I'd seed the start of the codes directory that we'll build on as we go, but happy to put it in another place / organize it some other way if it makes sense

client/e2e/configuration.spec.ts
```diff
@@ -18,13 +18,26 @@ test.describe('Configuration detail flow', () => {
...
+
+    await expect(page.getByRole('columnheader')).toHaveText([
+      'SNOMED code',
+      'Display name',
+    ]);
+    const rsgDetailsTable = page.getByRole('table').getByRole('row');
+    await expect(rsgDetailsTable.first()).toBeVisible();
+    const rowCount = await rsgDetailsTable.count();
+    // header plus at least one RSG row
+    expect(rowCount).toBeGreaterThanOrEqual(2);
```
jakewheeler: 👍 good test.

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -70,7 +74,23 @@ export function ConfigBuild() {
...
+        <div className="flex items-center">
+          <Title>{configuration.data.display_name}</Title>
+          <button
+            onClick={() => setIsRsgDetailsModalOpen(true)}
+            className="ml-2! text-white hover:cursor-pointer"
+            aria-label="Open reporting specification details modal"
+            name="rsg-details-open-button"
+          >
+            <InfoIcon />
+          </button>
```
jakewheeler: We can use the `Button` component here with `variant="unstyled"`. I'd also suggest using a flex container to group the title and button.
```ts
<div className='flex items-center gap-2'>
  <Title>{configuration.data.display_name}</Title>
  <Button
    variant="unstyled"
    onClick={() => setIsRsgDetailsModalOpen(true)}
    aria-label="Open reporting specification details modal"
  >
    <InfoIcon />
  </Button>
</div>
```

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -70,7 +74,23 @@ export function ConfigBuild() {
...
+          <RsgDetailsModal
+            open={isRsgDetailsModalOpen}
+            onClose={() => setIsRsgDetailsModalOpen(false)}
+            primaryConditionDisplayName={configuration.data.display_name}
+            rsgCodes={configuration.data.rsg_codes}
+          ></RsgDetailsModal>
```
jakewheeler:
```suggestion
          />
```

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -549,3 +569,59 @@ function OptionsListContainer({ children }: { children: React.ReactNode }) {
...
+}: RsgDetailsModal) {
+  return (
+    <Modal
+      className="max-w-160!"
+      open={open}
+      onClose={onClose}
+      position="center"
+    >
+      <ModalHeader className="pb-1!">
+        <ModalTitle className="font-public-sans">
```
jakewheeler: I'd suggest we either:
1. Not make these styling changes (yet)
2. Make these styles the defaults in the components themselves rather than using the prop here

I think it'd be ideal if we very rarely modify the default styling so that modals will look consistent throughout the app. Maybe we should ask design if we want to do this now or hold off?

client/e2e/configuration.spec.ts
jakewheeler: Actually, wouldn't hurt to throw an a11y check in here too while the modal is open.

client/src/components/Modal/index.tsx
```diff
@@ -106,8 +106,14 @@ interface ModalSectionProps {
   className?: string;
 }
 
-function ModalHeader({ children }: ModalSectionProps) {
-  return <div className="mx-auto mt-6 w-full pr-10 pb-6 pl-6">{children}</div>;
+function ModalHeader({ children, className }: ModalSectionProps) {
+  return (
+    <div
+      className={classNames('mx-auto mt-6 w-full pr-10 pb-6 pl-6', className)}
```
jakewheeler: Related to my other comment - unsure currently if we should allow flexibility here or make your changes the new defaults.

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -549,3 +569,59 @@ function OptionsListContainer({ children }: { children: React.ReactNode }) {
...
+          {primaryConditionDisplayName}
+        </ModalTitle>
+      </ModalHeader>
+      <ModalBody>
+        <p className="wrap-break-word">
+          Applies to eCR documents reportable for the codes below. Only one
+          output will be produced per condition group.
+        </p>
+
+        <table>
```
jakewheeler, line 601: You don't need to do this here but this is reminding me we still need to make a table component 😭

refiner/app/api/v1/configurations/base.py
jakewheeler: No need IMO. This request drives the page which now requires the data you've added, so this is correct.

refiner/app/api/v1/configurations/base.py
```diff
@@ -282,6 +283,9 @@ async def get_configuration(
         code_systems=code_systems,
     )
 
+    rsg_codes = await get_configuration_rsg_codes(
+        db=db, configuration_primary_condition_id=primary_condition.id
+    )
```
jakewheeler: Nitpick: I'd flip these since `db` is typically last.

```suggestion
    rsg_codes = await get_configuration_rsg_codes(
        configuration_primary_condition_id=primary_condition.id,
        db=db,
    )
```

client/src/pages/Configurations/ConfigBuild/index.tsx
fzhao99: I can get with Tiffany about it

jakewheeler: Thanks!

refiner/app/db/codes/db.py
```diff
@@ -0,0 +1,27 @@
...
+
+from psycopg.rows import class_row
+
+from app.db.codes.model import DbCode
+from app.db.pool import AsyncDatabaseConnection
+
+
+async def get_configuration_rsg_codes(
+    configuration_primary_condition_id: UUID, db: AsyncDatabaseConnection
+) -> list[DbCode]:
```
jakewheeler: I think this naming is a little confusing since it sounds like the configuration is required when it isn't. I might suggest making it more condition focused:

```suggestion
async def get_rsg_codes_by_condition_id_db(
    condition_id: UUID, db: AsyncDatabaseConnection
) -> list[DbCode]:
```

refiner/app/db/codes/db.py
jakewheeler: I think this is probably the right way to go. Codes are now their own entity in the DB so this makes sense to me.

The other place I was thinking about is `app/db/conditions/db.py` since codes are so heavily associated with conditions.

client/src/components/Modal/index.tsx
fzhao99: Chatted with Tiffany in the design channel and the new settings here should be the defaults.

- No padding between  `<ModalTitle/>` and the next element in `<ModalHeader/>`
- 24 px (`.pb-6`) between `<ModalHeader/>` and   `<ModalBody/>` and between `<ModalBody/>` and `<ModalFooter/>`

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -549,3 +570,58 @@ function OptionsListContainer({ children }: { children: React.ReactNode }) {
...
+}: RsgDetailsModal) {
+  return (
+    <Modal
+      className="max-w-160!"
+      open={open}
+      onClose={onClose}
+      position="center"
+    >
+      <ModalHeader className="pb-1!">
+        <ModalTitle className="font-public-sans">
```
jakewheeler: We can delete the classes from both of these now, right?

client/src/components/Modal/index.tsx
jakewheeler: Thanks for getting this worked out! Looks great.

client/src/pages/Configurations/ConfigBuild/index.tsx
```diff
@@ -70,7 +73,24 @@ export function ConfigBuild() {
   return (
     <>
       <TitleContainer>
-        <Title>{configuration.data.display_name}</Title>
+        <div className="flex items-center gap-2">
+          <Title>{configuration.data.display_name}</Title>
+          <Button
+            variant="tertiary"
+            onClick={() => setIsRsgDetailsModalOpen(true)}
+            className="pl-0!"
```
jakewheeler: Let's set this to `p-0!` so it eliminates the slight shift when clicking from build to another workflow page.

```suggestion
            className="p-0!"
```

jakewheeler (approved):
Good work!

PR #1356, opened 2026-06-11
fix: section table text contrast issue

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

Originally fixed in #1323 and then (assuming accidentally) reverted by #1318. I've updated the e2e tests to ensure color contrast issues will cause test failures.

## 🧪 How to test

- Sections table works as it did before
- No color contrast issues detected relating to `Not applicable for this section` in sections table


client/src/pages/Configurations/ConfigBuild/Sections/index.tsx
```diff
@@ -159,13 +159,23 @@ export function Sections({
...
+                        ) : (
+                          <RefineSwitch
+                            configurationId={configurationId}
+                            currentSection={section}
+                            sections={sectionProcessing}
+                            disabled={
+                              disabled || isDisabledSection(section.code)
+                            }
+                          />
+                        )}
```
rogeruiz, line 178: Looks like some stuff got lost in the merging across all the PRs. The `isNarrativeSection` is handled inside of the `RefineSwitch` on `main` and in the previously approved PRs. This looks like it's going to revert that work. I can work on getting just the a11y changes from this PR updated and not have it change the structure of the component. Or would you like to do it yourself @jakewheeler?

jakewheeler: This is an intentional refactor I made while updating the text. I noticed that the `RefineSwitch` was only taking the narrative section value to determine if it should show the text instead. I moved this to the higher level so the prop no longer needs to be passed in.

rogeruiz (approved):
lgtm 🌈

PR #1360, opened 2026-06-11
fix: change copy for diff rendering limit value

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
We were using the wrong constant for the error we render for the diff view as discovered during an office hour. Getting this out to fix it
Screenshot 2026-06-11 at 3 29 47 PM

## 🧪 How to test
Can use this file through the upload flow to get to this screen

[sddh-covid-3.5-mb.zip](https://github.com/user-attachments/files/28853229/sddh-covid-3.5-mb.zip)

## ℹ️ Additional Information

Any suggestions on a better test / way of catching something like this are welcome

### DIFF
--- a/client/src/components/Diff/index.test.tsx
+++ b/client/src/components/Diff/index.test.tsx
@@ -22,7 +22,7 @@ describe('Diff view', () => {
     );
     expect(
       screen.getByText(
-        `Maximum uncompressed file size is ${FileInfoResponseValue.max_for_diff_rendering_mb}`
+        `Maximum uncompressed file size is ${FileInfoResponseValue.max_for_diff_rendering_mb}MB`
       )
     ).toBeInTheDocument();
   });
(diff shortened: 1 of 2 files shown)


client/src/components/Diff/index.test.tsx
```diff
@@ -21,9 +21,12 @@ describe('ExternalLink', () => {
...
     expect(
-      screen.getByText('Maximum uncompressed file size is', {
-        exact: false,
-      })
+      screen.getByText(
+        `Maximum uncompressed file size is ${FileInfoResponseValue.max_for_diff_rendering_mb}`,
+        {
+          exact: false,
+        }
+      )
```
jakewheeler, line 26:
```suggestion
      screen.getByText(
        `Maximum uncompressed file size is ${FileInfoResponseValue.max_for_diff_rendering_mb}MB`
      )
```

client/src/components/Diff/index.test.tsx
```diff
@@ -21,9 +21,12 @@ describe('ExternalLink', () => {
       ></Diff>
     );
```
jakewheeler, line 22: Would you mind also updating the `describe` text? It's currently `ExternalLink` 😆

jakewheeler (approved):
🫡

PR #1363, opened 2026-06-12
refactor(refiner): Update RefinerComplete schema documentation & split RefinerCompleteFile class

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This establishes the contract for the upcoming error signal. This patch only
updates the documentation and prepares the Class structure.

## 🔗 Related Issue

- Closes #1358

## ✅ Acceptance Criteria

- [x] Update `RefinerComplete` schema documentation to include an optional
      `Error` field.
- [x] Update `RefinerCompleteFile` type definition to support an optional
      `Error` string and partial payloads (`total=False`).
- [ ] ~~Modify `lambda_handler` to ensure a `RefinerComplete` file is written to
      S3 when a fatal error occurs, provided a `persistence_id` could be extracted.~~
- [ ] ~~The error signal must set `RefinerSkip: true` and include the exception
      message in the `Error` field.~~
- [ ] ~~The Lambda must still re-raise the exception after writing the signal to
      maintain failure visibility in AWS.~~
- [ ] ~~Integration tests must verify that no signal is written if
      `persistence_id` extraction fails (e.g. an invalid S3 prefix).~~
- [ ] ~~Integration tests must verify that a signal with `RefinerSkip: true` and
      currently a missing `setID` in the document).~~

## 🧪 How to test

1. Read the updated documentation `refiner/app/lambda/RefinerComplete.md`.
1. Ensure that the `RefinerCompleteFile` type definition supports an optional
   `Error` string and partial payloads (`total=False`).

## ℹ️ Additional Information

> [!NOTE]
> This work is part of three separate PRs related to #1358. This first PR should
> be reviewed first. Subsequent PRs build off this work and will be merged into
> `main` only after the previous PR is merged in.

### DIFF
--- a/refiner/app/lambda/lambda_function.py
+++ b/refiner/app/lambda/lambda_function.py
@@ -51,7 +51,7 @@ ConditionCode = str
 RefinerMetadata = dict[JurisdictionCode, dict[ConditionCode, bool]]
 
 
-class RefinerCompleteFile(TypedDict):
+class RefinerCompleteFile(TypedDict, total=False):
     """
     Represents the completion file written after all refinement is done.
 
@@ -61,6 +61,7 @@ class RefinerCompleteFile(TypedDict):
     RefinerMetadata: RefinerMetadata
     RefinerSkip: bool
     RefinerOutputFiles: list[str]
+    Error: str
 
 
 @dataclass
--- a/refiner/app/lambda/lambda_function.py
+++ b/refiner/app/lambda/lambda_function.py
@@ -51,9 +51,9 @@ ConditionCode = str
 RefinerMetadata = dict[JurisdictionCode, dict[ConditionCode, bool]]
 
 
-class RefinerCompleteFile(TypedDict, total=False):
+class RefinerCompleteSuccess(TypedDict):
     """
-    Represents the completion file written after all refinement is done.
+    Represents a successful completion file written after all refinement is done.
 
     Required by AIMS infrastructure.
     """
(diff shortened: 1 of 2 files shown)


refiner/app/lambda/lambda_function.py
```diff
@@ -51,7 +51,7 @@
 RefinerMetadata = dict[JurisdictionCode, dict[ConditionCode, bool]]
 
 
-class RefinerCompleteFile(TypedDict):
+class RefinerCompleteFile(TypedDict, total=False):
```
jakewheeler: Since we have two different "complete" states now, should we split this into two different classes that require all variables to have values? This feels safer to me than marking everything as optional. What do you think?

```python
class RefinerCompleteFileSuccess(TypedDict):
    """
    Represents the RefinerComplete file content written after refinement is successful.

    Required by AIMS infrastructure.
    """

    RefinerMetadata: RefinerMetadata
    RefinerSkip: bool
    RefinerOutputFiles: list[str]

class RefinerCompleteFileFailure(TypedDict):
    """
    Represents the RefinerComplete file content written when refinement fails.

    Required by AIMS infrastructure.
    """

    Error: str
```

refiner/app/lambda/lambda_function.py
```diff
@@ -63,6 +63,20 @@ class RefinerCompleteFile(TypedDict):
...
+    Represents a fatal error completion file written despite of failure.
+
+    Required by AIMS infrastructure.
+    """
+
+    RefinerSkip: bool
+    Error: str
+
+
+RefinerCompleteFile = RefinerCompleteSuccess | RefinerCompleteError
```
jakewheeler, line 77: I didn't think about this with how the work is split 🤔 

I assume this is going to be deleted in a later PR when it no longer has a use?

rogeruiz: In #1365, I've added a comment that explains why I kept it. I think it's okay to keep even though it's not being used as it's partly documentation and partly there to be flexible for local development if there's another "complete" type in the future.

👀 https://github.com/CDCgov/dibbs-ecr-refiner/pull/1365/changes#r3405158713

jakewheeler (approved):
👍

PR #1364, opened 2026-06-12
fix: move RSG join status log 

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Realized in a recent debugging session that a status log for the RSG <> conditions seeding step needed to be moved to account for the new way we're grabbing the code ID's

## 🧪 How to test
Make sure the seeding script works the same way and the number of joins is accurately represented

## ℹ️ Additional Information
Any other script status logs we should change / add?


jakewheeler (approved):
Seems to be working for me 👍

PR #1365, opened 2026-06-12
refactor(refiner): Update error handling

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This updates the Lambda to ensure that exceptions still produce an error signal
to S3 in the form of a JSON object that APHL provided the shape for. This patch
also ensures that exceptions are re-raised for Lambda visibility in the AWS logs
to maintain that functionality while providing this new functionality.

## 🔗 Related Issue

- Closes #1358
- Builds on #1363

## ✅ Acceptance Criteria

- [x] ~~Update `RefinerComplete` schema documentation to include an optional
      `Error` field.~~
- [x] ~~Update `RefinerCompleteFile` type definition to support an optional
      `Error` string and partial payloads (`total=False`).~~
- [x] Modify `lambda_handler` to ensure a `RefinerComplete` file is written to
      S3 when a fatal error occurs, provided a `persistence_id` could be extracted.
- [x] The error signal must set `RefinerSkip: true` and include the exception
      message in the `Error` field.
- [x] The Lambda must still re-raise the exception after writing the signal to
      maintain failure visibility in AWS.
- [ ] ~~Integration tests must verify that no signal is written if
      `persistence_id` extraction fails (e.g. an invalid S3 prefix).~~
- [ ] ~~Integration tests must verify that a signal with `RefinerSkip: true` and
      currently a missing `setID` in the document).~~

## 🧪 How to test

> [!IMPORTANT]
> Make sure that Localstack is running. (`just dev up`).

1. Happy path (avoiding regressions)
   - Run tests (`pytest refiner/tests/integration/test_lambda.py`).
   - Check for the `RefinerComplete` file created in Localstack S3 with
     `RefinerSkip: false`.
   - Logs should show "Successfully processed X refiner outputs".
1. Failure during processing (new functionality)
   - You'll need to temporarily introduce a failure or provide a corrupted
     file. I'll show how to introduce the failure instead of using a bad file.
   - Modify `refiner/app/lambda/lambda_function.py` to raise an exception
     immediately after `persistenc_id` is extracted:
     ```py
     persistence_id = extract_persistence_id(...)
     raise Exception("Simulated Fatal Error") # <- Add this
     ```
   - Trigger the Lambda via the integration test harness or a curl request to
     the Localstack endpoint:

     ```sh
     curl -X POST \
          http://localhost:9000/2015-03-31/functions/functions/invocations \
          -d '{"Records": [...]}'
     ```

   - Ensure that the request returns a 500 error or a Lambda error response. You
     should see a stack trace.S3 for the completion file.

1. Early failure (edge case)
   - Trigger the Lambda with a malformed event (e.g. an S3 event missing the
     expected bucket/key structure)
   - Or, manually modify the code to raise an exception before `persistence_id =
None` is initialized.
   - The process should fail and log the error as usual since the
     `persistence_id` was never determined.

## ℹ️ Additional Information

> [!NOTE]
> This work is part of three separate PRs related to #1358. This second PR
> should be reviewed after #1363. Subsequent PRs build off this work and will be
> merged into `main` only after the previous PR is merged in.


refiner/app/lambda/lambda_function.py
```diff
@@ -74,6 +74,9 @@ class RefinerCompleteError(TypedDict):
     Error: str
 
 
+# NOTE: Alias exists for type flexibility and documentation for what a
+# completion file can be in this system. Use the specific types for assignments
+# to keep our types safe.
```
rogeruiz: @jakewheeler here's the comment I mentioned in #1363.

rogeruiz: If we decide to remove this and the line below it. Let's do it in this PR so that #1363 isn't held up and this PR doesn't need to rebase on those changes and force-pushed again.

jakewheeler: I can't help being the YAGNI guy so I'm in favor of deleting it 😆

refiner/app/lambda/lambda_function.py
```diff
@@ -173,66 +172,95 @@ def lambda_handler(event, context) -> dict:
...
+                    f"Successfully processed {len(result.output_file_keys)} refined outputs"
+                )
+
+            except Exception as e:
+                logger.error(f"Fatal error processing record: {str(e)}", exception=e)
+
+                # Attempt to write a skip file if we at least have a persistence_id
+                try:
+                    if persistence_id:
+                        complete_key = f"{REFINER_COMPLETE_PREFIX}{persistence_id}"
```
jakewheeler: We should be alright to delete this since we've built it higher up already.

refiner/app/lambda/lambda_function.py
```diff
@@ -173,66 +172,95 @@ def lambda_handler(event, context) -> dict:
...
+                            ContentType="application/json",
+                        )
+                        logger.info(f"Wrote fatal error signal to {complete_key}")
+                except Exception as s3_err:
+                    logger.error(
+                        f"Failed to write error signal to S3: {str(s3_err)}",
+                        exception=s3_err,
+                    )
+
+                raise e
```
jakewheeler: I think this means a single bad record will cause all record processing to halt, even if later ones would have been processed fine. I believe we still run 1 record per Lambda function but it'd probably make sense to not assume that's always going to be the case.

refiner/app/lambda/lambda_function.py
rogeruiz: I think we still need this. The `complete_key` gets built higher up, but the `try..except` block on L239 catches any errors that may occur _before_ that point. Specifically if an exception occurs during `extract_persistence_id` L177, `get_s3_object_content` L184 & L194, or `run_refinement` L204. The flow would then enter the `except` block on L239. This check is trying to ensure that a `RefinerCompleteError` is only written if the we can identify which document is being processed. Since the `persistence_id` would still be `None` resulting in an invalid S3 path or fail with a TypeError during string concatenation. If an exception occurs at or before L177 (e.g., during json.loads at L169, or if `extract_persistence_id` itself raises a ValueError as it is designed to do at L294), `persistence_id` will remain `None`.

refiner/app/lambda/lambda_function.py
rogeruiz: Good catch here. Yeah my assumptions were off here. I think I have a refactor in place that I'll push up after doing some testing locally. I want to avoid losing "bad" records from the queue so I want to track which `messageId` failed and return a `batchItemFailures` list in the Lambda response. This has the added benefit of allowing us to remove the outer `try..except` L148 since it'd be redundant if we're handling record-level failures instead. But I could keep it if we still want to handle failures like an SQS event format error.

refiner/app/lambda/lambda_function.py
jakewheeler: Should we handle `extract_persistence_id()` potentially throwing before we go on to do anything else? If we can't get that we can stop the record processing right there. We could then assume the rest of the way that `persistence_id` has valid data.

rogeruiz: hmm. sorry i'm not following. we should pair on this if you've got the time today. i'll reach out in Slack too.

rogeruiz: resolved over Slack 🔒 https://skylight-hq.slack.com/archives/C08LC62CVNE/p1781292804570379?thread_ts=1781292001.263389&cid=C08LC62CVNE

refiner/app/lambda/lambda_function.py
```diff
@@ -173,68 +173,100 @@ def lambda_handler(event, context) -> dict:
...
+                )
 
-            logger.info(
-                f"Successfully processed {len(result.output_file_keys)} refined outputs"
-            )
+                logger.info(
+                    f"Successfully processed {len(result.output_file_keys)} refined outputs"
+                )
 
-        return {"statusCode": 200, "message": "Refiner processed successfully"}
```
jakewheeler, line 237: Do we know for sure `statusCode` isn't being used for anything in AIMS? And also that `batchItemFailures` will be used as we expect?

rogeruiz: yeah I added some docs for `batchItemFailures` in the commit that introduced it. I don't believe they're using it for anything since they're just expecting a JSON file at the end of either a success or failure. Can you confirm this @geo-ruvos and @ntjoumas? Thanks

ntjoumas: @rogeruiz - We do not use statusCode in AIMS -- just the content from the JSON file.

jakewheeler (approved):
🥇

PR #1366, opened 2026-06-12
test(lambda): Add integration tests for Errors

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch adds two integration tests for Error Scenarios. These scenarios are:

- Ensure that Lambda doesn't write to S3 when the `persistence_id` cannot be
  extracted. This prevents us writing error files when the S3 URL is malformed.
- Ensure that a signal error is written when there's a fatal processing failure.

## 🔗 Related Issue

- Closes #1358
- Builds on #1363, #1365

## ✅ Acceptance Criteria

- [x] ~~Update `RefinerComplete` schema documentation to include an optional
      `Error` field.~~
- [x] ~~Update `RefinerCompleteFile` type definition to support an optional
      `Error` string and partial payloads (`total=False`).~~
- [x] ~~Modify `lambda_handler` to ensure a `RefinerComplete` file is written to
      S3 when a fatal error occurs, provided a `persistence_id` could be extracted.~~
- [x] ~~The error signal must set `RefinerSkip: true` and include the exception
      message in the `Error` field.~~
- [x] ~~The Lambda must still re-raise the exception after writing the signal to
      maintain failure visibility in AWS.~~
- [x] Integration tests must verify that no signal is written if
      `persistence_id` extraction fails (e.g. an invalid S3 prefix).
- [x] Integration tests must verify that a signal with `RefinerSkip: true` and
      currently a missing `setID` in the document).

## 🧪 How to test

> [!IMPORTANT]
> Make sure that Localstack is running. (`just dev up`).

1. You can verify things are working properly by running the integration test
   for the Lambda function: (`pytest refiner/tests/integration/test_lambda.py`).

## ℹ️ Additional Information

> [!NOTE]
> This work is part of three separate PRs related to #1358. This third PR should
> be reviewed after #1365. Subsequent PRs build off this work and will be merged
> into `main` only after the previous PR is merged in.


refiner/tests/integration/test_lambda.py
```diff
@@ -212,6 +212,83 @@ async def test_lambda_successful_refinement(
...
+
+        resp = await http_client.post(LAMBDA_BASE_URL, json=event)
+
+        # Lambda failure in Localstack might still return 200 but with an errorMessage in body
+        resp_json = resp.json()
+        assert (
+            "batchItemFailures" in resp_json
+            and len(resp_json["batchItemFailures"]) > 0
+            or resp.status_code != 200
+        )
```
jakewheeler: Maybe we can make this slightly more specific to also include what we'd expect inside the array. 

```suggestion
        assert "batchItemFailures" in resp_json
        assert len(resp_json["batchItemFailures"]) == 1
        assert "itemIdentifier" in resp_json["batchItemFailures"][0]
```

We could include `assert resp.status_code == 200` first if we wanted but I think it's no big deal if we leave it out. The test won't be able to proceed if the response isn't what we're expecting anyway.

refiner/tests/integration/test_lambda.py
```diff
@@ -212,6 +212,83 @@ async def test_lambda_successful_refinement(
...
+        complete_key = default_setup["complete_key"]
+        complete_response = s3_client.get_object(Bucket=bucket, Key=complete_key)
+        complete_body = json.loads(complete_response["Body"].read().decode("utf-8"))
+
+        assert complete_body["RefinerSkip"] is True
+        assert "Error" in complete_body
+        assert len(complete_body["Error"]) > 0
+        # RefinerMetadata and RefinerOutputFiles should be omitted when Skip is True
+        assert "RefinerMetadata" not in complete_body
+        assert "RefinerOutputFiles" not in complete_body
```
jakewheeler: This is good. Only suggestion I can think of is that since we expect the error to be an XML parsing error we could also check that. What do you think?
```suggestion
        assert complete_body == {"RefinerSkip": True, "Error": "Failed to parse XML"}
```

refiner/tests/integration/test_lambda.py
```diff
@@ -212,6 +212,83 @@ async def test_lambda_successful_refinement(
...
+            or resp.status_code != 200
+        )
+
+        # Verify no RefinerComplete file was created for the default path
+        try:
+            s3_client.get_object(Bucket=bucket, Key=default_setup["complete_key"])
+            # If it exists, check if it was just created now or existed before.
+            # In a fresh default_setup, it shouldn't be there yet.
+        except Exception:
+            pass
```
jakewheeler: This is a good use case for `pytest.raises` to check if the exception we're expecting to be raised meets what we are anticipating.
```suggestion
        with pytest.raises(ClientError) as exc_info:
            s3_client.get_object(Bucket=bucket, Key=default_setup["complete_key"])
        assert exc_info.value.response["Error"]["Code"] == "NoSuchKey"
```

We'd need to add the `ClientError` import too.
```python
from botocore.exceptions import ClientError
```

refiner/tests/integration/test_lambda.py
```diff
@@ -212,6 +212,81 @@ async def test_lambda_successful_refinement(
...
+        s3_client.put_object(Bucket=bucket, Key=invalid_key, Body=b"dummy content")
+
+        resp = await http_client.post(LAMBDA_BASE_URL, json=event)
+
+        resp_json = resp.json()
+        assert (
+            "batchItemFailures" in resp_json
+            and len(resp_json["batchItemFailures"]) > 0
+            or resp.status_code != 200
+        )
```
jakewheeler: We can probably match this with whatever we end up going with in the other test 👍

refiner/tests/integration/test_lambda.py
```diff
@@ -212,6 +213,76 @@ async def test_lambda_successful_refinement(
...
+        assert "batchItemFailures" in resp_json
+        assert len(resp_json["batchItemFailures"]) > 0
+        assert resp.status_code == 200
```
jakewheeler: Matching this with what we did above is the only thing left. Should be good to go otherwise 🚀 
```suggestion
        assert "batchItemFailures" in resp_json
        assert len(resp_json["batchItemFailures"]) == 1
        assert "itemIdentifier" in resp_json["batchItemFailures"][0]
```

rogeruiz: ah dang missed this thanks.

jakewheeler (approved):
Good work!

PR #1382, opened 2026-06-15
refactor: `configurations.s3_urls` should no longer be an array

### LINKED ISSUE #961: [TASK] `configurations.s3_urls` should no longer be an array
## 🚧 What needs to be done

`s3_urls` in the `configurations` table no longer needs to be an array since we only write a single prefix to S3 per condition now.

Write a migration to:
1. Create a new column called `s3_url` with type `text`
2. Move array data matching `configurations/SDDH/{condition-UUID}` into `s3_url`
3. Delete `s3_urls`

Update the app code to handle this change.

## 🏗️ Acceptance Criteria

- [ ] Steps listed above are completed
- [ ] There is no change to application functionality

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR does two main things:
1. Adds a migrations that renames `configurations.s3_urls` to `configurations.s3_url`, changes the type of the column from `TEXT[]` to `TEXT`, and inserts the first value from the previous array, if applicable
2. Updates the application to handle the new data type and naming

## 🔗 Related Issue

Fixes #961
- #961

## 🧪 How to test

👉  There should be no noticeable change to the application itself.

1. Checkout `main`, create one or more configurations, activate them
2. Checkout this branch
3. Run the `up` migration - you should see the column's name change and rows will have a single URL value
4. Try running the `down` migration - you should see the column's name change back and an array with a single value
5. Run the `up` migration again and check that the app is working as you expect
6. All automated tests should continue to work as before

### DIFF
--- a/refiner/app/api/v1/configurations/activation.py
+++ b/refiner/app/api/v1/configurations/activation.py
@@ -140,8 +140,8 @@ async def activate_configuration(
             detail="Could not find primary condition associated with configuration.",
         )
 
-    # Write the config data to S3 and get the URLs back
-    s3_urls = await run_in_threadpool(
+    # Write the config data to S3 and get the URL back
+    s3_url = await run_in_threadpool(
         upload_configuration_payload, config_payload, config_metadata, logger
     )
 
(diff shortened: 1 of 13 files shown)


lina-roth (approved):
Nice job! Sorry for the delayed review

PR #1386, opened 2026-06-16
chore: improve Lambda integration testing setup locally

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR simplifies local usage of the Lambda integration tests by removing the requirement to run `docker compose build --no-cache lambda` in order to ensure your Lambda image is up-to-date. This should help to prevent issues with Lambda tests appearing to work in the pipeline but not locally (due to forgetting the rebuild step 🙃 ).

## 🧪 How to test

A quick and easy way to see this is working:

1. Add a `print` statement to `lambda_function.py` that you'll be able to easily search for
2. Run `pytest tests/integration/test_lambda.py`. You should see your `print` text in the Lambda container's output without you rebuilding
3. Feel free to try this again with different text to see that it's updating as expected

## ℹ️ Additional Information

This PR does not add Lambda hot-reloading during development, but that could be a nice future improvement if we can figure out a way to get that working.


rogeruiz (approved):
lgtm 🌈

PR #1390, opened 2026-06-17
feat: add cg snomed mapping to config search

### LINKED ISSUE #840: [USER STORY] CG to SNOMED Mapping - Set Up Configuration
## 👤💭 User Story

As a user, I can see which codes map to which condition groupers when I set up a configuration, so that I can understand which eICRs my configurations will affect.

## 🏗️ Acceptance Criteria

- [ ] Condition drop-down updated to a search bar
- [ ] Search only activates after 3 characters entered
- [ ] Search matches on condition grouper name or condition SNOMED code or name

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=8983-47584&t=8DDgpvORfUduYQFV-1

## 📜 Additional context

Our "reportable conditions" within the app are actually aggregates of one or more SNOMED condition codes that would actually show up in the RR. We aren't currently mapping that out anywhere so there's no way for the users to know what reportable condition SNOMED codes in the RR map to each configuration.

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Adds SNOMED CG code / display name to the creation modal search / confirmation flow

https://github.com/user-attachments/assets/5d5c2c0b-e5dc-419e-9e06-c128d6f63f90

## 🔗 Related Issue

Fixes #840 
- #840 

## 🧪 How to test
- Select a new config from the modal. Notice the new table that displays child RSG codes / displays. Note down one of the codes
- Clear the selection and search by the display name / code. Notice a search match renders for it now.
- Create the config and notice the creation works as intended


client/src/components/Combobox/index.tsx
```diff
@@ -157,7 +157,9 @@ function ComboboxOption<T>({
       {...props}
     >
       {(bag) => (
-        <span>{typeof children === 'function' ? children(bag) : children}</span>
+        <span className="w-full">
```
fzhao99, line 160: added this so the combo options would flex to the full width

before:
Image

after:
Image

client/src/components/Modal/index.tsx
```diff
@@ -22,11 +22,14 @@ function useModalContext() {
...
 }
 
+type WidthSettings = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
+
 interface ModalProps {
   open: boolean;
   onClose: () => void;
   children: React.ReactNode;
   position?: 'center' | 'top';
+  maxWidth?: WidthSettings;
```
fzhao99, line 32: found myself wanting to tweak the width of the modals, so added this prop to make them more composable relative to the Tailwind breakpoints

client/src/hooks/useSearch.ts
```diff
@@ -1,13 +1,19 @@
 import { useMemo, useState } from 'react';
 import Fuse, { FuseResult, IFuseOptions } from 'fuse.js';
 
-export function useSearch<T>(data: T[], options?: IFuseOptions<T>) {
+export function useSearch<T>(
+  data: T[],
+  options?: IFuseOptions<T>,
+  useExtendedSearch = false
```
fzhao99: needed a precise prefix match for the SNOMED searches, so extending this as an optional prop per the documentation [here](https://www.fusejs.io/extended-search.html)

refiner/app/db/codes/model.py
```diff
@@ -2,13 +2,21 @@
...
 
-@dataclass
-class DbCode:
+@dataclass(frozen=True)
+class CodedConcept:
+    """
+    Code / display name minimal representation of a codeable concept.
+    """
+
+    code: str
```
fzhao99, line 11: we have [a class in `terminology.py`](https://github.com/CDCgov/dibbs-ecr-refiner/blob/281d27d69aa40b94c9888fd4ded7b4c3ad8cb15d/refiner/app/services/terminology.py#L39) called `Coding` that uses "code" and "display" for the fields that were labeled "value" and "name" before. I went back and forth on it, but thought ultimately it would be good to align them so we aren't using different labels for the same concepts in different parts of the codebase, so made the necessary updates to do so.

Let me know if we would rather keep them separate or if we want them to relate to each other in a different way, since they are being used in slightly different places / contexts.

client/src/components/Modal/index.tsx
```diff
@@ -138,7 +143,7 @@ function ModalFooter({
   return (
     <div className={className}>
       <div
-        className={classNames('mx-auto flex w-full max-w-md gap-3 px-6', {
```
fzhao99, line 141: think the intention was to get the buttons to align appropriately in the flex columns, so think we can just rely on the parent max width + the `justify-` settings to achieve the alignment we want

client/src/pages/Configurations/index.tsx
```diff
@@ -178,7 +178,7 @@ function AppUpdateBanner({
           <Link
             to="/app-updates"
             onClick={handleViewUpdates}
-            className="font-public-sans text-violet-warm-60 border-violet-warm-60 flex h-11 items-center justify-center rounded-sm border-2 bg-white px-5 text-center text-[1rem] leading-[1.4rem] font-bold lining-nums proportional-nums no-underline"
+            className="font-public-sans bg-violet-warm-60 border-violet-warm-60 flex h-10 items-center justify-center rounded-sm px-5 text-center text-[1rem] leading-[1.4rem] font-bold text-white lining-nums proportional-nums no-underline"
```
fzhao99: the banner button was displaying weirdly so I went ahead and fixed these styles

Before:
Image

After:
Image

client/src/pages/Configurations/index.tsx
```diff
@@ -175,7 +179,7 @@ function AppUpdateBanner({
           <Link
             to="/app-updates"
             onClick={handleViewUpdates}
-            className="font-public-sans text-violet-warm-60 border-violet-warm-60 flex h-[44px] items-center justify-center rounded-[4px] border-[2px] bg-white px-[20px] text-center text-[1rem] leading-[1.4rem] font-bold lining-nums proportional-nums no-underline"
+            className={SECONDARY_BUTTON_STYLES}
```
fzhao99: this was displaying weirdly so opted just to pull these from the component styles

jakewheeler: Not that you added this component in your PR but I believe it should be like this:
```ts
<Button
  variant="secondary"
  to="/app-updates"
  onClick={dismissNotification}
>
  View updates
</Button>
```

This will render a link styled like the "secondary" button.

client/src/pages/Configurations/index.tsx
```diff
@@ -205,31 +209,39 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
+  const hasSearchResultsToDisplay = searchTextLongEnough && results;
 
   function reset() {
     onClose();
     setSelectedCondition(null);
   }
 
   return (
-    <Modal open={open} onClose={reset} position="top">
+    <Modal open={open} onClose={reset} position="center" maxWidth="xl">
```
jakewheeler, line 231: I liked `top` a lot better 🥲 Nothing for you to do - just noticed it immediately.

client/src/pages/Configurations/index.tsx
jakewheeler: I'd suggest the `Button` option I mentioned above over turning these styles into exported constants. At least for now.

client/src/pages/Configurations/index.tsx
```diff
@@ -248,30 +260,127 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
+                <thead className="sr-only">
+                  <tr>
+                    <th scope="col">Display name</th>
+                    <th scope="col">Code</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  {selectedCondition.rsg_codes.map((c) => {
+                    return (
+                      <tr>
```
jakewheeler: We'll need to add a `key` to this.

refiner/migrations/20260615135704_rename_codes_columns.sql
```diff
@@ -0,0 +1,26 @@
+-- migrate:up
+ALTER TABLE codes
+  RENAME COLUMN value TO code;
+
+ALTER TABLE codes
+  RENAME COLUMN name TO display;
```
jakewheeler, line 6: My immediate reaction to this was that `code` is an obvious improvement and `display` is more confusing. I think I still lean slightly towards `name` since it's more specific at a glance but I don't think it's too big of a deal either way if we prefer to keep `display`.

client/src/components/Modal/index.tsx
```diff
@@ -55,6 +58,7 @@ function Modal({
   onClose,
   children,
   position = 'center',
+  maxWidth = 'lg',
   className,
 }: ModalProps) {
   return (
```
jakewheeler, line 64: This might be the first case we've needed it but I noticed that I can't scroll inside the modal when we have a large table. It looks like adding `overflow-auto` to line 72 will fix it. If you want to try that and fix it here I'd appreciate it but we could also do it in a different PR.

refiner/migrations/20260615135704_rename_codes_columns.sql
fzhao99: I'll just go ahead and change it to `code` and `name` in both places.

fzhao99: actually realizing there are seralization concerns of changing things to name with the goal of keeping things consistent between the types, so I'll just leave `code` and `display` for now

client/src/components/Modal/index.tsx
fzhao99: think I added it but couldn't quite recreate your issue, so lmk if what I did addressed the concern

jakewheeler: Yes, looks good! I doubt it will happen often but I was testing with very limited vertical spacing and wasn't able to proceed to setting up the config. This does seem to work now.

client/src/pages/Configurations/index.tsx
```diff
@@ -171,14 +174,13 @@ function AppUpdateBanner({
...
             </span>
           </div>
-
-          <Link
+          <Button
+            variant="secondary"
             to="/app-updates"
-            onClick={handleViewUpdates}
-            className="font-public-sans text-violet-warm-60 border-violet-warm-60 flex h-[44px] items-center justify-center rounded-[4px] border-[2px] bg-white px-[20px] text-center text-[1rem] leading-[1.4rem] font-bold lining-nums proportional-nums no-underline"
+            onClick={(e) => handleViewUpdates(e)}
```
jakewheeler: I don't think there's any real need to keep `handleViewUpdates()` and `handleDismiss()` around anymore. The link will handle navigation on its own and these functions are otherwise just wrappers for `dismissNotification()`.
```suggestion
            onClick={dismissNotification}
```

client/src/hooks/useSearch.ts
jakewheeler: Any reason to do this rather than use `useExtendedSearch` as an option when calling the hook?
```ts
  const { searchText, setSearchText, results } = useSearch(
    conditions,
    {
      keys: [
        { name: 'display_name' },
        { name: 'rsg_codes.display' },
        { name: 'rsg_codes.code' },
      ],
      threshold: 0.25,
      distance: 500, // Broaden the distance so Fuse doesn't penalize long strings
      includeMatches: true,
      findAllMatches: true,
      useExtendedSearch: true, // <---
    }
  );
```

client/src/pages/Configurations/index.tsx
```diff
@@ -248,30 +258,127 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
-                  setQuery('');
+                  setSearchText('');
                 }}
+                placeholder="Start typing to search (3 characters minimum)"
               />
               <ComboboxOptions anchor="bottom">
                 {({ option: condition }) => (
                   <ComboboxOption key={condition.id} value={condition}>
-                    {condition.display_name}
+                    {hasSearchResultsToDisplay
```
jakewheeler: Took a shot at splitting this up into a couple of components. Looks right to me but please make sure I didn't lose anything 😓 

I split it into:

`ConditionOption`
```tsx
interface ConditionOptionProps {
  condition: ConditionSummary;
  result?: FuseResult<ConditionSummary>;
}

function ConditionOption({ condition, result }: ConditionOptionProps) {
  if (!result) return <>{condition.display_name}</>;

  const { item, matches } = result;
  const conditionDisplayMatch = matches?.find((m) => m.key === 'display_name');
  const rsgMatches = matches?.filter(
    (m) => m.key === 'rsg_codes.display' || m.key === 'rsg_codes.code'
  );

  return (
    <div>
      <p className="pb-2 font-bold">
        {conditionDisplayMatch
          ? highlightMatches(item.display_name, conditionDisplayMatch.indices)
          : item.display_name}
      </p>
      <div>
        {rsgMatches?.map((match, i) => (
          <RsgMatchRow key={i} match={match} rsgCodes={item.rsg_codes} />
        ))}
      </div>
    </div>
  );
}
```

RsgMatchRow
```tsx
interface RsgMatchRowProps {
  match: FuseResultMatch;
  rsgCodes: CodedConcept[];
}

function RsgMatchRow({ match, rsgCodes }: RsgMatchRowProps) {
  const matchedCode = rsgCodes[match.refIndex as number];
  const isDisplayMatch = match.key === 'rsg_codes.display';
  const isCodeMatch = match.key === 'rsg_codes.code';

  return (
    <div className="flex justify-between">
      <div>⤷</div>
      <p className="ml-1 flex-2 pb-2">
        {isDisplayMatch
          ? highlightMatches(matchedCode.display, match.indices)
          : matchedCode.display}
      </p>
      <p className="flex-1 text-right">
        {isCodeMatch
          ? highlightMatches(matchedCode.code, match.indices)
          : matchedCode.code}
      </p>
    </div>
  );
}
```

And then this can be used here like this:
```tsx
<ComboboxOptions anchor="bottom">
  {({ option: condition }) => {
    const result = results.find(
      (r) => r.item.id === condition.id
    );
    return (
      <ComboboxOption key={condition.id} value={condition}>
        <ConditionOption
          condition={condition}
          result={hasSearchResultsToDisplay ? result : undefined}
        />
      </ComboboxOption>
    );
  }}
</ComboboxOptions>
```

Feel free to adapt if you like this! I think this makes it a bit less overwhelming.

refiner/app/api/v1/conditions.py
```diff
@@ -6,57 +6,40 @@
...
 
-    conditions = await get_latest_conditions_db(db=db)
-    return [
-        GetConditionsResponse(
-            id=condition.id,
-            display_name=condition.display_name,
-        )
-        for condition in conditions
-    ]
+    return await get_conditions_with_rsg_codes_db(db=db)
```
jakewheeler: I might still advocate for using a response model here even though the shape currently matches what we want to return to the client. It's probably the _right way_ to go but I know there are some cases we don't follow this 100% (thinking of events).

I don't think there's a ton of risk but I can't see any reason not to use a response model. Might make the most sense to always do that by default and fix any places that don't currently do this. Curious on your thoughts here.

client/src/pages/Configurations/index.tsx
```diff
@@ -205,31 +207,39 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
+      threshold: 0.25,
+      distance: 500, // Broaden the distance so Fuse doesn't penalize long strings
+      includeMatches: true,
+      findAllMatches: true,
+    },
+    true
+  );
+  const searchTextLongEnough =
+    searchText.length > MIN_CONFIG_SEARCH_TEXT_LENGTH;
+  const hasSearchResultsToDisplay = searchTextLongEnough && results;
```
jakewheeler: I was thinking we could tweak this line slightly for clarity.
```suggestion
  const isSearching = searchTextLongEnough && results.length > 0;
```

and then make use of it on line 263:
```tsx
virtual={{
  options: isSearching ? results.map((r) => r.item) : conditions,
}}
```

client/src/pages/Configurations/index.tsx
```diff
@@ -205,31 +207,39 @@ function NewConfigModal({ open, onClose }: NewConfigModalProps) {
...
+    {
+      keys: [
+        { name: 'display_name' },
+        { name: 'rsg_codes.display' },
+        { name: 'rsg_codes.code' },
+      ],
+      threshold: 0.25,
+      distance: 500, // Broaden the distance so Fuse doesn't penalize long strings
+      includeMatches: true,
+      findAllMatches: true,
```
jakewheeler: We might be able to avoid a lot of the string length related logic by setting [`minMatchCharLength`](https://www.fusejs.io/fuzzy-search.html#minmatchcharlength) to `3`.

Let me know if I'm missing anything but this seems to be working for me.

client/src/pages/Configurations/index.tsx
jakewheeler: If we go with [this](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1390/changes#r3438682882) then this would just need to be:
```ts
const isSearching = results.length > 0;
```

refiner/app/api/v1/conditions.py
fzhao99: That's fine as a default for me. I can fix this one and then do the others in a followup

client/src/pages/Configurations/index.tsx
fzhao99: Doing it this way seems to make Fuse break when there are strings with spaces in them (note the resetting of the match results after spaces). I toyed around with a few settings to try to get it to work but couldn't seem to figure out how.

https://github.com/user-attachments/assets/47a5e422-04dc-4e3d-89f8-fb40c83b2cfc

Lmk if you can repro. It seems to work fine if we do the checking in React, so I might just leave it like that unless there's a setting you can find that makes it behave like we expect

refiner/app/api/v1/conditions.py
```diff
@@ -6,57 +6,49 @@
...
     """
-
-    conditions = await get_latest_conditions_db(db=db)
-    return [
-        GetConditionsResponse(
-            id=condition.id,
-            display_name=condition.display_name,
-        )
-        for condition in conditions
-    ]
```
jakewheeler, line 59: Apologies if my wording about the response model was confusing. I meant to suggest that we actually go back to this model where we explicitly map what we get from the DB into a response object that only has the properties we intend to provide to the client.

The current change doesn't really buy us anything since we are directly passing what we receive from the DB directly to the client still. We just have one more wrapper involved now.

client/src/pages/Configurations/index.tsx
jakewheeler: I'm seeing the same thing. I think we'd need to tweak the hook to account for a minimum string length which we don't need to do here, so let's leave it as you have it.

client/e2e/configurations.spec.ts
```diff
@@ -18,9 +18,43 @@ test.describe('Configurations screen', () => {
       })
     ).toBeVisible();
     await expect(page.getByRole('searchbox')).not.toBeVisible();
+    const setupConfigurationButton = page.getByRole('button', {
```
jakewheeler, line 21: Nice tests 👍

jakewheeler (approved):
Looks good! 👏

PR #1392, opened 2026-06-17
refactor: replace pipeline `trace` with separate inputs and outputs

jakewheeler:
# 🔀 PULL REQUEST

> [!IMPORTANT]
> I reviewed the augmentation piece of refining and decided it's fine as it is. I focused on refactoring the input/outputs of the pipeline instead.

## 💡 Summary

This PR removes the concept of `RefinementTrace`, which was an object that was partially built up to be used as an input, built up the rest of the way by the refinement process, and then provided back as an output. This can be somewhat confusing at times since it's difficult to tell if a value has been "collected" yet or not.

In place of `RefinementTrace` there is now a dedicated input (`RefinementContext`) and dedicated outputs (`RefinementDocuments`, `RefinementMetrics`, and `RefinementReport`).

With these changes, it should now be very clear what entry point function `refine_for_condition` requires as an input and what information is guaranteed to you by its output.

## ⛑️ What does this change?

This is primarily just a refactor but does remove a few properties from the previous `trace` object that didn't provide value.

Please take a look at the updated snapshots to see which properties were removed and ensure you agree with my decisions here. Anything I removed was due to us already being able to acquire this info in another way.

## 🔗 Related Issue

Fixes #1168 
- #1168

## 🧪 How to test

- Refining continues to work as it did before
  - This includes both from Lambda and within the app
- Automated tests continue to pass


refiner/app/services/pipeline.py
```diff
@@ -223,12 +204,38 @@ def discover_reportable_conditions(
 # NOTE:
 # STAGE 2: REFINEMENT EXECUTION
 # =============================================================================
+class RefinementException(Exception):
```
jakewheeler: Is there a better place to put this? It is not used anywhere else.

refiner/app/services/pipeline.py
```diff
@@ -290,20 +293,6 @@ def refine_for_condition(
...
     """
 
-    trace.configuration_resolved = True
-
-    # * canonical_url participates in the deterministic ID derivation
-    # via its trailing uuid; it must be resolved before refinement
-    # begins
-    # * callers (testing.py, lambda_function.py) set this when
-    # they resolve the RSG -> grouper mapping before calling
-    if trace.canonical_url is None:
```
jakewheeler, line 300: `context` requires `canonical_url` so this is no longer possible.

refiner/app/services/pipeline.py
```diff
@@ -111,63 +108,49 @@ def create_augmentation_run_from_xml_files(
...
+    size_mib: float
+
+
+@dataclass
+class RefinementMetrics:
+    """
+    The metrics calculated during the refinement process.
+    """
+
+    eicr: RefinementMetricsEicr
```
fzhao99, line 141: could you say a little more about the value of bundling the eICR metrics into a wrapper like this is over just having a flat object?

refiner/app/services/pipeline.py
fzhao99: only other thought is to put it in `app/core/exceptions.py` (which we should probably make a decision to either use more consistently or remove entirely)

refiner/app/lambda/lambda_function.py
```diff
@@ -635,25 +615,49 @@ def process_condition(
...
 
+    log_refinement_summary(
+        rsg_code=rsg_code,
+        jurisdiction_code=jurisdiction_code,
+        metrics=result.metrics,
+        report=result.report,
+    )
+
+
+def log_refinement_summary(
```
fzhao99: is it worth having this function still to just call it once in the lines above it? Seems like it's more common to have the orchestration code just call `logger.info` directly in other places

refiner/app/services/pipeline.py
fzhao99: the other two objects make a lot of sense to me though: think we've broken the trace object up in a pretty intuitive way

refiner/app/services/pipeline.py
jakewheeler: I'll move it there for now!

refiner/app/services/pipeline.py
jakewheeler: We were previously scoping metrics by name (`eicr_size_mib`, `eicr_size_reduction_percentage`) so this change just improves upon that and gives us a cleaner access pattern (`eicr.size_mib` instead, for example). Not necessarily the main intention but this does also give us an organized way to add non-eICR related metrics if we end up wanting to collect other data within `RefinementMetrics` later.

The report and the documents are scoped as they were before and they just happen to be flat since we weren't scoping them at all.

refiner/app/lambda/lambda_function.py
jakewheeler: That's a good point 🤔 I'll delete it and just call `logger.info()` directly.

fzhao99 (approved):

PR #1409, opened 2026-06-22
fix: app updates page should use `ExternalLink` component

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR modifies the app updates page to use the standard `ExternalLink` component, which is used for links that navigate away from Refiner to other websites.

This does cause a slight change in appearance
image

## 🧪 How to test

- Links to releases on the app updates page should still work the same as before
- Links will appear with expected icon denoting navigation to a different website
- Automated tests continue to work as before


rogeruiz (approved):
lgtm 🌈

PR #1415, opened 2026-06-23
chore: pin dev dependencies

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

Pins the dependencies defined in `refiner/app/lambda/dev-requirements.txt` and `refiner/dev-requirements.txt` to explicit versions.

## 🧪 How to test

- Automated tests continue to pass
- Lambda container works as expected locally 
- App container works as expected locally


refiner/app/lambda/dev-requirements.txt
fzhao99, line 1: should we do the same thing in the app dev requirements too?

jakewheeler: Probably 👀 I didn't notice we didn't pin them there either. I can do that as well.

fzhao99 (approved):

PR #1420, opened 2026-06-24
docs: updating active configs in s3

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary

- propose solution of using ops container to run a migration to update all active configs in s3

## 🔗 Related Issue

Fixes #1248 

- #1248 

## ✅ Acceptance Criteria

 - [ ] Figure out and document a viable solution to this problem
 - [ ] Ensure limitations of solution are well-outlined
 - [ ] Provide a prototype / sample code if feasible


docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
```diff
@@ -0,0 +1,221 @@
...
+```py
+activate_configuration_db()
+upload_condition_mapping_payload()
+upload_current_version_file()
+```
+
+The configuration is already active, so the migration should not change its status, activation history, configuration version, condition mapping, or `current.json`.
+
+### Migration flow
+
```
rogeruiz, line 142: I think we should have a note or a paragraph about needing to address the race condition where the new Lambda is deployed before the ops container finishes migrating S3. I think something along the lines as being explicit that we're going to leverage SQS visibility timeouts/retries. That way any Lambda hits to `IncompatibleActiveConfigurationError` during the migration window will raise a transient error and let the message return to the queue to be retried.

rogeruiz: Maybe something like this?


```suggestion

To avoid deployment locks, the system will use SQS retries as a natural buffer. Lambda will be updated to raise transient errors on schema mismatch, allowing eCRs to be retried until the ops container completes the S3 migration.

```

docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
```diff
@@ -0,0 +1,221 @@
...
+```text
+Ops container starts
+    ↓
+Database migrations run
+    ↓
+TES data is loaded
```
rogeruiz, line 148: Should we have an additional step here? I'm thinking something to avoid unnecessary S3 writes and log noise. I think we should introduce a `SCHEMA_VERSION` constant in the code and a tracking table in Postgres.


```suggestion
```text
Ops container starts
    ↓
Query Postgres migration table for `current_schema_version`. If there's a match with `CODE_SCHEMA_VERSION`, skip the migration.
    ↓
Database migrations run
    ↓
TES data is loaded
```

docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
```diff
@@ -0,0 +1,221 @@
...
+- configuration version
+- jurisdiction
+- condition UUID
+- S3 key
+- schema version written
+- migration success or failure
+
+Lambda should log similar information when it finds a missing, invalid, or incompatible active file, including the expected and actual schema versions.
+
+This gives the team visibility if a migration is missed or fails in production.
```
rogeruiz, line 215: I think we need to clarify this section a bit more around the failure strategy. To prevent a single corrupt configuration from blocking the entire deployment pipeline, let's be explicit about a 'partial success' approach.


```suggestion
Lambda should log similar information when it finds a missing, invalid, or incompatible active file, including the expected and actual schema versions.

If an individual configuration fails to rebuild, the migration will log a critical error for that specific ID but continue with the remaining configurations. The global migration status in the DB will be marked as `PARTIAL_FAILURE` to trigger an alert for manual intervention.

This gives the team visibility if a migration is missed or fails in production.
```

docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### `docs: document active payload migrations`
+
+Document the difference between configuration versions and active payload schema versions and describe the process for future schema changes.
+
+### Open Questions and Next Steps
+
+- Enumerate possible race conditions when the ops migration runs while Lambda is processing eCRs.
+- Confirm whether the ops container can read and write Postgres.
+- Confirm whether the ops container can read and write S3 configuration artifacts.
```
rogeruiz: I think we have answers for these now right? The infra team confirmed that the ops container can be provisioned with the required IAM roles for S3 `PutObject`/`GetObject` and Postgres access.

rogeruiz (commented):
The core logic is spot on here, @lina-roth! Good work on this. I left some suggestions to make things a bit more explicit.

docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
fzhao99: Worth calling out here also that the application and ops containers run at the same time, so we'll need to ensure the implementation doesn't rely on them finishing sequentially.

docs/decisions/0011_2026-06-24_updating-active-configs-in-s3.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+This version is only for the shape of `active.json`. It is separate from the existing configuration version used in the S3 path and `current.json`.
+
+Lambda should validate this value after loading the file. If the version is not supported, it should raise and log a clear compatibility error instead of only returning a Pydantic validation error.
+
+```py
+if data.get("schema_version") != CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION:
+    raise IncompatibleActiveConfigurationError(...)
+```
```
fzhao99, line 84: would love to also see some of the discussion points we had about using the serialization function in an integration test (perhaps asserting against an artifact of some sort), and having the dev in question bumping the schema version whenever there's a failure on that test.

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+    )
+
+
+def remove_maintenance_lock() -> None:
+    """Remove the maintenance lock after a successful migration."""
+
+    delete_maintenance_lock()
+
+
+async def wait_for_lambda_to_drain(
```
lina-roth: This is something we will want to note, not only is it migration run time - but we will need lambdas executing to stop execution before we are able to start

jakewheeler (approved):
🫡

PR #1425, opened 2026-06-24
chore: Create and use custom icon components

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR swaps out all `@trussworks/react-uswds` icon components for our own icons. This aids our effort in moving off of this UI library and going all in on Tailwind/HeadlessUI.

We also had a good amount of SVG files in the `assets/` directory that were not being used. I've gone ahead and deleted those as a part of this work.

> [!IMPORTANT]
> We only have three truly reusable icon components that now live in `@components/Icons`.  One-off components are defined in the file in which they are being used. If they must become reusable later on we can make that change in a new PR.

## 🧪 How to test

- All app icons should appear exactly as they did before
- No more `<Icon.Name..../>` components should appear in the code


rogeruiz (approved):
lgtm 🌈 

Thanks for doing this!

PR #1310, opened 2026-06-02
chore: refactor CSV bulk upload

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
As part of #1160 started refactoring the CSV bulk upload route. Split that out into a follow up here that refactors the bulk upload process to better follow code base patterns

## 🔗 Related Issue

Fixes #1419
- #1419

## ✅ Acceptance Criteria
- [ ] CSV bulk upload code is more cleanly factored and maintainable
- [ ] More tests exist for the edge case rendering cases
- [ ] No regressions throughout the experience

PR #1432, opened 2026-06-25
chore: add csv bulk upload tests

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Part one of CSV bulk upload that adds some tests and does some misc renaming, split out for ease of review

More complicated refactor coming up in [a followup PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1310)
## 🔗 Related Issue

Part one of #1419 
- #1419

## ✅ Acceptance Criteria

- All CSV bulk upload functionality should remain the same

If you want to test the backwards compatibility of the old templates, you can use the below file:
[custom_code_upload_template (15).csv](https://github.com/user-attachments/files/29513040/custom_code_upload_template.15.csv)

## ℹ️ Additional Information

Let me know if there are edge cases we want to cover with tests to add to this PR


client/e2e/pages/ConfigurationPage.ts
```diff
@@ -71,7 +71,7 @@ export class ConfigurationPage {
     await this.page
       .getByRole('button', { name: 'Add new custom code' })
       .click();
-    await this.page.getByLabel('Code #').fill(code);
+    await this.page.getByLabel('Code').fill(code);
```
fzhao99: since we were using `Code` in the database table, updated this since `Code #` isn't used anywhere else (and is _technically_ inaccurate)

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -224,11 +224,16 @@ async def upload_custom_codes_csv(
...
 
-    required_columns = {"code_number", "code_system", "display_name"}
-    if not required_columns.issubset(set(csv_reader.fieldnames or [])):
+    headers = set(csv_reader.fieldnames or [])
+    base_required = {"code_system", "display_name"}
+
+    has_base = base_required.issubset(headers)
+    # maintain backwards compatibility with old template that used "code_number" as the column header
+    has_code_variant = "code" in headers or "code_number" in headers
+    if not (has_base and has_code_variant):
```
fzhao99: let me know if we think this is a bad idea: just didn't want to unnecessarily break existing spreadsheets that some of the phase 1 folks might have already.

I've also factored this more cleanly [in the upcoming PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1310/changes)

refiner/app/api/v1/configurations/model.py
```diff
@@ -3,7 +3,7 @@
 
 from pydantic import BaseModel, Field
 
-from app.db.code_systems.db import CodeSystemIndex
+from app.db.code_systems.db import IndexedCodeSystem
```
fzhao99, line 6: I got a little confused as to whether `CodeSystemIndex` was referring to the index of the code system, so swapped the order of the adjectives. Let me know if folks prefer the old name

client/src/utils/fixtures.ts
```diff
@@ -0,0 +1,154 @@
+import { DbCodeSystem } from '../api/schemas';
+import { DbConfigurationCustomCode } from '../api/schemas/dbConfigurationCustomCode';
+import { DbTotalConditionCodeCount } from '../api/schemas/dbTotalConditionCodeCount';
+import { GetConfigurationResponse } from '../api/schemas/getConfigurationResponse';
+import { GetConfigurationResponseVersion } from '../api/schemas/getConfigurationResponseVersion';
+import { IndexedCodeSystem } from '../api/schemas/indexedCodeSystem';
+
+export const MOCK_CONFIG_ID = 'd8cf3930-a7c2-4761-9ba9-ce72ff9191c8';
```
fzhao99: found myself reusing a bunch of fixtures across the new frontend tests, so just went ahead and centralized them. Let me know if folks think this is a bad pattern

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    ...actual,
+    useGetCodeSystems: vi.fn(() => ({
+      data: {
+        data: mockCodeSystems,
+      },
+    })),
+  };
+});
+
+describe('Custom codes upload', () => {
```
fzhao99: tried to seed the tests here that cover the "preview" set of interactions to track all the edge cases in the frontend state, leaving any content / validation-related that need the whole flow for the e2e's in the upcoming PR.

let me know if folks can think of other edge cases we want tests for here or in the e2e's. The main one are the row-level error reporting that I think I'd want to break out into a separate e2e.

client/e2e/configuration.spec.ts
```diff
@@ -438,6 +437,27 @@ test.describe('Configuration detail flow', () => {
 
       await expect(makeAxeBuilder).toHaveNoAxeViolations();
     });
+  });
+
+  test('User drafts and activates a new configuration', async ({
```
fzhao99: I broke this into two separate test steps because (at least locally) the old mega-test was starting to consistently time out at the ~30 second mark

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/Modals.tsx
```diff
@@ -133,7 +133,7 @@ export function PreviewEditModal({
       <ModalBody>
         <div>
           <Field>
-            <Label>Code #</Label>
+            <Label>Code</Label>
```
jakewheeler, line 136: Why is this changing?

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
```diff
@@ -437,6 +436,7 @@ export function ImportCustomCodes({
             accept=".csv"
             hidden
             onChange={handleFileChange}
+            aria-label="Bulk custom code upload file input"
```
jakewheeler: This won't be read by a screen reader since the element is hidden so I'd say it's fine to exclude it.

refiner/app/api/v1/configurations/custom_codes.py
jakewheeler: Why is this changing? It does seem like a more fitting name to me, I'm just unsure why it's happening as a part of this refactoring.

fzhao99: Mostly because I realized we were using `Code` in the database table and because we weren't using `Code #` (and is _technically_ inaccurate) isn't used anywhere else, it was in my head as "things to generally clean up" so I went ahead and changed it. 

Do realize it's a user facing change though so happy to get a sign off from product if we feel we should

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+export const EXAMPLE_OTHER_CODE_SUFFIX = '1534';
+export const EXAMPLE_OTHER_CODE = EXAMPLE_CVX_CODE + EXAMPLE_OTHER_CODE_SUFFIX; // used to test prefix matching
+
+export const CSV_DOWNLOAD_TEMPLATE = `code,code_system,display_name
+${EXAMPLE_SNOMED_CODE},SNOMED,SNOMED Example
+${EXAMPLE_OTHER_CODE},Other,Other Example
+6789,ICD-10,ICD-10 Example
+${EXAMPLE_LOINC_CODE},LOINC,LOINC Example
+${EXAMPLE_CVX_CODE},CVX,CVX Example
+198440,RxNorm,RxNorm Example
```
jakewheeler: This may change in your other PR but I would suggest that we:
1. Move `CSV_DOWNLOAD_TEMPLATE` into `ImportCustomCodes.tsx` and don't export it
2. Turn `CSV_DOWNLOAD_TEMPLATE` into a literal or move the variables into `ImportCustomCodes.tsx` and don't export them
3. Delete this file

The reason being that these variables are being used for testing and I'd suggest we avoid doing this. Doing something like this:
```tsx
      expect(
        screen.getByText(`Edit ${EXAMPLE_SNOMED_CODE}`)
      ).toBeInTheDocument();
    });
```
can bite us as we are just checking that the variable we defined is equal to itself rather than writing out what we expect the value to be. We actually saw this  happen during some file sizing work we were doing in the UI if I recall correctly.

tldr: I'd recommend we don't use shared constants in tests so that we can be very explicit about what we expect the values to be.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
fzhao99: I was having trouble getting this element in the frontend tests to test the file upload and so added this label so I could `getByLabel` it. Would we prefer for me to use `getByTestId` here, or do you know of another way to get hidden inputs from vitest?

refiner/app/api/v1/configurations/model.py
```diff
@@ -217,9 +217,9 @@ class UploadCustomCodesCsvInput(BaseModel):
 class UploadCustomCodesPreviewItem(BaseModel):
     """Validated CSV row ready for confirmation."""
 
+    id: UUID
```
jakewheeler, line 228: Why are we adding this?

client/src/utils/fixtures.ts
jakewheeler: A decent amount of these seem to be used for one specific file, in which case I'd suggest we move the structure directly into the file using it.

For the others (a couple shared between `ConfigBuild`, `ConfigTest`, etc.) I think we could move the data closer to where it's being used.

For example, `mockCustomCodes` is used by both `ConfigBuild` and `ConfigTest`. In this case we could put that data in `pages/Configurations/test/fixtures.ts` or something along those lines.

client/e2e/configuration.spec.ts
jakewheeler: I'm surprised we are just now seeing this 🤔 We could bump the timeout up a bit if we want to keep the mega test.

The mega test is kind of nice because it will let us know if other steps are breaking something along the way. This test skips just about everything and goes straight to activing which I personally don't find to be very useful since it's not likely users will be doing this.

client/e2e/configuration.spec.ts
```diff
@@ -505,33 +558,41 @@ test.describe('Configuration detail flow', () => {
...
+        .getByRole('searchbox', { name: 'Search codes' })
+        .fill('test code_name');
+
+      await editButton.click();
       await expect(
         page.getByRole('heading', { name: `Edit ${testCode}`, level: 2 })
       ).toBeVisible();
-      await page.getByRole('button', { name: 'Save changes' }).click();
-
+      await page.getByLabel('Close this window').click();
```
jakewheeler: We can get this by role which looks to be what we're doing in a couple of other tests.
```suggestion
      await page.getByRole('button', { name: 'Close this window' }).click();
```

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const user = userEvent.setup();
```
jakewheeler: This should be defined per test rather than once in the suite so it's not being shared between tests. There are details on this in the [docs](https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent).

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    );
+    expect(uploadCsvButton).toBeInTheDocument();
+
+    const file = new File([CSV_DOWNLOAD_TEMPLATE], 'test.csv', {
+      type: 'text/csv',
+    });
+
+    await user.upload(uploadCsvButton, file);
+
+    uploadCsvButton.click();
```
jakewheeler: I think this was meant to be
```suggestion
    await userEvent.click(uploadCsvButton);
```

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+      >
+        <TestQueryClientProvider>
+          <ImportCustomCodes configurationId={MOCK_CONFIG_ID} />
+        </TestQueryClientProvider>
+      </MemoryRouter>
+    );
+    expect(await screen.findByText('Import from CSV')).toBeInTheDocument();
+    const uploadCsvButton = screen.getByLabelText(
+      'Bulk custom code upload file input'
+    );
```
jakewheeler: One thing to note about this pattern in react testing library is that saving a query to a variable does not re-query the DOM before it gets used, which Playwright does do.

In a lot of cases it likely doesn't matter but did want to mention it since PW and RTL look extremely similar to one another but don't actually behave the same way. A lot of times I end up querying directly each time I want an element in RTL just to avoid potential issues with this.

refiner/app/api/v1/configurations/model.py
jakewheeler: I don't have an issue with it either way. This is intended to go away once we move to using system IDs, right?

refiner/app/api/v1/configurations/custom_codes.py
jakewheeler: Yeah, it's a good change. Just want to make sure they're good with it since I was still seeing `Code #` in the recent designs. I also wouldn't mind that being its own PR.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
jakewheeler: That makes sense. In this case I'd say a test ID is fine. If you could make a comment about it being used due to the element being hidden that'd be great.

client/e2e/configuration.spec.ts
fzhao99: rolled it back together and then added a timeout

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/utils.ts
fzhao99: I've gone ahead and pulled in a function that dynamically generates the CSV when past a list of the code systems so that the template is driven by the backend code system config. This was done in response to a user request to have a row per system as an example.

This requires some mocking / variable references since I made the template code number generation random. The mocks aren't being shared across tests, but we are using some constants within the file for readability, so let me know if what I've done conflicts with your suggestions.

refiner/app/api/v1/configurations/model.py
fzhao99: yep!

refiner/app/api/v1/configurations/model.py
fzhao99: I needed a way of identifying the preview item without referring to any of its constituent elements (which is how the item is being identified now).

I've gone ahead and made a slight tweak [to use the ID as a key](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1432/changes#diff-3dcfba97458016a65f4a1b5b47ee540a5e1381bd40e5a5a3115a2792c747429eR562) in order to not have an empty property, but will eventually consume it to do some id-based filtering in state updaters in the future.

client/e2e/pages/ConfigurationPage.ts
jakewheeler: Sorry, I apparently skipped right over this comment 😂 you answered my question about this below.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
```diff
@@ -232,12 +236,15 @@ export function ImportCustomCodes({
...
+    if (!codeSystems?.data) return;
+
+    const uploadTemplateCsvContent = buildCsvDownloadTemplate(
+      Object.values(codeSystems.data)
+    );
 
-    const blob = new Blob([csv], { type: 'text/csv;charset=utf-8;' });
+    const blob = new Blob([uploadTemplateCsvContent], {
+      type: 'text/csv;charset=utf-8;',
+    });
```
jakewheeler, line 247: Is there a specific reason we construct the CSV client-side?  We don't do this elsewhere so I'm wondering why it was done here. I realize you didn't change this in your PR. It just stuck out to me and am curious if you know.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function buildCsvDownloadTemplate(systemsSupported: DbCodeSystem[]) {
+  const headers = 'code,code_system,display_name';
+
+  let content = headers + '\n';
+  systemsSupported.forEach((s) => {
+    const randomLengthAtLeastThree = 3 + Math.floor(Math.random() * 10);
+    const randomCode = Array.from({ length: randomLengthAtLeastThree }, () =>
+      Math.floor(Math.random() * 10)
+    ).join('');
```
jakewheeler: I think I'd prefer this to not be random for the sake of simplicity. I wouldn't expect a template file to be different each time I download it and am not sure what the benefit is.

client/src/utils/fixtures.ts
fzhao99: done! lmk if the new organization is in line with your suggestion

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
fzhao99: hmm, guessing it was just because it was simpler to have a CSV string and create it on the fly. Don't think it was an intentional decision.

Guessing prefer us making it on the backend and packaging having the frontend read it from the response?

client/src/pages/Configurations/ConfigBuild/index.test.tsx
```diff
@@ -692,42 +676,4 @@ describe('Config builder page', () => {
       screen.getByText('Import from CSV', { selector: 'h2' })
     ).toBeInTheDocument();
   });
-
-  it('downloads the custom code upload template CSV', async () => {
```
fzhao99, line 835: deleted this guy because we're covering it in an e2e / it's not testing anything super useful

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/utils.ts
fzhao99: makes sense. would something like `111111-{system_key}` be an ok replacement for the code values? Or would we want something else

refiner/app/api/v1/configurations/model.py
jakewheeler: I think the main part that's throwing me off is the [client needing to send a random ID to the server](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1432/changes#diff-3dcfba97458016a65f4a1b5b47ee540a5e1381bd40e5a5a3115a2792c747429eR24). Is this something we can avoid?

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
jakewheeler: I would expect the server to be in control of file generation for sure. Totally fine for this to come later if we'd prefer not too change too much else here.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/utils.ts
jakewheeler: Yes, I think so. Or just a few handcrafted examples of what kind of typical format would be acceptable. I don't know that we explicitly define what is allowed anywhere in the UI 🙃

refiner/app/api/v1/configurations/model.py
fzhao99: ahh ok. don't actually think we need to generate one on the fly given it's the empty state. Went ahead and removed it.

I do get rid of that null state in favor of an actual null value [in the followup](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1310/changes#diff-3dcfba97458016a65f4a1b5b47ee540a5e1381bd40e5a5a3115a2792c747429eL22).

refiner/app/api/v1/configurations/custom_codes.py
fzhao99: updated both the code #  --> code and code name --> display name per product feedback

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
fzhao99: I'll handle this in a followup to move this to the server.

Think it might be worth setting up regex for each of the systems / use that to generate example strings so that the values are something you can see in real life.

client/e2e/fixtures/api.ts
```diff
@@ -77,7 +77,7 @@ export class Api {
     const payload = [];
     let row = 2;
     for (const c of codes) {
-      payload.push({ ...c, row });
+      payload.push({ id: crypto.randomUUID(), ...c, row });
```
jakewheeler: Are we planning to revert this? It looks like the issue is this model:
```python
class ConfirmUploadCustomCodesInput(BaseModel):
    """Payload used to confirm a previously validated CSV import."""

    custom_codes: list[UploadCustomCodesPreviewItem]
```

This is the input to `/confirm` which includes IDs when it doesn't need to. We should use a model similar to `UploadCustomCodesPreviewItem` without requiring an ID. 

It looks like the new input would only need to require `code`, `system_key`, and `name` to work.

client/e2e/configuration.spec.ts
```diff
@@ -269,6 +269,7 @@ test.describe('Configuration detail flow', () => {
     configurationsPage,
     configurationPage,
   }) => {
+    test.setTimeout(120000);
```
jakewheeler: I wouldn't be opposed to bumping up the local timeout in `playwright.config.ts` if we wanted to take that route. It's currently at 30 seconds but we could bump it to 45 or just set it to 60 for both local and CI?

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
```diff
@@ -232,12 +236,15 @@ export function ImportCustomCodes({
...
   const handleDownloadTemplate = () => {
-    const csv = `code_number,code_system,display_name
-12345,Other,Other Example
-6789,ICD-10,ICD-10 Example
-99999A,LOINC,LOINC Example`;
+    if (!codeSystems?.data) return;
+
+    const uploadTemplateCsvContent = buildCsvDownloadTemplate(
+      Object.values(codeSystems.data)
+    );
```
jakewheeler, line 243: I think this is fine for now if we'll be changing it in the follow up.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
```diff
@@ -437,6 +445,9 @@ export function ImportCustomCodes({
             accept=".csv"
             hidden
             onChange={handleFileChange}
+            // defined as a test ID since the input element is hidden and we
+            // needed a hatch to grab it via the frontend unit tests
+            data-testid="bulk-upload-file-input"
```
jakewheeler, line 450: 👍

client/src/pages/Configurations/ConfigBuild/CustomCodes/CsvImport/ImportCustomCodes.tsx
```diff
@@ -462,7 +473,11 @@ export function ImportCustomCodes({
         ) : step === 'intro' ? (
           <div className="w-full max-w-xl space-y-6">
             {renderCsvInstructions()}
```
jakewheeler, line 475: No need to handle this now but maybe worth noting for the following up. This can be a component instead of a JSX-returning function.

client/e2e/fixtures/api.ts
fzhao99: Do we think the change that we're doing on [line 559](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1432/changes#diff-3dcfba97458016a65f4a1b5b47ee540a5e1381bd40e5a5a3115a2792c747429eR559) of the `ImportCustomCodesCsv` page is worth the change here? From what I understand, the keys on those props should be driven from ID's delivered via data. The composite-value key seems a bit of an anti-pattern since those values might change as we edit the frontend state

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -638,7 +645,7 @@ async def validate_custom_code(
         db (AsyncDatabaseConnection, optional): The database connection
 
     Returns:
-        bool: Returns True if the code name has not been used, otherwise returns False
+        bool: Returns True if the code  has not been used, otherwise returns False
```
jakewheeler:
```suggestion
        bool: Returns True if the code has not been used, otherwise returns False
```

client/e2e/fixtures/api.ts
jakewheeler: Let me know if I have this wrong at all 👀 

When we upload the CSV (`/upload`) we receive back `preview_items` with this structure:
```json
"preview_items": [
        {
            "id": "60f142e6-531a-4782-85b0-047e2cf20d09",
            "code": "1115565",
            "system_key": "snomed",
            "name": "SNOMED Example",
            "row": 2
        },
       ...
]
```
This `id` is what we use for the `key` which seems fine to me. However, when go to actually save the data (`/confirm`) we're sending `id` (and `row`) which are not needed by the server.

Does that make sense? It seems like we're attempting to reuse `UploadCustomCodesPreviewItem` for two different purposes: preview (sending data back to the UI) and saving (committing to the DB).

client/e2e/configuration.spec.ts
fzhao99: done!

client/e2e/fixtures/api.ts
fzhao99: ah seeing how I misunderstood you now. agreed that we can simplify what the client is sending back to the server.

I've broken the client / server sides into two models that inherit from each other. Lmk if this is what you were thinking of

jakewheeler: Excellent 💯 thank you for making this change!

jakewheeler (approved):
Thanks Bob! 🎉

PR #1445, opened 2026-06-26
feat(server): Point the eICR External Document Reference in the RR to its refined eICR pair 

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

We're mutating the eICR External Document Reference in the RR so that it points to its refined eICR pair. This will make the link explicit. This refined RR historically was just a filtering out of information that isn't relevant. This will not mutate part of the refined RR. Doing this in `augmentation.py` made more sense because we need the augmented identifiers. Additionally, this has to happen in `pipeline.py` because that is when all of the info will be in context. There may have been a better seam for this but with the time crunch to get this in dev/test by monday this feels like the cleanest place for it.

Also, snapshots were updated because the RR content is not different (all refined RR's point to their refined pair).

## 🔗 Related Issue

Fixes #1444 

- #1444


refiner/app/services/ecr/augment.py
```diff
@@ -35,6 +35,13 @@
 # document source label -> from data augmentation document source value set (Vol 2 Table 3)
 ORIGINAL_DOCUMENT_SOURCE: Final[str] = "original-document"
 
+# template identifiers for locating the eICR external document reference
+# inside the refined RR:
+#   * received eICR information act (RR IG STU 1.1 Vol 2 §3.13)
+#   * eICR external document reference it contains (Vol 2 §3.8.1)
+RECEIVED_EICR_INFO_ACT_TEMPLATE_ROOT: Final[str] = "2.16.840.1.113883.10.20.15.2.3.9"
```
fzhao99, line 42: is this value one that we got from the sample files Robert Willams sent us? Can't seem to cross reference where these came from, at least [in the sections of the IG](https://aphl8515.sharepoint.com/:w:/r/sites/DIBBsToolDevelopmentCollaboration/_layouts/15/Doc.aspx?sourcedoc=%7BCC0250DA-EDF6-4E00-B989-9BFC227B8EA3%7D&file=Volume%202%20Electronic%20Case%20Reporting%20(eICR%20%26%20RR)%20Data%20Augmentation%20(Version%205).docx&action=default&mobileredirect=true) that I was looking in.

I did find the external doc template ID in 2.16 of the [other volume](https://aphl8515.sharepoint.com/:w:/r/sites/DIBBsToolDevelopmentCollaboration/_layouts/15/Doc.aspx?sourcedoc=%7BDBF01D6E-60FB-4108-AD1F-E4F0F757AACD%7D&file=Volume%201%20DataAugmentationStrategy%20(Version%205).docx&action=default&mobileredirect=true), but can't seem to track this one down

refiner/app/services/ecr/augment.py
```diff
@@ -523,6 +530,132 @@ def augment_rr(
...
+        _insert_before_first_found(external_document, new_set_id, ["hl7:versionNumber"])
+
+    # versionNumber — 1..1 SHALL (CONF:3315-738); replace, or append as
+    # the last element in the content model
+    new_version = _make_element("versionNumber", value=eicr_version_value)
+    old_version = external_document.find("hl7:versionNumber", HL7_NS)
+    if old_version is not None:
+        _replace_preserving_tail(external_document, old_version, new_version)
+    else:
+        external_document.append(new_version)
```
fzhao99, line 625: it's still strange to me that they want us to just grab the version from the original eCR and stamp it (as opposed to incrementing it to indicate something's happened) but 🤷

fzhao99 (approved):
one question about the provenance of a template ID, but otherwise lgtm!

refiner/app/services/ecr/augment.py
robertmitchellv: these aren't from the sample files; they're from the base RR IG:

- `…15.2.3.9` Received eICR Information (the act): section 3.13
- `…15.2.3.10` eICR External Document Reference: section 3.8.1

the augmentation IG only references `…15.2.3.10` (to set its values for a refined pair); it never mentions the containing act, so `…15.2.3.9` only lives in the RR IG. the samples just happen to contain both--the RR IG is the actual source. we anchor on the act so we grab the right `externalDocument` and not one of the other ones an RR can carry (even tho i'm not sure how often other `externalDocument` references come through without more visibility into the data jurisdictions receive)

refiner/app/services/ecr/augment.py
robertmitchellv: yeah, initially i wanted to use some kind of decimal for version and keep the set identity but that would break the CDA rules. since these are unique at the point of their document identity and set identity we can think of them as sort of the parallel world condition specific "version 2" eICR/RR pairs. the version ends up describing that it's a part of this group of changes from the original

Issue #1227, opened 2026-05-15
[USER STORY] Reconstruct Medications Administered section narrative


kplaneaux:
## 👤💭 User Story

As a user, I can choose to reconstruct the Medications Administered narrative section from the coded data, so that I can make sure I'm getting only the data I want

## 🏗️ Acceptance Criteria

- [ ] Reconstruct option appears in the drop-down for Medications Administered section
- [ ] Narrative reconstructed from the coded data replaces existing narrative

## 🎨 Designs
https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=9197-46703&t=T7WfiQLMRfFQqrTN-1

PR #1446, opened 2026-06-26
feat(server): Adds narrative reconstruction to Problems, Immunizations, and Medications Administered

robertmitchellv:
# 🔀 PULL REQUEST

Adding reconstruction for the Problems, Immunizations, and Medications Administered sections. Also changing the way we handle panel and specimen related data for the Results section as well. We're taking a slightly different approach by grouping panel/organizer data by their own table in the narrative. This will scale easier for sections like Plan of Treatment, which is coming next.

## 🔗 Related Issue

Fixes #1226 #1227 #1228

- #1226 
- #1227 
- #1228 

## 🧪 How to test

Unit tests

1. Ensure dependencies are all installed with `just server install-dev`
1. Test specific paths within the `refiner/` directory with just server test -vv tests/`
1. Additionally you can check coverage with `just server test --cov=app tests`

or

Testing the application:

1. Run `just dev up`
1. Check that all services are up and running with `docker ps`
1. In your browser, navigate to [http://localhost:8081](http://localhost:8081)
   to test the application
1. You should be able to see the reconstructed narrative for the sections that you configured thusly.


refiner/app/services/ecr/specification/constants.py
```diff
@@ -25,20 +25,30 @@
 # entry match rule traceable to a named code system without having to
 # memorize OIDs
 
+CPT4_OID: Final[str] = "2.16.840.1.113883.6.12"
 CVX_OID: Final[str] = "2.16.840.1.113883.12.292"
 ICD10_OID: Final[str] = "2.16.840.1.113883.6.90"
 LOINC_OID: Final[str] = "2.16.840.1.113883.6.1"
+NCI_OID: Final[str] = "2.16.840.1.113883.3.26.1.1"
+NDC_OID: Final[str] = "2.16.840.1.113883.6.69"
```
fzhao99, line 33: When would it be worthwhile to add these to the seed script for systems now that we're starting to check for them in the refinement passes? Would imagine that we'd want to have the option to have custom codes in these systems if we're checking for them in the reconstruction passes

refiner/app/services/ecr/specification/constants.py
```diff
@@ -25,20 +25,30 @@
...
-CODE_SYSTEM_LABELS: Final[dict[str, str]] = {
+# human-readable names resolved from the codeSystem OID, never the source
+# codeSystemName attribute, which is unreliable and inconsistently populated
+# (missing, or "SNOMED-CT" vs "SNOMED CT"). used both for PHA-facing narrative
+# display and for the internal entry-match provenance comments--one OID->name
+# map, using the technically correct names (e.g. "SNOMED CT", "ICD-10-CM")
+CODE_SYSTEM_DISPLAY_NAMES: Final[dict[str, str]] = {
     LOINC_OID: "LOINC",
-    SNOMED_OID: "SNOMED",
+    SNOMED_OID: "SNOMED CT",
```
fzhao99, line 45: we should also try to keep these aligned with the in-application display names that we're showing in the app if we're also going to display them in the reconstructed data. 

The one issue that's front of mind is that we use display name in some of the custom code upload validation, so there's probably a need to swap that into using the display key so that they're stable / consistent. Happy to do this in a followup though since it's changing things in a few different places

refiner/app/services/ecr/specification/constants.py
robertmitchellv: these really are only here for the narrative reconstruction tables--it just so happens that there was some example data using some of these systems so i folded it in. i'd really like the users to decide what systems we end up folding in next (or we take a different approach to curation). either way, hopefully this can just be a user directive. i don't think we'll need these systems in the db but they may be in translations or other parts of the reconstructed narrative, which is why i added them

refiner/app/services/ecr/specification/constants.py
robertmitchellv: it's hard; technically it's "SNOMED CT" and i'm opting for it specifically so that it's consistent in CDA for the narrative reconstruction. at some point it would be good to align everything but i think a more general "SNOMED" isn't bad in the app. but in the outputs it should probably be more correct. in the export csv, i think just "SNOMED" is better but we can always see if it bothers any of the users

fzhao99 (approved):

PR #1461, opened 2026-06-29
chore: switch from `python:3.14-slim` to `python:3.14-alpine` base image

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the base images of the app and ops images to use Alpine instead of the Python slim images, which are based on Debian. We are swapping over to Alpine in an effort to eliminate as much software bloat as possible so that we can limit vulnerabilities and decrease image sizes.

> [!IMPORTANT]
> This change does not impact the Lambda prod image or development images.

## ✅ Acceptance Criteria

- The application and ops containers should continue to function as they did previously

## 🧪 How to test

I'd recommend running the CI setup locally, which will build and use the production ops container and app container. When the setup completes, you can click around the app as normal at http://localhost:8081/.

Shut down your local environment and run this command:

```
docker compose -f docker-compose.yml -f docker-compose.ci.yml up --build
```

### DIFF
--- a/Dockerfile.app
+++ b/Dockerfile.app
@@ -17,9 +17,28 @@ ENV VITE_APP_VERSION=$VERSION
 COPY ./client ./
 RUN npm run build
 
+# Install system deps
+FROM python:3.14-alpine AS system-builder
+
+RUN apk add --no-cache \
+    gcc \
+    g++ \
+    musl-dev \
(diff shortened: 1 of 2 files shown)


fzhao99 (approved):
🚢

PR #1463, opened 2026-06-30
chore: remove USWDS dependencies from Toast

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the `Toast` component to no longer rely on imports from `@trussworks/react-uswds`. 

The `Toast` should look and function about the same as before.

### Success

We can test the "success" version of the component by adding/removing code sets from a configuration.

Example toast:
image

### Error

The easiest way to see that the "error" version works is to create a configuration, go back to the configurations page, and then try to create that same configuration again. This will show an error toast since a draft for that condition already exists.

Example toast:
image


client/src/tailwind.css
```diff
@@ -134,7 +134,7 @@
   --color-state-warning-lighter: #fff5c2;
   --color-state-warning: #ddaa01;
   --color-state-warning-darker: #5c4809;
-  --color-state-error-lighter: #fff3f2;
+  --color-state-error-lighter: #f4e3db;
```
jakewheeler, line 137: Not sure if this changed at some point but [this](https://www.figma.com/design/QK7dVoNaG6qIF52r1imkwY/eCR-Refiner?node-id=4216-49037&m=dev) appears to be the color we want.

rogeruiz (approved):
lgtm 🌈 

🫡

PR #1464, opened 2026-06-30
feat: Split out references for version into a separate TES table

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a `tes` table that tracks versions of the TES imported into the Refiner's database.

- Migration script to add the `tes` table and backfill its data
  - Also adds `tes_id` to `conditions` and links them appropriately 
  - `created_at`/`updated_at` are based on the time the conditions related to that version were imported
- Updates to `load_static_data.py` to upsert new TES versions and associate its ID with related conditions

> [!NOTE]
> This PR does not remove `version` from `conditions`. I'll do this in a follow-up PR since this will require app changes.

## 🔗 Related Issue

Fixes #1347

- #1347

### Migrations

1. With your data already loaded, run `just migrate local`
2. Take a look at the `tes` table, `SELECT * FROM tes`. You should see 6 rows (one per version)
3. Take a look at the `conditions` table, `SELECT * FROM conditions`. You'll see that every condition row has a `tes_id` which links to its TES version

### Script

Trying running the data loading script: 

`python scripts/seeding/load_static_data.py`

Nothing will change since the migration has backfilled the table for all current versions. If you'd like to see the script add a missing version, run these queries:

```sql
DELETE FROM conditions_context_groupers
WHERE condition_id IN (
    SELECT id FROM conditions WHERE version = '4.0.0'
);
DELETE FROM conditions WHERE version = '4.0.0';
DELETE FROM tes WHERE version = '4.0.0';
```

This will delete all condition data and the TES record associated with version `4.0.0`. Run the data loader script and you should see:

- All `4.0.0` condition data is added
- TES record for `4.0.0` is added
- IDs of existing `tes` and `conditions` records are unchanged


refiner/migrations/20260630173611_add_tes_table.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE TABLE tes (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    version TEXT NOT NULL UNIQUE,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- populate tes with the distinct versions from conditions
+INSERT INTO tes (version, created_at, updated_at)
+SELECT version, MIN(created_at), MIN(created_at) -- use condition insert values
```
jakewheeler, line 13: This will backfill the data using the time the conditions for a TES version were inserted.

refiner/migrations/20260630173611_add_tes_table.sql
```diff
@@ -0,0 +1,41 @@
+-- migrate:up
+
+-- create the tes table
+CREATE TABLE tes (
```
jakewheeler, line 4: This data seemed to be what is relevant currently but let me know if I'm missing anything.

fzhao99 (approved):
realizing reviewing this that we somehow didn't get rid of the `child_rsg_snomed_code` column on conditions even though we're using the new join table on that now.

I'll work on cleaning that up as well

PR #1469, opened 2026-07-02
feat(debug): add serialized configuration view

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR introduces a development tool to the app that allows you to quickly and easily view the activation files stored in S3 for any active configuration in the jurisdiction associated with your account. See discussion about this [here](https://skylight-hq.slack.com/archives/C08LC62CVNE/p1783024810929369?thread_ts=1782496534.025479&cid=C08LC62CVNE).

When navigating to an active configuration's build page you'll see a `View serialized configuration` link. When you click on this link you'll be taken to a page that gives you the full contents of the `active.json` file, the `metadata.json` file, and the `current.json` file. `active.json`'s content shows by default since that's the file we'll likely be checking most often.

This link only appears on the build page and only for active configurations.

> [!IMPORTANT]
> This is currently only available when running the application locally. Another PR adding support for the dev environment will come after this is merged.

## 📺 Demo

Kapture 2026-07-07 at 09 26 17

### Functionality

1. Navigate to any active configuration
2. Click on the link `View serialized configuration (local only)`
3. You'll be taken to a new page with each files name, the full path, and the contents of each file
4. Click the link at the top to go back to the configuration's build page

### Local only

This feature only exists locally for now. This is somewhat difficult to check due to the way auth works locally.

- `useGetEnv` is used to determine whether to render the link or not, which we also use for the local-only test file upload
- The route raises a 403 if `ENVIRONMENT["ENV"] != "local"`. This is a check we use in various other areas of the code

Note that it is possible to directly navigate to the new page in a live environment if someone happens to know the URL, however, error text will be displayed immediately and nothing can be done besides returning to the build page.

Here is an example of what this would look like:
image


fzhao99 (approved):
thanks for getting this in Jake!

PR #1488, opened 2026-07-06
feat: ops container migration to regenerate activations in s3

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary

As a user, I need to be able to use active configurations without having to manually reactivate every time something changes with the activation model.

## 🔗 Related Issue

Fixes #1494 

- #1494 

#### schema version
 - [ ] Add a schema_version field to newly generated active.json files and define the current supported schema version
  - [ ] Add a Postgres table that records the target schema version, migration status, timestamps, success count, and failure count.
#### lambda
  - [ ] Add an S3 maintenance lock that is created by the ops migration and checked by Lambda before processing each SQS record.
  - [ ] Add structured logging and metrics when Lambda encounters a missing, invalid, or incompatible active.json file
#### ops container
  - [ ] Add a database query that returns all currently active configurations and the information needed to rebuild their S3 artifacts
 - [ ]  Create a shared service that rebuilds and uploads an active configuration without reactivating it or changing its activation history
  - [ ] Add an ops command that queries active configurations and regenerates their existing active.json files in S3.
  - [ ] Run the active payload migration after database migrations and TES data loading are complete (Added to entrypoint prepare-db)
#### test
  - [ ] Add an integration test that generates an active payload using convert_config_to_storage_payload() and compares the serialized result against an approved fixture, snapshot, or JSON Schema. (see notes)
  - [ ] Verify that regeneration updates the S3 payload without changing the configuration version, status, activation history, condition mapping, or current.json.
#### prototyping / proof of concept
  - [ ]  Create 100 configs and test locally with regeneration

### Test with 100 activated configurations 
1. `docker compose up`

2. import 100 configuration db

Don't run migrate or seed. Just run this command to add tables and contents.

Download, unzip, upload
[100-activated-configs-db.sql.zip](https://github.com/user-attachments/files/30134559/100-activated-configs-db.sql.zip)

```
psql \
  -h localhost \
  -p 5432 \
  -U postgres \
  -d refiner \
  -f 100-activated-configs-db.sql
```

3. copy exported activations back into s3 

these were activated before i added schema version so they are ready to be activated by just running the ops command. if you want to test after that just increment schema version again

may need to `brew install awscli` 

[local-s3-configs.zip](https://github.com/user-attachments/files/30134525/local-s3-configs.zip)
```
AWS_ACCESS_KEY_ID=refiner \
AWS_SECRET_ACCESS_KEY=refiner \
AWS_DEFAULT_REGION=us-east-1 \
aws \
  --endpoint-url=http://localhost:4566 \
  s3 cp \
  ./local-s3-configs/ \
  s3://local-config-bucket/configurations/ \
  --recursive
```

4. ` docker compose run --rm ops prepare-db` or `docker compose run --rm ops regenerate-active-configs`

You probably need to rebuild ops container if you built before the container before these changes or entrypoint script has changed - - - `docker compose build --no-cache ops`

5. View results 

[example_logs.txt](https://github.com/user-attachments/files/30134833/example_logs.txt)

`http://localhost:4566/local-config-bucket/configurations/SDDH/07221093-b8a1-4b1d-8678-259277bfba64/1/active.json`

### Results from my local testing 

100 configs regenerated in ~1.5 seconds
~0.015 seconds per config
~67 configs per second


ops/entrypoint.sh
```diff
@@ -40,6 +40,10 @@ case "$COMMAND" in
         echo "Importing static data"
         exec python3 ./scripts/seeding/load_static_data.py
         ;;
+    regenerate-active-configs)
+        echo "Regenerating active configuration files"
+        exec python3 ./scripts/migrations/regenerate_active_configs.py "$@"
+        ;;
```
jakewheeler, line 61: Is this working for you? It seems to fail immediately for me.
```sh
docker run --network host \                  
  -e DB_URL=postgresql://postgres@localhost:5432/refiner \
  -e DB_PASSWORD=refiner \
  -e SSL_MODE=disable \
  updconfig regenerate-active-configs
Regenerating active configuration files
  File "/app/./scripts/migrations/regenerate_active_configs.py", line 33, in <module>
    from app.db.configurations.db import get_configurations_db
ModuleNotFoundError: No module named 'app'
```

lina-roth: sorry about that, fixed

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+)
+from app.services.configurations import (
+    convert_config_to_storage_payload,
+    get_config_payload_metadata,
+)
+
+logger = logging.getLogger(__name__)
+
+MAINTENANCE_LOCK_KEY = "configurations/maintenance.lock"
+MIGRATION_NAME = "active-payload-schema-v2"
```
jakewheeler: I would prefer not to reuse the word `migrations` since we already have a `refiner/migrations` that is related to database migrations. Maybe just `reactivation(s)`?

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+    )
+
+
+def remove_maintenance_lock() -> None:
+    """Remove the maintenance lock after a successful migration."""
+
+    delete_maintenance_lock()
+
+
+async def wait_for_lambda_to_drain(
```
jakewheeler: The ops container doesn't currently have any permissions to interact with Lambda in any way FWIW.

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+
+    lock_expiration_minutes = get_integer_environment_variable(
+        "MAINTENANCE_LOCK_EXPIRATION_MINUTES",
+        DEFAULT_LOCK_EXPIRATION_MINUTES,
+    )
+
+    migration_limit = get_integer_environment_variable(
+        "ACTIVE_CONFIG_MIGRATION_LIMIT",
+        0,
+    )
```
jakewheeler: Certainly not set in stone but I imagine any values like this will end up being hard-coded in this script.

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+    get_config_payload_metadata,
+)
+
+logger = logging.getLogger(__name__)
+
+MAINTENANCE_LOCK_KEY = "configurations/maintenance.lock"
+MIGRATION_NAME = "active-payload-schema-v2"
+
+DEFAULT_LAMBDA_DRAIN_SECONDS = 60
+DEFAULT_LOCK_EXPIRATION_MINUTES = 60
```
jakewheeler: My hunch is that this is too long of a lock time but I don't know what "right" length of time would be. 15 mins?

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+
+    try:
+        return int(raw_value)
+    except ValueError as exc:
+        raise RuntimeError(
+            f"{name} must be a valid integer. Received: {raw_value}"
+        ) from exc
+
+
+def get_lock_expiration(lock: dict[str, object]) -> datetime | None:
```
jakewheeler: Definitely worth defining a type to represent a `Lock`.

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+
+    now = datetime.now(UTC)
+
+    lock_payload = {
+        "reason": "active_configuration_migration",
+        "migration": MIGRATION_NAME,
+        "owner": "ops-container",
+        "started_at": now.isoformat(),
+        "expires_at": (now + timedelta(minutes=expiration_minutes)).isoformat(),
+    }
```
jakewheeler: One downside of using the lock file is that it's another file that we need to read in and validate the format of before we can do anything with it.

It looks like we only use `expires_at` programmatically so it's not _too bad_. And you've handled that case.

Another option could be writing an empty file and the "expired time" could be 15 mins in the future of the [last modified time](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#SysMetadata). Not sure it makes a huge difference one way or the other but figured it's worth mentioning.

refiner/scripts/migrations/regenerate_active_configs.py
```diff
@@ -0,0 +1,590 @@
...
+            "total": total,
+            "successful": successful,
+            "failed": len(failures),
+        },
+    )
+
+    if failures:
+        raise RuntimeError(
+            "Failed to regenerate active configurations: " + ", ".join(failures)
+        )
```
jakewheeler: I wonder if we should do retries? And how many times would we retry?

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: Yeah so this just adds a small set delayed timer, currently 60 seconds, so that if the lockfile is placed, it allows any lambda currently processing to potentially finish what it's doing before we start regenerating configs - hopefully so it wouldnt mess up something currently in process

refiner/scripts/migrations/regenerate_active_configs.py
lina-roth: good idea

refiner/scripts/migrations/regenerate_active_configs.py
lina-roth: the empty-file approach is simpler, but then the expiration logic depends on S3 object metadata / LastModified, which works but I think for early versions having more explicit logging around Why is Lambda pausing? Who created this? When did it start?When should it expire? What operation is running? could be useful for debugging

refiner/scripts/migrations/regenerate_active_configs.py
lina-roth: yeah this is a good call, renamed to reactivation over migration

refiner/scripts/migrations/regenerate_active_configs.py
lina-roth: added some retry logic

refiner/app/lambda/lambda_function.py
```diff
@@ -270,6 +300,18 @@ def lambda_handler(event, context) -> dict:
...
 
+            except MaintenanceModeError as e:
+                # Do not write RefinerComplete for maintenance mode.
+                # Returning the record as a batch failure allows SQS to retry it
+                # after the queue visibility timeout.
+                logger.warning(
+                    "Deferring record because active configuration maintenance is in progress.",
+                    operation="active_configuration_maintenance",
+                    persistence_id=persistence_id,
+                    error=str(e),
```
jakewheeler:
```suggestion
                    exception=e,
```

refiner/dev-requirements.txt
```diff
@@ -15,3 +15,4 @@ requests==2.34.2
 rich==15.0.0
 types-Authlib
 types-requests
+jsonschema
```
jakewheeler:
```suggestion
jsonschema==4.26.0
```

ops/entrypoint.sh
```diff
@@ -49,6 +53,8 @@ case "$COMMAND" in
         dbmate --no-dump-schema --migrations-dir ./migrations --url "$DATABASE_URL" migrate
         echo "Migration step complete"
         exec python3 ./scripts/seeding/load_static_data.py
+        echo "Regenerating active configuration files"
+        exec python3 ./scripts/migrations/regenerate_active_configs.py
```
jakewheeler: This isn't going to work. `exec` replaces the current process so once we hit `exec python3 ./scripts/seeding/load_static_data.py` and that runs, the ops container will end.

```suggestion
        dbmate --no-dump-schema --migrations-dir ./migrations --url "$DATABASE_URL" migrate
        echo "Migration step complete"
        python3 ./scripts/seeding/load_static_data.py
        echo "Regenerating active configuration files"
        python3 ./scripts/migrations/regenerate_active_configs.py
```

However, I'm seeing a different issue when I try to run this after the fix:
```sh
INFO:config:⏳ Upserting code <> child RSG relationships...
INFO:config:🏁 Done!
INFO:config:⏱️  Static data loaded in 12.343 seconds
Regenerating active configuration files
  File "/app/./scripts/migrations/regenerate_active_configs.py", line 34, in <module>
    from app.db.configurations.db import get_configurations_db
  File "/app/app/db/configurations/db.py", line 24, in <module>
    from app.services.logger import get_logger
  File "/app/app/services/logger.py", line 7, in <module>
    from ..core.config import ENVIRONMENT
  File "/app/app/core/config.py", line 21, in <module>
    "ENV": get_env_variable("ENV"),
           ~~~~~~~~~~~~~~~~^^^^^^^
  File "/app/app/core/utils.py", line 19, in get_env_variable
    raise OSError(f"Missing environment variable: {name}")
OSError: Missing environment variable: ENV
```

What environment variables does this process require? We may need to figure out a way around the ops container importing `ENVIRONMENT`.

refiner/app/lambda/lambda_function.py
```diff
@@ -46,6 +48,8 @@
 S3_BUCKET_CONFIG = get_env_variable("S3_BUCKET_CONFIG")
 S3_ENDPOINT_URL = os.getenv("S3_ENDPOINT_URL")  # No need to set this in a live env
 
+MAINTENANCE_LOCK_KEY = "configurations/maintenance.lock"
```
jakewheeler: Is there somewhere we could centralize this and then import it here and in `regenerate_active_configs.py`?

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,788 @@
...
+        return
+
+    logger.info(
+        "Waiting for existing Lambda executions to finish.",
+        extra={"drain_seconds": drain_seconds},
+    )
+
+    await asyncio.sleep(drain_seconds)
+
+    logger.info("Lambda drain wait complete.")
```
jakewheeler: What does this do in practice? It seems like:

1. Create maintenance lock
2. Hard 30 second wait
3. Move on to regeneration

I think I'm not understanding what is supposed to be happening during the wait.

refiner/tests/fixtures/activations/active_config_payload_schema_v1.json
```diff
@@ -0,0 +1,81 @@
+{
```
jakewheeler, line 1: Should this live in `tests/fixtures` instead?

refiner/app/db/configurations/model.py
```diff
@@ -11,6 +11,8 @@
 
 type DbSectionType = Literal["standard", "custom"]
 
+CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION = 1
```
jakewheeler, line 16: If this needs to be updated because the schema has changed, what will happen to let us know this value needs to be modified?

refiner/app/lambda/lambda_function.py
```diff
@@ -494,7 +599,23 @@ def read_configuration_file(s3_client, bucket: str, key: str) -> dict:
...
+            key=key,
+            expected_schema_version=CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
+            actual_schema_version=schema_version,
+        )
+        raise IncompatibleActiveConfigurationError(
+            "Active configuration schema version is incompatible. "
+            f"Expected {CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION}, got {schema_version}."
+        )
+
+    return configuration
```
jakewheeler, line 627: Probably worth making the error checking slightly more specific. We can check `schema_version` isn't `None`, and then that it's an `int`, and then finally that it matches the active version.

We don't perform validation until we call `ProcessedConfiguration.from_dict(serialized_configuration)` later on so this feels like a decent extra safety later.

```suggestion
    configuration = parse_s3_content_to_dict(config_file_content)
    schema_version = configuration.get("schema_version")

    if schema_version is None:
        raise IncompatibleActiveConfigurationError(
            "Active configuration is missing required field 'schema_version'."
        )

    if not isinstance(schema_version, int):
        raise IncompatibleActiveConfigurationError(
            f"'schema_version' must be an int, got {type(schema_version).__name__}."
        )

    if schema_version != CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION:
        logger.error(
            "Active configuration schema version is incompatible.",
            operation="active_configuration_schema_mismatch",
            key=key,
            expected_schema_version=CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
            actual_schema_version=schema_version,
        )
        raise IncompatibleActiveConfigurationError(
            "Active configuration schema version is incompatible. "
            f"Expected {CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION}, got {schema_version}."
        )

    return configuration
```

refiner/app/lambda/lambda_function.py
```diff
@@ -666,7 +787,7 @@ def process_condition(
 
     logger.info(
         "Refinement complete for condition.",
-        jurisidiction_code=jurisdiction_code,
+        jurisdiction_code=jurisdiction_code,
```
jakewheeler, line 799: This was 100% me 💀 thanks.

Dockerfile.ops
```diff
@@ -57,6 +59,11 @@ COPY ./refiner/migrations ./migrations
 COPY ./refiner/scripts/seeding ./scripts/seeding
 COPY ./refiner/scripts/data/source-tes-groupers ./scripts/data/source-tes-groupers
 
+# Copy active configuration regeneration script
+COPY ./refiner/scripts/migrations ./scripts/migrations
+
+COPY ./refiner/app ./app
```
jakewheeler: This is not ideal but I don't necessarily think we need to fix it in this PR. For now maybe we can scope it slightly better like we do for the Lambda image? This is what we're doing there.

```dockerfile
# Copy refiner modules
COPY ./refiner/app/core app/core
COPY ./refiner/app/services app/services
COPY ./refiner/app/db app/db
```

That way we could at least exclude some of the things we know the ops container won't need.

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,788 @@
...
+    successful = 0
+    failures: list[str] = []
+
+    logger.info(
+        "Starting active configuration regeneration.",
+        extra={
+            "configuration_count": total,
+            "limit": limit,
+            "target_schema_version": CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
+        },
```
jakewheeler: Anything in `extra` will never be printed because of how the logger is set up in this file. We use `JsonFormatter` in the app to handle JSON logs. 

Take a look at [setup_logger()](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/services/logger.py#L48) to see how we do that configuration there.

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,788 @@
...
+            logger.exception(
+                "Failed to regenerate active configuration.",
+                extra={
+                    "configuration_id": configuration_id,
+                    "configuration_version": configuration.version,
+                    "jurisdiction_id": configuration.jurisdiction_id,
+                },
+            )
+        else:
+            successful += 1
```
jakewheeler: I had no idea this was a thing 🤔 I had to google what this does. Should we just put `successful += 1` directly under `regenerate_active_configuration()` in the `try` block and get rid of `else` since it's not doing anything else?

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,788 @@
...
+    """
+    Run active configuration reactivation using an existing database connection.
+
+    This is separated from main() so integration tests can exercise the
+    reactivation tracking behavior without creating a new DB connection from env.
+    """
+
+    latest_complete_schema_version = (
+        await get_latest_complete_reactivation_schema_version_db(db=db)
+    )
```
jakewheeler, line 647: Should we do an explicit `None` check on this just in case?

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,788 @@
...
+        if lock_created:
+            logger.error(
+                "The reactivation failed while the maintenance lock was active. "
+                "The lock remains in S3 and must be reviewed before processing "
+                "is resumed.",
+                extra={
+                    "bucket": S3_CONFIGURATION_BUCKET_NAME,
+                    "key": MAINTENANCE_LOCK_KEY,
+                },
+            )
```
jakewheeler, line 752: Reviewed by who? What is the specific process that would need to happen if we received this error?

refiner/app/lambda/lambda_function.py
lina-roth: this is a good idea

refiner/tests/fixtures/activations/active_config_payload_schema_v1.json
lina-roth: do you mean just not in the `/lambda` folder of `tests/fixtures`?

refiner/app/db/configurations/model.py
lina-roth: If the active payload shape changes, this test will fail `refiner/tests/integration/test_activations.test_active_config_payload_matches_approved_schema`. On that test there is comments 

>          """
>         Verify that generated active.json payloads match the approved schema.
> 
>         This test should fail when the serialized active payload shape changes
>         unexpectedly. For an intentional shape change, the developer should:
> 
>         1. Review the serialization diff.
>         2. Increment CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION.
>         3. Update the approved schema fixture.
>         4. Confirm the reactivation script and Lambda validation support the new version.
> 
>         The schema fixture should not update automatically.
>         """

Which should cover that process of needing to update the test and increment the schema version

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: The idea is that we place the lock, but any lambda that had already started processing has some time to finish before we start updating activations and potentially causing additional failures. This timer could be set to 0 if we end up not wanting it, I can remove it. It was just as an additional precaution.

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: I rewrote it so it was more clear but it was basically doing this 

```
for each active configuration:
    try to regenerate it

    if regeneration fails:
        add that configuration ID to failures
        log the error

    if regeneration succeeds:
        add 1 to the successful count
 ```

File: refiner/scripts/reactivations/regenerate_active_configs.py (hunk 22 above)
Comment (lina-roth) [reply refiner/scripts/reactivations/regenerate_active_configs.py:752]: I think that this error should not really ever happen since the lock has an expiration timer. I guess this is just in the case that somehow it got stuck. I think reviewed is maybe not the best word - but maybe best to leave it so we could analyze what happened. But this would hopefully signal to APHL that we are in a weird state and they or we could look into why the lock file got stuck

File: refiner/scripts/reactivations/regenerate_active_configs.py (hunk 21 above)
Comment (lina-roth) [reply refiner/scripts/reactivations/regenerate_active_configs.py:647]: it should be ok, 

Because if the helper returns None, this comparison is just:

`    if latest_complete_schema_version == CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION:
`

None == 2

which evaluates to False, so the script continues and creates a new tracking record.

File: refiner/scripts/reactivations/regenerate_active_configs.py (hunk 19 above)
Comment (lina-roth) [reply refiner/scripts/reactivations/regenerate_active_configs.py]: Ah thank you

File: refiner/app/lambda/lambda_function.py (hunk 12 above)
Comment (lina-roth) [reply refiner/app/lambda/lambda_function.py]: good idea I added it alongside current schema version 

```
from app.db.configurations.model import (
    CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
    DbConfiguration,
    MAINTENANCE_LOCK_KEY,
)
```

File: ops/entrypoint.sh (hunk 11 above)
Comment (lina-roth) [reply ops/entrypoint.sh]: Just like the Lambda container, Ops needs these env vars for `regenerate_active_configs.py` to run properly, Im not sure what else could be done here

Comment (lina-roth) [reply ops/entrypoint.sh]: Also, this is working perfectly fine for me, curious what might be going on. I think you probably need to blow away the ops container you have currently built `docker compose build --no-cache ops`, perhaps all of them?

File: Dockerfile.ops (hunk 18 above)
Comment (lina-roth) [reply Dockerfile.ops]: had to add `app/api` too, not sure if its still an improvement or if this is basically just the same as importing `/app` now

File: refiner/app/lambda/lambda_function.py [hunk 23]
```diff
@@ -201,6 +214,24 @@ def lambda_handler(event, context) -> dict:
...
+                    s3_client=s3_client,
+                    bucket=s3_config_bucket_name,
+                )
+
+                if maintenance_lock is not None:
+                    logger.warning(
+                        "Active configuration maintenance is in progress.",
+                        operation="active_configuration_maintenance",
+                        lock_key=MAINTENANCE_LOCK_KEY,
+                        migration=maintenance_lock.get("migration"),
```
rogeruiz: Should this be `reactivation` now?

refiner/app/lambda/lambda_function.py
```diff
@@ -308,6 +351,67 @@ def lambda_handler(event, context) -> dict:
...
+        )
+    except ClientError as e:
+        error_code = e.response.get("Error", {}).get("Code")
+
+        if error_code in {"404", "NoSuchKey", "NotFound"}:
+            return None
+
+        raise
+
+    lock = parse_s3_content_to_dict(response["Body"].read().decode("utf-8"))
```
rogeruiz: If the lock file is "" or null, `parse_s3_content_to_dict` will raise a `json.JSONDecodeError`. This falls through as a fatal error and writes a `RefinerCompleteError` / `batchItemFailure`. We should do something similar to the catch in `regenerate_active_configs.py` for `json.JSONDecodeError`. I think an unparseable lock should be treated as expired/no-lock.

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,766 @@
...
+                status="PARTIAL_FAILURE",
+                success_count=result["successful"],
+                failure_count=failure_count,
+            )
+            tracking_record_finalized = True
+
+            raise RuntimeError(
+                "Failed to regenerate active configurations: "
+                + ", ".join(result["failures"])
+            )
```
rogeruiz: I think we should truncate this error since if we have a lot of failures those UUIDs will grow pretty big. For the error message maybe we only have the first ten results?

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,766 @@
...
+            return False
+
+        raise
+
+    try:
+        lock = json.loads(response["Body"].read())
+    except (json.JSONDecodeError, TypeError) as exc:
+        raise RuntimeError(
+            "The existing maintenance lock contains invalid JSON."
+        ) from exc
```
rogeruiz: Maybe we should remove corrupt locks and treat them as expired. Since we can't verify the corruption anyway. We should retry instead.

refiner/scripts/reactivations/regenerate_active_configs.py
```diff
@@ -0,0 +1,766 @@
...
+            "target_schema_version=%s latest_complete_schema_version=%s",
+            CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
+            latest_complete_schema_version,
+        )
+        return
+
+    reactivation_id = await create_reactivation_tracking_record_db(
+        db=db,
+        target_schema_version=CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION,
+    )
```
rogeruiz: We should move this down so that we're creating a lock & then a tracking record. That way the tracking record is only created once the lock is confirmed to exist. With the tracking record becoming the "source of truth" for whether the lock was held, we'll have a clearer error message from the script that we need to remove the stale lock.

rogeruiz (dismissed):
❌ This PR needs changes ...

Good work on this so far. I have some suggestions and nits

refiner/tests/fixtures/activations/active_config_payload_schema_v1.json
jakewheeler: Yeah I mean outside of `lambda` since it isn't being used in there at all.

refiner/scripts/reactivations/regenerate_active_configs.py
jakewheeler: I think if we can't explicitly check what's running Lambda wise then it's not worth having. I imagine most of the time it's going to be a wait with no tangible benefit since the new Lambda image is deployed immediately, Lambda functions could be running and still have multiple minutes remaining until completion, etc.

Let me know if I'm thinking about this incorrectly.

refiner/scripts/reactivations/regenerate_active_configs.py
jakewheeler: If this were to happen wouldn't all Lambda processing be blocked for 15 mins? Is the idea that the ops container be re-run within this window of time?

refiner/app/db/configurations/model.py
jakewheeler: How does the update process work for #3?

ops/entrypoint.sh
jakewheeler: If I want to run the ops container not using docker compose, what environment variables need to be provided? I'm wondering if it truly is everything the app needs or it's only a small subset. [This](https://github.com/CDCgov/dibbs-ecr-refiner/tree/main/ops) is what it needs currently for reference.

If it doesn't need everything we could try to adjust how the variables get pulled in so that'll allow a subset.

It works running in docker compose since all of the app's environment variables exist there.

Dockerfile.ops
jakewheeler: This needs to be reworked more in the future for sure but it's still an improvement IMO.

refiner/app/lambda/lambda_function.py
lina-roth: good catch

refiner/app/lambda/lambda_function.py
lina-roth: an invalid/empty/null lock should not become a fatal processing error for the eCR. i've updated the read to treat it like “no usable active lock,” log a warning, and continue.

lina-roth: good suggestion

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: good idea. adding a constant for MAX_FAILURE_IDS_IN_ERROR = 10 and updating logging statement to utilize that

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: Made some updates. I think treating a corrupt lock as expired/stale makes sense.

`create_maintenance_lock()` already retries once when an existing lock is removed. So if `remove_expired_maintenance_lock()` deletes a corrupt lock and returns `True`, the script will try to create the lock one more time.

That gives us the behavior we want:

* Valid, unexpired lock → do not overwrite
* Expired lock → delete and retry once
* Corrupt/unreadable lock → delete and retry once

I would not add any extra retries. The existing single retry keeps timing controlled and still protects against races if another process creates the lock first.

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: I updated this so we don’t leave behind a stale prod lock that requires manual intervention.

If the lock is created but the tracking insert fails, we now remove the lock immediately and fail loudly. That gives us a clear error without blocking Lambda behind a maintenance lock that we can't easily remove

refiner/app/db/configurations/model.py
lina-roth: To update the approved schema fixture, compare the new generated `active.json` payload against the existing JSON schema and update the difference.

For example, if the payload schema changes from v1 to v2:

1. Increment `CURRENT_ACTIVE_CONFIG_SCHEMA_VERSION`.
2. Update the schema fixture title/filename to match the new version.
3. Update `schema_version.const` to the new version.
4. Update required fields, allowed properties, or nested shapes to match the new intended payload.
5. Re-run the schema test to confirm the generated payload matches the approved fixture.

The fixture should not be regenerated automatically because it is acting as a review checkpoint. If this test fails, the developer should confirm whether the payload changed intentionally before updating the schema.

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: I updated this message to be more specific. The lock has an expiration, so this should not require immediate manual S3 intervention in the normal case. If the script fails while the lock is active, Lambda may be paused until the lock expires or the ops command is rerun successfully. I also removed the “reviewed” wording since that made the expected process unclear

refiner/scripts/reactivations/regenerate_active_configs.py
lina-roth: I think you’re right. The wait was meant as a precaution so Lambda executions that started before the lock was created would have some time to finish before regeneration begins.

But since we are not actually checking active Lambda executions, the delay does not guarantee that anything has drained. It could just add 30 seconds of waiting without improving safety, and long-running Lambdas could still overlap with regeneration anyway.

I’m going to remove the drain step rather than keep it as a configurable no-op.

jakewheeler (approved):
✅ :shipit:✅

PR #1489, opened 2026-07-06
feat: TES updates banner

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a "TES updates" banner to the Configurations overview page when a new version of the TES is available that the user has not yet seen/dismissed. Once the user dismisses/views the update the banner will no longer appear.

## 🔗 Related Issue

Fixes #1429

- #1429

## 📺  Example

After running `just db refresh` you'll see this
image

## 🧪 How to test

This functionality is based off of the pre-existing notifications framework in place, so we mainly just want to check that users will receive a TES notification when a new version is added to the database.

1. On this branch, run `just db refresh`
2. Navigate to the configurations overview page and you should see notifications for app updates and TES updates

At this point, I'd suggest running this SQL, which removes all data related to TES version `6.0.0`:
```sql
DELETE FROM conditions_context_groupers
WHERE condition_id IN (
    SELECT id FROM conditions WHERE tes_id = (SELECT id FROM tes WHERE version = '6.0.0')
);
DELETE FROM conditions WHERE tes_id = (SELECT id FROM tes WHERE version = '6.0.0');
DELETE FROM tes WHERE version = '6.0.0';
```

Run `just db seed`. Once done you should see the banner.


refiner/app/api/v1/tes.py
```diff
@@ -1,11 +1,11 @@
-from dataclasses import asdict, dataclass
+from dataclasses import dataclass
 from datetime import datetime
 from uuid import UUID
 
 from fastapi import APIRouter, Depends
 
 from app.db.pool import AsyncDatabaseConnection, get_db
-from app.db.tes_version.db import get_tes_updates_db
+from app.db.tes.db import get_loaded_tes_versions_db
```
jakewheeler, line 8: FYI @fzhao99 these are the slight changes I made upon merging in your work. Not much changed in the end but let me know if you have thoughts.

refiner/app/api/auth/handlers.py
```diff
@@ -261,13 +267,29 @@ def from_db_user(cls, db_user: DbUser) -> "UserResponse":
 
         should_show_app_update = latest_release_dt > app_update_ack_dt
 
+        latest_tes_version = get_latest_tes_version(
+            await get_loaded_tes_versions_db(db=db)
+        )
+        latest_tes_release_dt = _map_to_aware_dt(latest_tes_version.created_at)
```
jakewheeler, line 273: This seems to work as expected and allows us to stick with the same timestamp approach we're using for app updates.

refiner/app/db/tes/db.py
```diff
@@ -0,0 +1,24 @@
...
+    Returns an array off all loaded TES version records.
+    """
+    query = """
+    SELECT
+        id,
+        version,
+        created_at,
+        updated_at
+    FROM tes
+    ORDER BY version
```
jakewheeler, line 18: Worth noting that I'm just using `ORDER BY version` here to get the same return value every time we call the function. Since `version` is a string we can't rely on this being ordered in the way we'd expect (although it'd likely work fine unless we got an unusually formatted version number at some point in the future).

fzhao99 (approved):

PR #1493, opened 2026-07-07
feat: tes updates page

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Adds the TES updates page skeleton for the TES update flow

## 🔗 Related Issue
Fixes #1428 
- #1428

## ✅ Acceptance Criteria
- [ ] Adds TES updates page with:
   - [ ] Updates History Sidebar with versions and their effective dates & times (can be when added to Refiner or when published by TES, whichever is easiest)
   - [ ] Title and sub-title for each version
- [ ] "TES Updates" added to user menu

## 🧪 How to test
- Spin up the app
- Click into the TES updates page via the menu
- Notice the page renders

## ℹ️ Additional Information
Anything else we should add?


client/e2e/fixtures/index.ts
```diff
@@ -43,6 +47,12 @@ const extendedTest = baseTest.extend<Fixtures>({
   activityLogPage: async ({ page }, use) => {
     await use(new ActivityLogPage(page));
   },
+  tesUpdatesPage: async ({ page }, use) => {
+    await use(new TesUpdatesPage(page));
+  },
+  appUpdatesPage: async ({ page }, use) => {
```
fzhao99, line 53: went ahead and made a corresponding app updates POM fixture for completeness sake

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+          {tesUpdates.data.tes_updates.map((t) => {
+            return (
+              <Button
+                variant="unstyled"
+                key={t.id}
+                className={classNames(
+                  'px-6 py-2 hover:cursor-pointer',
+                  t.id === selectedUpdate?.id
+                    ? 'border-l-blue-cool-50 border-y-gray-cool-20! -mr-px! border-y border-l-8 bg-white'
+                    : 'text-blue-cool-60 px-6 py-2'
```
fzhao99: the text here is `blue-cool-60` as opposed to 50 in the designs bc of text contrast issues

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+import classNames from 'classnames';
+import { TesUpdate } from '../../api/schemas/tesUpdate';
+
+export function TesUpdates() {
+  const { data: tesUpdates, isPending, isError } = useGetTesUpdates();
+  const [selectedUpdate, setSelectedUpdate] = useState<TesUpdate>();
+
+  if (tesUpdates && selectedUpdate === undefined) {
+    setSelectedUpdate(tesUpdates.data.tes_updates[0]);
+  }
```
jakewheeler: I'd suggest tweaking the types and default value slightly just to be explicit that an "unselected" (or `null`) state is legit. The functionality works perfectly IMO and this won't change it.

```suggestion
  const [selectedUpdate, setSelectedUpdate] = useState<TesUpdate | null>(null);

  if (tesUpdates && !selectedUpdate) {
    setSelectedUpdate(tesUpdates.data.tes_updates[0]);
  }
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  if (isPending) return 'Loading...';
+  if (isError) return 'Error occurred!';
```
jakewheeler: If `isPending` is checked before the `setSelectedUpdate` there is no need to check that `tesUpdates` is truthy. 

```tsx
  if (isPending) return 'Loading...';
  if (isError) return 'Error occurred!';

  if (!selectedUpdate) {
    setSelectedUpdate(tesUpdates.data.tes_updates[0]);
  }
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export function TesUpdates() {
+  const { data: tesUpdates, isPending, isError } = useGetTesUpdates();
+  const [selectedUpdate, setSelectedUpdate] = useState<TesUpdate>();
+
+  if (tesUpdates && selectedUpdate === undefined) {
+    setSelectedUpdate(tesUpdates.data.tes_updates[0]);
+  }
+
+  if (isPending) return 'Loading...';
```
jakewheeler: Do we want to use our spinner here?
```suggestion
  if (isPending) return <Spinner variant="centered" />;
```

client/src/pages/TesUpdates/index.tsx
jakewheeler: Thanks for checking on this 👍

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+            return (
+              <Button
+                variant="unstyled"
+                key={t.id}
+                className={classNames(
+                  'px-6 py-2 hover:cursor-pointer',
+                  t.id === selectedUpdate?.id
+                    ? 'border-l-blue-cool-50 border-y-gray-cool-20! -mr-px! border-y border-l-8 bg-white'
+                    : 'text-blue-cool-60 px-6 py-2'
+                )}
```
jakewheeler: We can use a `classNames` object arg instead of the ternary here
```suggestion
                className={classNames('px-6 py-2 hover:cursor-pointer', {
                  'border-l-blue-cool-50 border-y-gray-cool-20! -mr-px! border-y border-l-8 bg-white':
                    t.id === selectedUpdate?.id,
                  'text-blue-cool-60 px-6 py-2': t.id !== selectedUpdate?.id,
                })}
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+          {tesUpdates.data.tes_updates.map((t) => {
+            return (
+              <Button
+                variant="unstyled"
+                key={t.id}
+                className={classNames(
+                  'px-6 py-2 hover:cursor-pointer',
+                  t.id === selectedUpdate?.id
+                    ? 'border-l-blue-cool-50 border-y-gray-cool-20! -mr-px! border-y border-l-8 bg-white'
```
jakewheeler: `-mr-px` is an interesting trick to get the effect we want.

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    month: '2-digit',
+    day: '2-digit',
+    year: '2-digit',
+    hour: '2-digit',
+    minute: '2-digit',
+    hour12: true,
+  };
+
+  return (
+    <div className="my-8 px-20">
```
jakewheeler: It might be nice to give more horizontal space on smaller displayers. This seems decent to me.
```suggestion
    <div className="my-8 md:px-20 px-2">
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Title className="mb-6!">TES Updates</Title>
```
jakewheeler: We can eliminate styling here if we make the container div a flex box too.

```tsx
    <div className="my-8 flex flex-col gap-6 px-20">
      <Title>TES Updates</Title>
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+      <div className="flex">
+        <div className="bg-blue-cool-5 border-gray-cool-20! flex h-full min-h-160 flex-col border">
```
jakewheeler: I see that `!` is needed for the border color but I am missing what is overriding it 🤔

jakewheeler: Do we want the button sidebar to scroll as the list of versions gets longer or do we want the entire panel to grow and we scroll the full page? Something like this

```suggestion
        <div className="bg-blue-cool-5 border-gray-cool-20! flex h-160 max-h-160 overflow-auto flex-col border">
```

jakewheeler: The overflow breaks the `-mr-px` trick 🙈 We'd need to mess with this a bit more beyond my suggestion if that's the direction we want to take.

refiner/app/api/v1/tes.py
```diff
@@ -0,0 +1,55 @@
...
+        db (AsyncDatabaseConnection): Database connection.
+
+    Returns:
+        TesResponse: A bundle with a list of TesUpdates, including
+            - The version
+            - The when it was created
+    """
+    updates = await get_tes_updates_db(db=db)
+
+    return TesResponse(tes_updates=[TesUpdate(**asdict(t)) for t in updates])
```
jakewheeler, line 59: It might be a good idea to do our own sorting here just to be sure the sorting will always be what we expect regardless of how the db function changes over time. I don't know that we have a version sorting function on hand but sorting by `created_at` in descending order would work I think.

refiner/app/db/tes_version/db.py
```diff
@@ -0,0 +1,43 @@
...
+    Returns:
+        list[DbTesUpdate]: a list of the TES information
+    """
+    query = """
+    SELECT
+        id,
+        created_at,
+        version
+    FROM tes
+    ORDER BY version DESC;
```
jakewheeler: This should work unless we get a TES version in an unexpected format at some point. As I mentioned in my other comment, probably best not to rely on this ordering being correct.

refiner/app/db/tes_version/db.py
```diff
@@ -0,0 +1,43 @@
+from dataclasses import dataclass
```
jakewheeler, line 1: I ended up needing to start a similar file in my PR. Once yours is merged I'll plan to rectify everything in my work. This looks good 👍

jakewheeler (changes requested):
Great work! Left some minor feedback and a couple questions but everything seems to be working well.

refiner/app/api/v1/tes.py
fzhao99: This should work everywhere but locally after a full DB refresh, since I think the DB script sticks in the versions `created_at` at the same timestamp. Think I'll sort by `created_at` first and then tie break with the version name

jakewheeler: Yeah, good plan. Locally is the one case I was thinking about as well.

client/src/pages/TesUpdates/index.tsx
fzhao99: I think this is a function of [`border` defaulting to the text color](https://github.com/tailwindlabs/tailwindcss/discussions/14508), which is where the black border is coming from without the `!`. Strangely enough it doesn't override when you supply the color 🤔 

I tried messing around with the default themeing etc. but can't seem to get it to work without the `!`, so I'll just leave it for a followup.

fzhao99: moved the border styling to the buttons themselves and opted to add a hidden spacer element that grows to the rest of the un-filled space that we're applying a border to (with an aria-hidden to indicate it's an aesthetic-only element). A bit of a hack, but think it accomplishes what we're after unless you see another issue.

client/src/pages/TesUpdates/index.tsx
```diff
@@ -56,6 +56,11 @@ export function TesUpdates() {
               </Button>
             );
           })}
+          {/* element here to allow border for the un-filled portion of the sidebar */}
+          <div
+            className="border-gray-cool-20! grow border-r"
+            aria-hidden="true"
```
fzhao99, line 62: @jakewheeler lmk if you think this will cause issues

jakewheeler: I believe this is fine!

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    month: '2-digit',
+    day: '2-digit',
+    year: '2-digit',
+    hour: '2-digit',
+    minute: '2-digit',
+    hour12: true,
+  };
+
+  return (
+    <div className="my-8 flex flex-col gap-6 px-20 md:px-20">
```
jakewheeler: I was wondering why this didn't seem to be working 😆 
```suggestion
    <div className="my-8 flex flex-col gap-6 px-2 md:px-20">
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <Title>TES Updates</Title>
+
+      <div className="flex">
+        <div className="bg-blue-cool-5 border-gray-cool-20! flex h-160 max-h-160 flex-col overflow-y-scroll border-y border-l">
```
jakewheeler: I'm noticing it's slightly jumpy when clicking from the latest version to older versions. Maybe we just throw a min width on here to alleviate that? This seems to do the job for me.

```suggestion
        <div className="bg-blue-cool-5 border-gray-cool-20! flex h-160 max-h-160 flex-col overflow-y-scroll border-y border-l min-w-30 md:min-w-52">
```

client/package.json
```diff
@@ -51,7 +51,7 @@
     "@testing-library/user-event": "14.6.1",
     "@types/node": "24.12.2",
     "@types/pg": "8.20.0",
-    "@types/react": "19.2.17",
+    "@types/react": "^19.2.17",
```
jakewheeler: Was this intentional?

refiner/app/api/v1/tes.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Returns:
+        TesResponse: A bundle with a list of TesUpdates, including
+            - The version
+            - The when it was created
+    """
+    updates = sorted(
+        await get_tes_updates_db(db=db),
+        key=lambda r: (r.created_at, r.version),
+        reverse=True,
```
jakewheeler, line 56: Nice 👍

jakewheeler (approved):
🎉 🫡

PR #1496, opened 2026-07-08
fix(ui,ux): Have search text for selecting configurations to conform to its stated minimum

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

I noticed this weird behavior where when searching for text such as HIV, it
wouldn't find a hit until I typed `HIV ` with a space at the end. This to me
felt like the 3 character minimum was actually more like 4. So this updates it
to be an inclusive comparison.

## 🔗 Related Issue

- n/a

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

1. `just dev::up`
1. Setup a new configuration
1. Type only three characters into the search to get results


jakewheeler (approved):
Thanks!

PR #1498, opened 2026-07-09
refactor(child_rsg_snomed_codes): Redirect data access to the normalized tables

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

> [!IMPORTANT]
> This work is split across three different PRs. This is the first one.

This is a patch that redirects the data access to the normalized tables to have
the application return codes from the `conditions_rsg_codes` join table and
leave the legacy `child_rsg_snomed_codes` column completely ignored. Search
queries also use the join table and not the array column. There are two more PRs
for this work as part of the same work.

## 🔗 Related Issue

- Closes #1465

## ✅ Acceptance Criteria

- [ ] ~~JSONB is removed from the conditions table~~
- [x] Backend / database interface code is refactored to use the new codes table / RSG join table
- [x] No application regressions are introduced

## 🧪 How to test

Since the goal of this patch is to confirm that the data access has been
redirected to the normalized tables, you'll need to do some manual SQL
verification and run a specific integration test.

### Run new integration test

1. Run `just server test tests/integration/test_rsg_normalization.py`

### Manual SQL verification

> [!NOTE]
> This should be run on a fresh DB. `just db::refresh`

1. Find an existing condition ID to test with:
   - `just db::query "SELECT id, display_name FROM conditions LIMIT 1;"`
   - The ID here is referenced as `<condition_id>` in the subsequent steps.
1. Insert a fake "legacy" code into the old `child_rsg_snomed_codes` array column:
   - `just db::query "UPDATE conditions SET child_rsg_snomed_codes = '{\"LEGACY\"}' WHERE id = '<condition_id>'`
1. Verify the fake "legacy" code was applied:
   - `just db::query "SELECT id, display_name, child_rsg_snomed_codes FROM conditions WHERE id = '<condition_id>';`
1. Insert the correct value into normalized tables
1. Get an existing system ID (for the foreign key constraint):
   - `just db::query "SELECT id FROM systems LIMIT 1;"`
   - The ID here is referenced as `<system_id>` in the subsequent steps.
1. Generate UUIDs for the new code record:
   - `just db::query "SELECT gen_random_uuid() AS code_id;"`
   - The ID here is referenced as `<code_id>` in the subsequent steps.
1. Insert the code into the `codes` table:
   - `just db::query "INSERT INTO codes (id, code, display, version, system_id) VALUES ('<code_id>', 'NORMAL_VAL', 'Normal Display', '1.0.0', '<system_id>');"`
1. Link the code to the condition via the join table:
   - `just db::query "INSERT INTO conditions_rsg_codes (condition_id, code_id) VALUES ('<condition_id>', '<code_id>');"`
1. Query the database directly to see what the app will return:
   - ```
     just db::query "
     SELECT
       c.id,
       c.display_name,
       c.child_rsg_snomed_codes AS legacy_column,
       ARRAY(
         SELECT codes.code
         FROM conditions_rsg_codes crc
         JOIN codes ON crc.code_id = codes.id
         WHERE crc.condition_id = c.id
       ) AS normalized_codes
     FROM conditions c
     WHERE c.id = '<condition_id>';
     "
     ```
   - Expected output:
     - `legacy_column`: `["LEGACY_VAL"]` <- ignored by the app
     - `normalized_codes`: `["NORMAL_VAL"]` <- what the app returns
1. Search by the normalized code:
   - ```
     just db::query local "
     SELECT c.id, c.display_name
     FROM conditions c
     WHERE EXISTS (
         SELECT 1
         FROM conditions_rsg_codes crc
         JOIN codes ON crc.code_id = codes.id
         WHERE crc.condition_id = c.id
         AND codes.code = 'NORMAL_VAL'
     );
     "
     ```
   - Expected output should be the condition returned.
1. Search by the legacy code:
   - ```
     just db::query local "
     SELECT c.id, c.display_name
     FROM conditions c
     WHERE EXISTS (
         SELECT 1
         FROM conditions_rsg_codes crc
         JOIN codes ON crc.code_id = codes.id
         WHERE crc.condition_id = c.id
         AND codes.code = 'LEGACY_VAL'
     );
     "
     ```
   - Expected output should be an empty set.

## ℹ️ Additional Information

> [!IMPORTANT]
> The manual SQL checking isn't necessary since we now have a new integration
> test, but I included it as a sanity check.


refiner/tests/integration/test_rsg_normalization.py
```diff
@@ -0,0 +1,93 @@
...
+import pytest
+
+from app.db.conditions.db import (
+    get_condition_by_id_db,
+    get_conditions_by_child_rsg_snomed_codes_db,
+)
+
+
+@pytest.mark.asyncio
+async def test_rsg_code_redirection(db_pool):
```
fzhao99, line 12: is the thought that we'll delete this test once the columns get cleaned up? Or will it still pass somehow after we get rid of the legacy column?

refiner/app/db/conditions/db.py
fzhao99, line 1: I think there are a couple of functions in this file that use the `SELECT cond.*` pattern in this file that we'll need to extend a read into the join table in a similar pattern in order to get the child SNOMED array, like [this guy](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/db/conditions/db.py#L348) and [this guy](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/app/db/conditions/db.py#L372)

refiner/tests/integration/test_rsg_normalization.py
rogeruiz: yeah I have it removed in #1503

refiner/app/db/conditions/db.py
rogeruiz: yeah. I have those working in #1501 maybe we should merge that into this one before merging to `main`?

fzhao99: yeah since they're in the same category of "read refactor", think it makes sense to process those as a hunk

fzhao99 (approved):

PR #1490, opened 2026-07-06
chore(server): update the narrative footnote's language around configuration details so they're aligned with the UI

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This keeps the language between the UI and the data that surfaces in the narrative footnotes consistent. When talking about configuration related mechanisms, it's best to keep the wording consistent and this helps protect against any future drift.

There may be a better option in the future but this helps us at least mirror the wording immediately ahead of getting a CAB approval for tomorrow.

## 🔗 Related Issue

Fixes #1468 

- #1468

PR #1499, opened 2026-07-09
fix: switch error message spacing

fzhao99:
Fixing a visual regression that got introduced in [1490](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1490/changes), replacing a flex-based layout of the sections page with the grid-based layout that keeps the switch "anchored" in the correct place like was intended in [1487](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1487)

Before:
https://github.com/user-attachments/assets/65d7a40d-7423-40a2-a5d4-6ad99ff6d42a

After:
https://github.com/user-attachments/assets/d73a4c52-e33a-42d7-889d-bb510b274e56


rogeruiz (approved):
lgtm 🌈

PR #1500, opened 2026-07-09
fix: Serialized content link is hidden when env is `live`

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR fixes a small issue with #1469 where the proper check was not in place for when the serialized content link will render 🙃  It should only render when `useGetEnv()` returns `local`.


robertmitchellv (approved):
```bash
echo $ENV, laugh, love
live, laugh, love
```

PR #1502, opened 2026-07-09
refactor(just): remove any CI checks

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

These commands aren't used in CI, so there's no real need to add them. If you want to use `-vvv` you can use `just server test ...` and otherwise all tests run without verbosity just for convenience and to not be too noisy.

## 🔗 Related Issue

- n/a

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

1. All server tests should still pass: `just server::test-all`


jakewheeler (approved):
🫡

PR #1503, opened 2026-07-09
refactor(db): Migrate away the `child_rsg_snomed_codes` JSONB

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

> [!IMPORTANT]
> This work is split across three different PRs. This is the third one.

This is the final patch of the RSG code normalization effort. It permanently
removes the legacy `child_rsg_snomed_codes` column and its GIN index from the
`conditions` table. The application now relies entirely on the normalized
`codes` and `conditions_rsg_codes` tables.

## 🔗 Related Issue

- Closes #1465
- Builds off of #1501

## ✅ Acceptance Criteria

- [x] JSONB is removed from the conditions table
- [x] ~~Backend / database interface code is refactored to use the new codes table / RSG join table~~
- [x] ~~No application regressions are introduced~~

## 🧪 How to test

1. `just migrate::local`
1. `just server::test-integration`


refiner/migrations/20260709201220_drop_legacy_rsg_codes_column.sql
```diff
@@ -0,0 +1,37 @@
+-- migrate:up
+-- 1. Final data migration (JSONB ➡️ Join Table) for any remaining records.
+-- This ensures that any data entered in the legacy column during the transition period
+-- is captured in the new normalized structure.
```
jakewheeler: My preference would be to lose the data rather than attempt to backfill it. I would prefer to keep `codes` "clean" by only allowing data to be inserted into it by `load_static_data.py`. Data making its way into that table by any other means seems unexpected to me.

What do you think? Am I understanding the intention here correctly?

refiner/migrations/20260709201220_drop_legacy_rsg_codes_column.sql
```diff
@@ -0,0 +1,37 @@
...
+DROP INDEX IF EXISTS idx_conditions_child_snomed_codes;
+
+-- 3. Drop the legacy column
+ALTER TABLE conditions DROP COLUMN IF EXISTS child_rsg_snomed_codes;
+
+-- migrate:down
+-- Since dropping a column is destructive, we cannot easily restore the data
+-- without a backup. We will leave the down migration empty or implement a
+-- basic column restoration if the schema allows.
+ALTER TABLE conditions ADD COLUMN child_rsg_snomed_codes JSONB;
```
jakewheeler:
```suggestion
ALTER TABLE conditions ADD COLUMN child_rsg_snomed_codes TEXT[];
```

jakewheeler (changes requested):
A question and column type fix but otherwise looks good 👍

refiner/migrations/20260709201220_drop_legacy_rsg_codes_column.sql
rogeruiz: ah that's true. okay i'll update this. thanks @jakewheeler

jakewheeler (approved):
Thanks Roger 👍

PR #1520, opened 2026-07-13
feat: Export activity log

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds a CSV export feature to the activity log page. Clicking the `Export as CSV` button will download a CSV containing the data from the page.

- CSV will be filtered to a specific condition if a `Condition` from the dropdown has been selected
- The `Date` column will attempt to use the user's local timezone and will fall back to UTC if it can't be determined
- Custom code upload events will include the action text and each code upload on their own line. Here is an example
    ```
    Added 6 custom codes from CSV
    SNOMED | 1115565 | SNOMED Example
    Other | 1431534 | Other Example
    ICD-10 | 6789 | ICD-10 Example
    LOINC | 99999A | LOINC Example
    CVX | 143 | CVX Example
    RxNorm | 198440 | RxNorm Example
    ```

👉 [Kyle approved](https://skylight-hq.slack.com/archives/C08QQ1SCDLY/p1784139633488779?thread_ts=1784126049.190359&cid=C08QQ1SCDLY) ✅ 

## 🔗 Related Issue

Fixes #1506 

- #1506

## 🧪 How to test

1. Log in and create a configuration
2. Perform actions that generate events (add/remove condition code sets, upload custom codes, add custom codes, modify sections, create custom sections, etc.)
3. Navigate to the activity log page
4. Use the `Export as CSV` button to download an export
5. Open the export and ensure it looks as expected

Try filtering by condition and then generating the export. It should show only the events related to that condition.

### Stress testing

I tested this with >10k event records. You can generate a bunch of records using a SQL query like this, in case you want to do any stress testing / file size testing:

```SQL
INSERT INTO events(jurisdiction_id, user_id, configuration_id, event_type, action_text, created_at)
SELECT
    'SDDH',
    '673da667-6f92-4a50-a40d-f44c5bc6a2d8', -- <-- put your user ID here
    '0e161c66-1ad7-4e81-8fdf-1056833cec78', -- <-- put your config ID here
    'add_code',
    'test ' || i,
    NOW() - (i || ' minutes')::interval
FROM generate_series(1, 10000) AS s(i); -- <-- modify the upper bound if you'd like to generate more records
```


client/src/components/Button/index.tsx
```diff
@@ -13,6 +13,8 @@ interface ButtonProps extends HeadlessButtonProps {
   variant?: ButtonVariant;
   to?: string;
   href?: string;
+  anchorProps?: React.AnchorHTMLAttributes<HTMLAnchorElement>;
+  linkProps?: Omit<LinkProps, 'to'>;
```
jakewheeler, line 17: This doesn't feel very elegant but also doesn't seem too unreasonable. If you have a suggestion about how to make this a little nicer I'd be interested in hearing it.

[Here](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1520/changes#diff-fe16633984d6d899a9e5d7ce7f5f55d8b513baeb4d26672eb348f51ef46b1ea0R112) is an example of how this is being used.

client/src/pages/ActivityLog/index.tsx
```diff
@@ -80,3 +90,28 @@ export function ActivityLog() {
...
+  // See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#timezone
+  const params = new URLSearchParams({
+    timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+  });
+
+  if (canonicalUrl) {
+    params.set('canonical_url', canonicalUrl);
+  }
+
+  return (
```
jakewheeler, line 109: I don't know if this is the place to solve it (if we think it's a problem) but we have multiple different "download file" link/button styles.

- This one is a link styled like a button that downloads directly
- We have the "export configuration" zip download link styled like a link
- We have the test file zip download button styled like a link that shows a download spinner

client/e2e/activity-log.spec.ts
```diff
@@ -139,4 +139,19 @@ test.describe('Activity log', () => {
     ).toBeVisible();
     await expect(makeAxeBuilder).toHaveNoAxeViolations();
   });
+
+  test('Export button downloads a CSV file', async ({
```
jakewheeler, line 143: This is really just here to check that the file downloads correctly from the UI. The actual content of the file is checked in the integration tests.

refiner/app/db/events/db.py
```diff
@@ -182,6 +199,54 @@ async def get_events_by_jd_db(
...
+        ORDER BY e.created_at DESC;
+    """
+    params = (canonical_url, canonical_url, jurisdiction_id, canonical_url)
+
+    async with db.get_connection() as conn:
+        async with conn.cursor(row_factory=class_row(CsvEvent)) as cur:
+            await cur.execute(query, params)
+            while chunk := await cur.fetchmany(500):
+                for row in chunk:
+                    yield row
```
fzhao99, line 247: I've been trying to figure out what the proper pattern is for execute / retrieve operations, since psycopg has a couple different API's for cursor execution and retrieve

Could you talk me through a bit why we're using this pattern here? Guessing it's because we're not quite sure what the size of the file is and don't want to load it all into memory at once.

Should our default row streaming be something like: 

```
cur.executemany(query, params, returning=True)

results = []
results.append(cur.fetchone()) 
if cur.nextset():
    results.append(cur.fetchone())
```

or is a different pattern better here?

fzhao99 (approved):
lgtm!

refiner/app/db/events/db.py
jakewheeler: > Guessing it's because we're not quite sure what the size of the file is and don't want to load it all into memory at once.

Yeah, this is exactly it. I can add a comment about this if it'd be helpful. I was being a little extra cautious here because this is a case where a jurisdiction could create a huge amount of events over time. Since it's pretty likely the only filter that will be used the vast majority of time is the jurisdiction ID (the only required filter) I figured batching the results like this would be a good way to go.

I mainly went with this because there isn't much extra complexity required to account for the possibility, so I figured I might as well do it.

For most circumstances I think what you're suggesting makes sense. Even in this situation something like that would _probably_ be fine in reality.

PR #1524, opened 2026-07-15
fix: virtualize content in serialized file view

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

I was experiencing performance issues when taking a look at the serialized view for a COVID-19 config that also had the Influenza code set attached. The `active.json` file is very large and the browser struggles to render it.

This PR virtualizes the content in the `CodeBlock` components in order to address this problem.

## 📺 Demo

Kapture 2026-07-15 at 11 26 32

## 🧪 How to test

1. Create a COVID-19 config
2. Attach the Influenza code set to it
3. Activate it
4. Return to the `Build` tab
5. Click on the `View serialized configuration (local only)` link in the top right corner of the page
6. The `active.json` file should be expanded by default. Try scrolling down and making sure all of the content loads in without performance hitches


fzhao99 (approved):
👍

PR #1527, opened 2026-07-16
fix: Update outdated/incorrect activity log display text

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the event action text related to section changes to use current language. We were previously continuing to use older verbiage in action text even though the wording in the UI has changed over time. The action text for connecting a code set to a condition now uses the word `Added` instead of `Associated`.

Please see the ticket for the exact wording changes being made here.

## 🔗 Related Issue

Fixes #1521 

- #1521 

## 📺 Demo

image

## 🧪 How to test

1. Log in, create a configuration, and make changes to the `Coded data` and `Narrative data` under the sections tab
2. Take a look at the activity log. You should see that event action text now uses the same wording that's found within the section tab

I've also added automated tests to ensure that all `DbSectionAction` and `DbNarrativeAction` will have an associated label going forward.


refiner/tests/unit/test_configurations.py
```diff
@@ -517,3 +524,28 @@ async def mock_s3_upload(*args, **kwargs):
     )
 
     test_app.dependency_overrides.clear()
+
+
+def test_all_coded_data_actions_have_labels():
```
jakewheeler, line 529: Try adding a new value to `type DbSectionAction` (in `db/configurations/model.py`) without adding a label (in `db/configurations/labels.py`) if you want to get a sense of how this these tests work. The opposite (adding a new label to a non-existent value) will also cause a failure.

fzhao99 (approved):

PR #1528, opened 2026-07-16
refactor: custom codes are stored in `custom_codes` table

jakewheeler:
# 🔀 PULL REQUEST

> [!NOTE]
> This PR is on the larger side but it allows you to test 100% of the custom codes functionality and guarantee everything works once this PR is merged (this is especially important to check following the manual testing steps below). 
>
> The actual code changes are not super interesting since much of it is just small changes to handle new property names, going from `system_key` to `id`, and updating tests. I've gone through and mentioned what I intend to refactor in a follow-up PR, so hopefully that will help to limit time spent reviewing.

## 💡 Summary

This PR:

- Adds a migration script to move custom codes data from the configuration record to the `custom_codes` table
- Modifies all application code required to make use of the `custom_codes` table
- Modifies all tests required to handle the app changes

This PR does not introduce or remove any features.

## 🔗 Related Issue

Fixes #1424 

- #1424 

## 🧪 How to test

All of the automated tests should pass as expected. I did need to make updates to these tests due to implementation changes. Please take a look at those and make sure nothing sticks out!

### Manual testing

1. Checkout `main`, run `just db refresh`
2. Create a configuration (or multiple) and add some custom codes, either manually or via the upload functionality
3. Checkout this branch, run `just migrate local`
4. The `custom_codes` table should now be populated and all rows will tie back to their original configuration's ID
5. All custom codes functionality should continue to work as before
6. Try `just migrate local down` when you're finished manually testing if you'd like to ensure the custom codes revert back to their old state correctly

Some custom code related features to check:

- Adding / editing / deleting one at a time
  - Check that events created look accurate 
- Adding codes using the CSV upload
  - Edit preview rows
  - Delete previews
  - Undo and delete all rows
  - Create the rows
- Activating the config
  - Use the "serialized config" view to see that custom codes end up in the serialized configuration correctly
  - Create a new draft to see that cloning custom codes to the new config works

## ℹ️ Additional Information

There is more clean up to be done but I will continue that work in a follow up PR.


client/e2e/db/index.ts
```diff
@@ -13,10 +13,20 @@ export const db = new Pool({
...
 
+export async function deleteAllCustomCodes(): Promise<void> {
+  await db.query('DELETE FROM custom_codes');
+}
+
 export async function deleteAllConfigurations(): Promise<void> {
   await db.query('DELETE FROM configurations');
 }
 
+export async function clearDb(): Promise<void> {
```
jakewheeler, line 24: Custom codes now rely on configurations due to being linked. The integration test cleanup step is very similar to this.

client/src/api/code-systems/code-systems.ts
```diff
@@ -131,9 +131,3 @@ export function useGetCodeSystems<TData = Awaited<ReturnType<typeof getCodeSyste
 
   return withQueryKey(query, queryOptions.queryKey);
 }
-
```
jakewheeler, line 134: I'm not sure why these files are being changed. I assume one of our tools (pre-commit?) is modifying these when it's not supposed to.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -52,12 +56,12 @@ def _validate_add_custom_code_input(input: AddCustomCodeInput):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail='Required field "code" is missing.',
         )
-    if not input.system_key:
+    if not input.system_id:
```
jakewheeler, line 59: (Note to myself for follow up) not sure we need this validation since FastAPI should handle checking input.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -152,11 +157,24 @@ async def add_custom_code(
...
-        custom_codes=updated_config.custom_codes,
+        custom_codes=[
+            CustomCodeResponse(
+                id=cc.id,
+                display=cc.display,
+                code=cc.code,
+                system_id=cc.system_id,
+                system_name=get_code_system_by_id_or_raise(
+                    id=cc.system_id, systems=systems
+                ).display_name,
```
jakewheeler, line 174: (Note to myself for follow up) We have multiple ways to get a code system by ID. Evaluate what makes sense to keep.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -218,64 +236,52 @@ async def _get_requested_config_or_raise(
...
     name = (row.get("display_name") or "").strip()
 
-    row_errors = []
-    row_system = None
+    system_names = [s.display_name for s in supported_systems]
+
+    # get the DbCodeSystem that matches CSV system
+    matching_system = next(
+        (s for s in supported_systems if s.display_name == code_system_raw), None
+    )
```
jakewheeler, line 251: (Note to myself for follow up) could make sense to have a service function handle this.

refiner/app/api/v1/configurations/custom_codes.py
```diff
@@ -666,10 +593,10 @@ def _validate_edit_custom_code_input(input: UpdateCustomCodeInput):
             status_code=status.HTTP_400_BAD_REQUEST,
             detail='Required field "code" is missing.',
         )
-    if not input.system_key:
+    if not input.system_id:
```
jakewheeler, line 596: (Note to myself for follow up) worth rechecking all validation for all handlers.

refiner/app/db/configurations/db.py
```diff
@@ -351,6 +352,19 @@ async def insert_configuration_db(
...
+                    await cur.executemany(
+                        """
+                        INSERT INTO custom_codes (configuration_id, code, display, system_id)
+                        VALUES (%s, %s, %s, %s)
+                        """,
+                        [
+                            (config_id, cc.code, cc.display, cc.system_id)
+                            for cc in config_to_clone.custom_codes
+                        ],
+                    )
```
jakewheeler, line 359: (Note to myself for follow up) See if we can clean this function up.

refiner/app/db/configurations/db.py
```diff
@@ -746,49 +747,40 @@ async def add_bulk_custom_codes_to_configuration_db(
...
-        SET custom_codes = %s::jsonb
-        WHERE id = %s
-        RETURNING
-            id;
+    placeholders = ", ".join(["(%s, %s, %s, %s)"] * len(custom_codes))
+    query = f"""
+        INSERT INTO custom_codes (configuration_id, display, code, system_id)
+        VALUES {placeholders}
+        ON CONFLICT DO NOTHING
+        RETURNING *;
```
jakewheeler, line 747: I tried doing this a few different ways and this way allowed me to get the return value, which I needed to use for the event insert.

refiner/app/db/configurations/db.py
```diff
@@ -746,49 +747,40 @@ async def add_bulk_custom_codes_to_configuration_db(
...
-                return None
+            new_codes_added = await cur.fetchall()
 
             # Insert a single audit event if codes were added
             await insert_custom_code_upload_events_db(
                 configuration=config,
                 user_id=user_id,
-                custom_codes=new_codes_added,
+                custom_codes=[
+                    # TODO: add `from_db_row`?
```
jakewheeler, line 764: We do this twice (here and in `configurations/model.py`) so maybe worth it. I'll keep the TODO for now and either add the function or delete the TODO in my next PR.

refiner/app/db/configurations/db.py
```diff
@@ -859,79 +842,85 @@ async def delete_custom_code_from_configuration_db(
...
             # 3. System changed
-            if new_system_key is not None and new_system_key != prev_system_key:
+            if system.id != custom_code.system_id:
+                prev_system = await get_code_system_by_id_db(
+                    id=custom_code.system_id, db=db
+                )
+                if prev_system is None:
+                    raise ValueError(
+                        f"Could not find code system with ID {custom_code.system_id}"
+                    )
```
jakewheeler, line 907: This both satisfies mypy and is a good "just in case" check. It's not likely we hit this since the ID we're looking for has already been inserted previously.

refiner/app/db/configurations/model.py
```diff
@@ -39,22 +41,6 @@ class DbTotalConditionCodeCount:
     total_codes: int
 
 
-# TODO: Revisit this to see if we can figure out how to reduce overlap with other types.
-# This is a "custom_code" column on the configuration row in the configurations table.
-# This is one object in the `custom_codes` list and we have many objects that are Similar
-# shaped. A flexible FHIR R4 Coding model with standard metadata could simplify a lot of
-# redundancy.
-@dataclass(frozen=True)
-class DbConfigurationCustomCode:
```
jakewheeler, line 48: We now only use `DbCustomCode` which represents a custom code row in the DB.

refiner/app/db/events/db.py
```diff
@@ -275,14 +276,20 @@ async def get_custom_code_upload_events_by_event_id(
...
     """
     Helper function to insert a bulk custom code upload event and its subevents.
     """
 
+    def _get_system_name(id: UUID) -> str:
+        system = next((s for s in code_systems if s.id == id), None)
+        if system is None:
+            raise ValueError(f"Unable to determine code system by ID: {id}")
+        return system.display_name
+
```
jakewheeler, line 292: (Note to myself for follow up) this is another situation where it could be nice to have this as a service function.

refiner/migrations/20260716184236_populate_custom_codes_table.sql
```diff
@@ -0,0 +1,61 @@
...
+
+-- Add configuration_id to custom_codes table
+ALTER TABLE custom_codes
+    ADD COLUMN configuration_id UUID REFERENCES configurations(id);
+
+-- Replace the unique constraint to scope it per configuration
+ALTER TABLE custom_codes
+    DROP CONSTRAINT IF EXISTS custom_codes_system_id_value_key,
+    ADD CONSTRAINT custom_codes_configuration_id_system_id_code_key
+        UNIQUE (configuration_id, system_id, code);
```
jakewheeler, line 11: No duplicates rule at the DB level 🙏

refiner/tests/integration/scenarios/conftest.py
```diff
@@ -424,11 +427,22 @@ async def _build(scenario) -> tuple[ProcessedConfiguration, str]:
...
+        systems = await get_systems()
+
+        def _get_system_id(system_key: str) -> UUID:
+            matching_system = next(
+                (s for s in systems if s["key"] == system_key),
+                None,
+            )
+            if matching_system is None:
+                raise ValueError(f"Could not find system matching key: {system_key}")
+            return matching_system["id"]
```
jakewheeler, line 466: (Note to myself for follow up) very similar to the helper in the events file.

refiner/tests/utils/scenario_builder.py
```diff
@@ -102,11 +103,28 @@ async def build_and_activate(self, scenario: Scenario, jurisdiction_id: str):
...
+        systems = resp.json()
+
+        def _get_system_id(system_key: str) -> UUID:
+            matching_system = next(
+                (s for s in systems if s["key"] == system_key),
+                None,
+            )
+            if matching_system is None:
+                raise ValueError(f"Could not find system matching key: {system_key}")
+            return matching_system["id"]
```
jakewheeler, line 117: (Note to myself for follow up) Another instance of the same helper 🙃

client/src/pages/Configurations/ConfigBuild/CustomCodes/index.tsx
```diff
@@ -4,61 +4,32 @@ import {
...
   configurationId: string;
-  customCodes: DbConfigurationCustomCode[];
-  codeSystems: { [key: string]: DbCodeSystem };
+  customCodes: CustomCodeResponse[];
   disabled: boolean;
   isOpen: boolean;
   setIsOpen: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
-type ConfigurationCustomCodeDisplay = DbConfigurationCustomCode & {
```
jakewheeler, line 24: `CustomCodeResponse` includes the data this was attaching.

client/e2e/fixtures/api.ts
```diff
@@ -18,9 +18,15 @@ interface Condition {
...
 interface CustomCode {
   code: string;
-  system_key: string;
-  system_display_name: string;
-  name: string;
+  system_id: string;
+  display: string;
+}
+
+interface System {
```
fzhao99: I seem to be able to replace these redefinitions with Orval inputs as long as I prefix them with `type`:

`import type { CodeSystemsReponse } from '../../src/api/schemas/codeSystemsReponse';`

Any reason to not use the codegen output and prefer a redefinition here?

client/src/api/code-systems/code-systems.ts
fzhao99: i've noticed these too. think there may be a dueling formatter issue somewhere.

will try to keep an eye out in case it shows up in another PR

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
```diff
@@ -137,12 +137,10 @@ function CustomCodeForm({
...
-            code: selectedCustomCode.code,
-            system_key: selectedCustomCode.system_key,
-            name: selectedCustomCode.name,
-            new_code: code.trim(),
-            new_system_key: selectedSystem,
-            new_name: name.trim(),
+            id: selectedCustomCode.id,
+            code: code.trim() || selectedCustomCode.code,
+            system_id: selectedSystemId || selectedCustomCode.system_id,
+            display: name.trim() || selectedCustomCode.display,
```
fzhao99, line 143: glad we could simplify the `new_` attributes away

client/src/pages/Configurations/ConfigBuild/index.test.tsx
```diff
@@ -751,23 +753,7 @@ const baseMockConfig: GetConfigurationResponse = {
...
-      },
-      snomed: {
-        key: 'snomed',
-        id: MOCK_SNOMED_DB_ID,
-        display_name: 'SNOMED',
-        oid: '2.16.840.1.113883.6.96',
-      },
-    },
-  },
+  custom_codes: mockCustomCodes,
```
fzhao99, line 756: 🎉

refiner/migrations/20260716184236_populate_custom_codes_table.sql
fzhao99: Do we foresee any issues for baking the `custom_codes` <> `configurations` into the base custom codes table rather than farming it out into a join like we do for `codes` <> `conditions` --> `conditions_codes`? Don't think it's a current issue, but we are mixing the two patterns together across the join types.

Could see the argument for treating them differently bc we're assuming custom codes always belong to a condition, and that we don't need to dedupe within custom codes, but raising in case it flags anything for you

fzhao99 (commented):
a few questions but otherwise functionality is looking good! great that we can simplify a lot of the complexity of juggling types down :)

refiner/migrations/20260716184236_populate_custom_codes_table.sql
jakewheeler: It's a good point. I've had a similar thought about the way sections are handled as well. I agree it's not a current issue for the reason you've mentioned.

client/src/pages/Configurations/ConfigBuild/CustomCodes/CustomCodeModal.tsx
jakewheeler: Same!

client/e2e/fixtures/api.ts
jakewheeler: Oh nice 👀  I can't think of a reason not to use it. I think I forgot or didn't realize we could import from this location within the `e2e` package.

fzhao99 (approved):

PR #1529, opened 2026-07-17
feat (backend): extend seed script to include all codes

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Updates our local seed script to ingest all the TES codes.

locally, getting ~60 seconds for a full refresh and ~25 seconds for a rerun afterwards. I've set up the script to only load the last two env variables if `ENV=local` (reducing the refresh ~25 seconds), which you can check by running `just db refresh` with updated containers or by running the script with a locally set .env file.
## 🔗 Related Issue

Fixes #1576
- #1576

## ✅ Acceptance Criteria
  - [x] TES gets seeded with all values
  - [x] Some validation is provided that the seeded numbers are correct
  - [x] Optional: script affords the ability to load only the previous two TES versions on local-only runs

## 🧪 How to test

- Switch to main and run a full `just db refresh`
- Run these below SQL queries to get a count of both the unique codes and the total number of code assignments to the conditions

To count the distinct codes, we need to unroll the JSONB columns from the conditions table, join it with the `conditions_rsg_codes` table, and then group things by TES ID (version). Then, we want the distinct count across all TES ID's. Verify that we're getting these numbers below.

Distinct codes: 379,745 
```sql
SELECT SUM(unique_code_count) AS total
FROM (
    SELECT 
        tes_id,
        COUNT(DISTINCT (col_name, code)) AS unique_code_count
    FROM (
       -- 1. Unnest JSONB array columns directly from conditions
        SELECT 
            c.tes_id,
            col_list.col_name,
            elem ->> 'code' AS code
        FROM conditions c,
        LATERAL (
            VALUES 
                ('snomed', c.snomed_codes),
                ('loinc',  c.loinc_codes),
                ('rxnorm', c.rxnorm_codes),
                ('icd10',  c.icd10_codes),
                ('cvx',    c.cvx_codes)
        ) AS col_list(col_name, json_col),
        LATERAL jsonb_array_elements(COALESCE(col_list.json_col, '[]'::jsonb)) AS elem

        UNION ALL

        -- 2. Join and tag joined codes as 'snomed'
        SELECT 
            c.tes_id,
            'snomed' AS col_name,
            cd.code
        FROM conditions_rsg_codes cc
        JOIN codes cd ON cd.id = cc.code_id
        JOIN conditions c ON c.id = cc.condition_id
        WHERE cd.code IS NOT NULL
    ) combined_codes
    GROUP BY tes_id
) AS subquery;
```
To count the code releationships, we need to unroll the JSONB columns from the conditions table, join it with the `conditions_rsg_codes` table, and then group things by condition ID. Then, we want the distinct count across all conditions. Verify that we're getting these numbers below.

Distinct code relationships: 1292377
```sql
SELECT SUM(relationship_count) AS total
FROM (
    SELECT 
        id,
		COUNT(DISTINCT(col_name, code)) AS relationship_count
   FROM (
        -- 1. Unnest JSONB array columns directly from conditions
        SELECT 
            c.id,
            col_list.col_name,
            elem ->> 'code' AS code
        FROM conditions c,
        LATERAL (
            VALUES 
                ('snomed', c.snomed_codes),
                ('loinc',  c.loinc_codes),
                ('rxnorm', c.rxnorm_codes),
                ('icd10',  c.icd10_codes),
                ('cvx',    c.cvx_codes)
        ) AS col_list(col_name, json_col),
        LATERAL jsonb_array_elements(COALESCE(col_list.json_col, '[]'::jsonb)) AS elem
UNION ALL

        -- 2. Join and tag joined codes as 'snomed'
        SELECT 
            c.id,
            'snomed' AS col_name,
            cd.code
        FROM conditions_rsg_codes cc
        JOIN codes cd ON cd.id = cc.code_id
        JOIN conditions c ON c.id = cc.condition_id
        WHERE cd.code IS NOT NULL
    ) combined_codes    
    GROUP BY id
) AS subquery;
```
- Switch to this branch and run a local migration / seed with [this line commented out](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1529/changes#diff-2286cdeca85adb587a9640aa34352ba7c2fb6e49c0129dab05892ff67d0d4c58R57). Verify the app works correctly.

- Run a full refresh using `just db refresh`. Take note of the outputs in the logs
- For the `child_snomed_rsg` codes, we should see 1614 codes as was validated in [this PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1293). 
     - Verify this by running 
      ```
      SELECT * FROM conditions_codes WHERE is_child_rsg = True
      ```
- For the distinct codes, you should see 379,745 (size of the `codes` table) and 1,292,377 codes for the relationship join (size of the `conditions_codes` table). You can verify this by doing a bulk select from those tables.
 
## ℹ️ Additional Information
- Followups for 1) only seeding for new TES versions and 2) removing the "version" column from the codes table will be handled in a future PR.


refiner/migrations/20260728212408_update_conditions_join.sql
```diff
@@ -0,0 +1,9 @@
+-- migrate:up
+ALTER TABLE conditions_rsg_codes RENAME TO conditions_codes;
+ALTER TABLE conditions_codes ADD COLUMN is_child_rsg BOOLEAN DEFAULT FALSE;
```
fzhao99, line 3: I opted to change the conditions <> codes table to house all the joins with a column rather than making a separate table. Think it's conceptually simpler this way and allows us to do one round-trip in the seeding script

refiner/scripts/seeding/load_static_data.py
```diff
@@ -353,109 +446,185 @@ def _upsert_conditions_and_groupers(
...
+    total_child_rsg_staged = 0
+    total_non_child_rsg_staged = 0
+
+    def code_generator():
+        nonlocal total_child_rsg_staged
+        nonlocal total_non_child_rsg_staged
+        for (url, cond_ver), trace in data.items():
+            for r in trace.get("child_rsg_codes") or []:
+                total_child_rsg_staged += 1
+                yield (
```
fzhao99: there were some minor performance improvements to explicitly defining these yields versus factoring them together using something like `yield from`, so leaving them separate for now.

refiner/scripts/seeding/load_static_data.py
```diff
@@ -353,109 +446,184 @@ def _upsert_conditions_and_groupers(
...
+            code_version TEXT NOT NULL,
+            is_child_rsg BOOLEAN NOT NULL
+        ) ON COMMIT DROP
+    """)
+    cursor.execute("TRUNCATE stage_codes")
+
+    total_child_rsg_staged = 0
+    total_non_child_rsg_staged = 0
+
+    def code_generator():
```
jakewheeler, line 528: From my testing so far it seems like handling de-duping here would be a big performance gain due to limiting how much data the SQL steps need to handle. `DISTINCT ON` is costly but if we can limit the data that should help a good bit.

Not 100% sure if correct but if I de-dupe here I see `Staged 378,353 total non_child_rsg rows, and 1,392 rows.`

This takes my `just db refresh` from ~62 seconds  to ~36 seconds.

fzhao99: Think was working on this for so long this obvious improvement didn't occur to me even when staring me in the face 😅

Moved this up higher in the build method since there's a bit of nuance on how we do the de-duping across code types. Think I'm getting a similar improvement controlling for computer clock speed

refiner/migrations/20260713183905_update_condition_codes_table.sql
```diff
@@ -0,0 +1,9 @@
+-- migrate:up
+ALTER TABLE conditions_rsg_codes RENAME TO conditions_codes;
+ALTER TABLE conditions_codes ADD COLUMN is_child_rsg BOOLEAN DEFAULT FALSE;
+CREATE UNIQUE INDEX IF NOT EXISTS codes_upsert_constraint_idx 
+    ON codes (system_id, version, code);
+
+-- migrate:down
+ALTER TABLE condition_codes DROP COLUMN is_child_rsg;
```
jakewheeler:
```suggestion
ALTER TABLE conditions_codes DROP COLUMN is_child_rsg;
```

refiner/migrations/20260728212408_update_conditions_join.sql
```diff
@@ -0,0 +1,9 @@
...
+CREATE UNIQUE INDEX IF NOT EXISTS codes_upsert_constraint_idx 
+    ON codes (system_id, version, code);
+
+-- migrate:down
```
jakewheeler, line 7: We need to drop the index as well.
```suggestion
-- migrate:down
DROP INDEX IF EXISTS codes_upsert_constraint_idx;
```

refiner/scripts/seeding/load_static_data.py
```diff
@@ -353,109 +440,128 @@ def _upsert_conditions_and_groupers(
...
+                yield (cond_id, system_id, version, code, False)
+
+    logger.info("🚀 Streaming relationships into stage table...")
+    with cursor.copy(
+        """COPY stage_relationships (condition_id, system_id, version, code, is_child_rsg) FROM STDIN"""
+    ) as copy:
+        for row in relationship_generator():
+            copy.write_row(row)
+
+    cursor.execute("ANALYZE stage_relationships;")
```
jakewheeler, line 484: What is this supposed to be doing? Or is it just left over from debugging?

jakewheeler: Oh, I see it used below as well so I assume it's intentional 👀

refiner/scripts/seeding/load_static_data.py
```diff
@@ -353,109 +440,128 @@ def _upsert_conditions_and_groupers(
...
+    with cursor.copy(
+        "COPY stage_codes (id, system_id, version, code, display) FROM STDIN"
+    ) as copy:
+        for record in code_generator():
+            copy.write_row(record)
+
+    logger.info(f"📥 Staged {len(data['codes_to_insert'])} unique code rows.")
+    cursor.execute("ANALYZE stage_codes;")
+    # bump up local memory to help with the large joins
+    cursor.execute("SET LOCAL work_mem = '128MB'")
```
jakewheeler: FWIW I am not seeing any noticeable performance gain compared to not having it. Are you?

refiner/scripts/seeding/load_static_data.py
```diff
@@ -142,37 +158,107 @@ def _upsert_tes_data(
...
+        cond_canonical_url = condition.get("url", "")
+        cond_version = condition.get("version", "")
         cond_index = (cond_canonical_url, cond_version)
 
+        condition_child_rsg_snomed_codes: set[SystemCodeTuple] = set()
+        condition_non_child_rsg_snomed_codes: set[SystemCodeTuple] = set()
+
+        child_tuples: set[FhirCodeTuple] = set()
+
         for child_vs in get_child_rsg_valuesets(
```
jakewheeler, line 182: Tried to simplify this loop a bit. Double check me if you decide to take it but my numbers seem to be coming out correctly. I noticed `code_index` and `system_code_tuple` were the same and reworked that a bit too.

```python
        for child_vs in get_child_rsg_valuesets(
            parent=condition, all_vs_map=valuesets_map
        ):
            child_rsg_code = parse_snomed_from_url(child_vs.get("url", ""))
            child_tuples.update(set(extract_codes_from_compose(child_vs)))

            if not child_rsg_code:
                continue

            display = parse_child_rsg_details_from_use_context(
                child_vs.get("useContext", "")
            )
            code_tuple = (snomed_db_id, cond_version, child_rsg_code)

            if code_tuple not in codes_seen_so_far:
                codes_seen_so_far.add(code_tuple)
                codes_for_codes_table.append(
                    CodeRow(
                        id=uuid4(),
                        code=child_rsg_code,
                        version=cond_version,
                        display=display,
                        system_id=snomed_db_id,
                    )
                )

            condition_child_rsg_snomed_codes.add(code_tuple)
```

refiner/scripts/seeding/load_static_data.py
```diff
@@ -142,37 +158,107 @@ def _upsert_tes_data(
...
+            system_id = oid_indexed_system_db_ids.get(system_oid)
+            if not system_id:
+                continue
+
+            for c in code_list:
+                code = c.get("code")
+                if not code:
+                    continue
+
+                code_index = (system_id, cond_version, code)
```
jakewheeler: Same thing with `code_index` and `system_code_tuple` here. Maybe this could be cleaned up similarly.

refiner/scripts/seeding/load_static_data.py
fzhao99: yeah in my research for using stage tables, [it was recommended to add this for large, newly created tables to give the Postgres engine to make followup queries more efficient.](https://www.postgresql.org/docs/current/sql-analyze.html)

refiner/scripts/seeding/load_static_data.py
fzhao99: think this was something I was toying around with but agreed it's not impacting things too much. removed!

refiner/tests/integration/test_simulator.py
```diff
@@ -239,18 +240,18 @@ async def test_discovers_configs_across_all_tes_versions(
     current_covid_id = await get_condition_id("COVID-19")
     current_flu_id = await get_condition_id("Influenza")
 
-    old_covid_id = await get_condition_id("COVID-19", "3.0.0")
-    old_flu_id = await get_condition_id("Influenza", "4.0.0")
+    old_covid_id = await get_condition_id("COVID-19", PREV_TES_VERSION)
```
fzhao99: lmk if just checking the one previous version defeats the purpose of this test

docker-compose.ci.yml
```diff
@@ -12,6 +12,9 @@ services:
         - VITE_GIT_HASH=${VITE_GIT_HASH}
         - VITE_GIT_BRANCH=${VITE_GIT_BRANCH}
         - VERSION=${VERSION}
+    depends_on:
+      migrate:
+        condition: service_completed_successfully
```
fzhao99, line 17: lmk if this has any unintended consequences for the seeding

refiner/tests/integration/conftest.py
```diff
@@ -69,6 +69,7 @@
 TEST_JD_STATE_CODE = "GC"
 
 DEFAULT_TES_VERSION = "6.0.0"
+PREV_TES_VERSION = "5.0.0"
```
jakewheeler, line 70: Minor change but we could make these fixtures like the rest of the constants we set in here.

```suggestion
DEFAULT_TES_VERSION = "6.0.0"
PREV_TES_VERSION = "5.0.0"

@pytest.fixture
def default_tes_version():
    return DEFAULT_TES_VERSION

@pytest.fixture
def previous_tes_version():
    return PREV_TES_VERSION
```

We could remove the `conftest` import in the other files and the tests could use the values from these fixtures instead.

refiner/tests/integration/test_simulator.py
jakewheeler: I think this should be alright!

docker-compose.ci.yml
```diff
@@ -12,6 +12,9 @@ services:
         - VITE_GIT_HASH=${VITE_GIT_HASH}
         - VITE_GIT_BRANCH=${VITE_GIT_BRANCH}
         - VERSION=${VERSION}
+    depends_on:
+      migrate:
```
jakewheeler, line 16: I'm running this locally to test the CI command: `docker compose -f docker-compose.yml -f docker-compose.ci.yml up -d`.

It seems like this is loading all TES versions still. Is that intended or no?
```sh
INFO:config:📝 Loading TES file 30 / 30: additional_context_grouper_4.0.0.part02.json
INFO:config:📊 Loaded 3159 unique ValueSets from all TES files.
INFO:config:🔎 Identified 1360 condition groupers to process.
INFO:config:⏳ Upserting TES records...
INFO:config:🛠️  Total TES rows upserted: 6
INFO:config:🛠️  Total condition rows processed: 1360
INFO:config:⬆️  Total conditions to upsert: 1360
INFO:config:⏳ Upserting condition records...
INFO:config:⏳ Starting codes upsert process...
INFO:config:🚀 Streaming codes into stage table...
INFO:config:📥 Staged 379745 unique code rows.
INFO:config:✨ 379,745 total new rows inserted in codes table.
INFO:config:⏳ Refreshing relationships table...
INFO:config:🚀 Streaming relationships into stage table...
INFO:config:🔗 Linking codes table to relationship joins...
INFO:config:📥 Inserted 1,292,377 total relationships (1,614 child_rsg, 1,290,763 non_child_rsg).
INFO:config:🏁 Done!
INFO:config:⏱️  Static data loaded in 66.490 seconds
```

fzhao99: I set things up that way assuming we want CI to mirror the data for a live env, which would have all the data from all the versions, which is why I left out the filtering / reserved it only locally. Feels like the right balance of trying to speed up local dev while making sure the test envs mirror live ones, but that may just be me.

Do you think we can get by with just two previous versions?

jakewheeler: Since we aren't explicitly testing versions beyond "current' and "previous" anymore I was thinking maybe it would make sense to just go ahead and get the speed benefits in CI as well.

Otherwise we're kind of just checking that the ops container is able to run against a fresh DB using all available TES versions. I wouldn't say that's not useful but I'm unsure if it's worth the decently large hit we take in the total pipeline runtime.

What do you think? I could go either way on it honestly. No problem if we'd prefer to leave it as is for now.

fzhao99: I think I still lean towards leaving all the data in CI in the event there's an issue that we'd catch with a full DB that we wouldn't catch without it. Think it's a relatively simple swap in the event that the timing hit is too much

jakewheeler (approved):
Great job!

PR #1551, opened 2026-07-22
fix: remove original and augmented doc_id from lambda logs

lina-roth:
# 🔀 PULL REQUEST

## 💡 Summary

In order to remove any potential PII or PHI risks from lambda logs, removing original and augmented doc id from being written (was only happening when remainder RR was being written). Change will be mentioned to APHL in Tech Sync on 7/23

## 🔗 Related Issue

Fixes #1549 

- #1549 

## ✅ Acceptance Criteria

 - [ ] rr_doc_id is removed from the logs in the augmentation module
- [ ]  Change is communicated to APHL


jakewheeler (approved):
🫡

PR #1552, opened 2026-07-22
feat(rfc): add RFC for removing default code sets in configurations

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

This patch adds the findings from CDCGov/dibbs-ecr-refiner#1525 and the
prototype that was created to support that spike. After presenting things to the
team, we came up with a second option of toggling custom codes as the single
source of truth for code sets in a configuration. This is a different approach
that deserves to be documented as another option.

The main reason for this RFC is that this work is essentially paused while we
figure out from product decisions whether this work is viable for our team to do
currently or not.

## 🔗 Related Issue

- Related to #1525
- Discovered in #1548 

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

1. Read through the new RFC

### DIFF
--- a/docs/decisions/0013_2026-07-22_zero-code-set-refinement.md
+++ b/docs/decisions/0013_2026-07-22_zero-code-set-refinement.md
@@ -53,6 +53,9 @@ configuration identity from mapping logic.
 **Cons**: Requires schema/API changes and more extensive testing of constraints.
 Need to update `seed_db.py` to ensure deletions persist across redeploys.
 
+> [!TIP]
+> These changes can be viewed at [CDCgov/dibbs-ecr-refiner#1525](https://github.com/CDCGov/dibbs-ecr-refiner/issues/1525).
+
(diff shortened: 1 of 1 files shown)


docs/decisions/0013_2026-07-22_zero-code-set-refinement.md
```diff
@@ -0,0 +1,273 @@
+# 13. Zero-Code-Set Refinement
+
+Date: 2026-07-22
+
+## Status
+
+Proposed
```
fzhao99:
```suggestion
Deprecated 
```

Think we might want to add a note that this is a "won't do" bc of stakeholder decisions and maybe add a note in the context section

fzhao99 (commented):
one housekeeping suggestion, but otherwise lgtm!

fzhao99 (approved):

PR #1559, opened 2026-07-24
feat(locks): rework config lock lifecycle

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Reworks the configuration lock lifecycle. Previously the client only
released a lock via `sendBeacon` on `beforeunload`, so locks were never
explicitly acquired on page entry and SPA navigation between the build
and activate views could drop or strand locks.

- Adds `POST /api/v1/configurations/{id}/acquire-lock` endpoint;
  `release-lock` moves under the same `/{configuration_id}` router
  prefix.
- Replaces the `useConfigLockRelease` hook with `useConfigLock`:
  acquires/refreshes the lock on mount, schedules release on unmount
  with a 150ms delay, and cancels pending releases on remount. This
  keeps the lock stable across React StrictMode double-invokes and
  build <-> activate SPA navigation.
- Updates client test setup to mock `global.fetch` and
  `navigator.sendBeacon`.
- Regenerates `openapi.json`.

## 🔗 Related Issue

- Discovered while working on #1548

## ✅ Acceptance Criteria

- Opening a configuration acquires the lock immediately on mount.
- Navigating between build and activate views does not drop the lock.
- Closing the tab or navigating away releases the lock via beacon.
- StrictMode double-mount does not prematurely release the lock.

## 🧪 How to test

Unit tests:

1. `just client test src/hooks/useConfigLock.test.ts`

Testing the application:

1. Run `just dev up`
1. Check that all services are up and running with `docker ps`
1. In your browser, navigate to [http://localhost:8081](http://localhost:8081)
1. Open a configuration in one browser, then open the same
   configuration as a second user; the second user should see the
   lock banner immediately.
1. Navigate between the build and activate steps as the first user;
   the lock should remain held (no banner flicker for user two).
1. Close the first user's tab; the lock should release and user two
   should be able to edit.

## ℹ️ Additional Information

Split out of the SPIKE #1525 prototype branch
(`rogeruiz/1525/spike-prototype`); this lock rework is a general
improvement independent of the zero-code-set work.


fzhao99 (approved):
🚢

PR #1561, opened 2026-07-24
feat(server): add plan of treatment to reconstruction

robertmitchellv:
# 🔀 PULL REQUEST

## 💡 Summary

This work adds the Plan of Treatment section (`18776-5`) to the set of sections that support reconstruction. It joins Results, Problems, Immunizations, and Medications Administered, which already had it.

**Why it was more than a copy-paste**

The four existing sections are homogeneous--every entry is the same kind of thing (all results, all problems, etc.), so each renders as one table. Plan of Treatment is the first heterogeneous section: a single <section> holds five unrelated clinical statements side by side:

- planned observations
- procedures
- activities
- medications
- immunizations

Each has different attributes, so they can't share one table.

The solution: one captioned table per entry kind. A reader gets a "Planned Procedures" table, a "Planned Medications" table, and so on, only for the kinds that actually have surviving entries.

**How it works**

The reconstruction engine is built in three layers, and most of this work reused the first and added to the second and third:

1. Shared renderers (reused as-is)--turn a coded element into "Azithromycin 500 MG Oral Tablet (RxNorm `248656`)", format HL7 dates into readable ones, etc. One new renderer was added here: performer--resolves "who is expected to do this" to the clinician's name if present, otherwise the organization.
2. Field maps (new data)--for each of the five planned entry kinds, a plain list of "this column ← this piece of the entry." These came from a spreadsheet Tim produced (refiner-narrative.xlsx), which we treat as the minimum correct set. We ship a small superset: Status and Performer on every table, because both are always present in the structured data and are exactly what a public health reviewer wants to know about a plan.
3. The section function (new code)--groups entries by kind and emits a table for each non-empty one. The only subtlety: medications and immunizations are the same XML element (`substanceAdministration`) and are told apart by their template ID--the same way the refiner's matching rules already distinguish them.

A key guarantee runs through all of this: every column must be reproducible from the surviving entries alone. The reconstructed narrative is stamped as "derived from the entries," which is a formal claim that the two are clinically equivalent--so if a value can't be sourced from an entry, it doesn't belong in the table.

What's tested

- Unit tests pin all five table shapes, the medication/immunization split, the performer rendering, the empty-section fallback, and the captioning.
- An integration scenario runs a real eICR through the full production pipeline and validates the output against the CDA R2 schema and schematron — so we know the reconstructed narrative is not just correct but conformant.

Deliberately deferred

- Indication column. Some planned items carry a nested "indication" (the condition driving the plan). It's not in the SME spreadsheet and every other section has the same latitude, so we left it out — a clean follow-up if we want stricter coverage.
- Non-Epic sample data. Our reference for "what content belongs in these tables" is Epic, since it's the most consistent long-term producer. The design stays vendor-neutral (no Epic-specific formatting hard-coded), but we'd benefit from sample documents from other EHRs to confirm the field maps hold up.

## 🔗 Related Issue

Fixes #1483 

- #1483


refiner/app/services/ecr/narrative/README.md
```diff
@@ -52,10 +64,65 @@ narrative elements directly.
...
+
+Reconstruction is the one narrative writer that MUTATES surviving entries: it
+strips the now-dangling source references, relinks each entry to its minted
+row, and stamps `typeCode="DRIV"`. It only runs on the refine path — a retained
+section never reconstructs, and when nothing survived (or a section has no
+registered reconstructor) it falls back to retaining the original narrative.
+
+On house style: the reconstruction stays vendor-neutral — it does not encode
+one EHR's stylesheet quirks (see
+`docs/decisions/0011_2026-06-24_narrative-reconstruction-real-data-blocks-and-linkage.md`).
```
jakewheeler: You might already know this but the doc can be directly linked to like this
```suggestion
one EHR's stylesheet quirks ([see here](/docs/decisions/0011_2026-06-24_narrative-reconstruction-real-data-blocks-and-linkage.md)).
```

refiner/app/services/ecr/narrative/reconstruction.py
```diff
@@ -508,9 +596,24 @@ def extract_fields(anchor: _Element, field_map: list[FieldSpec]) -> dict[str, st
...
 
 # prepended to a negated substanceAdministration's leading cell so the row reads
-# as the negative it is (a not-administered / "no known" statement) rather than
-# as an administered product
+# as the negative it is rather than as a product. the wording follows moodCode:
+# negating an EVN statement says the act did NOT happen ("No Known Medications",
+# a refused vaccine); negating a planned one says it is NOT going to be done --
+# a contraindication or a cancelled order, not a missing administration
 _NEGATED_PREFIX: str = "Not administered: "
+_NEGATED_PLANNED_PREFIX: str = "Not planned: "
```
jakewheeler: We could define these values within `_negated_prefix()` if we want to ensure they're not used directly. It looks like calling the function is the intended usage method.

Your call on whether or not we want to change it. This is all very self contained so I'm not overly concerned about it.

refiner/tests/unit/test_service_policy.py
```diff
@@ -57,9 +58,21 @@ def test_active_reconstructable_sections(self):
...
+        """
+        The policy gate and the dispatch table must not drift.
+
+        `is_reconstructable_section` is what the API lets a jurisdiction
+        configure; SECTION_RECONSTRUCTORS is what actually runs. A code in
+        the first but not the second silently degrades to the retained
+        narrative at refinement time.
+        """
+
+        assert set(RECONSTRUCTABLE_SECTIONS) == set(SECTION_RECONSTRUCTORS)
```
jakewheeler, line 74: 👍

refiner/tests/unit/test_service_narrative_reconstruction.py
```diff
@@ -13,11 +13,13 @@
     reconstruct_immunizations,
```
jakewheeler, line 13: Nice tests 👀

refiner/app/services/ecr/narrative/README.md
robertmitchellv: oh yeah--not sure why i didn't do that

jakewheeler (approved):
Left a few minor potential changes but looks great overall Robert!

refiner/app/services/ecr/narrative/reconstruction.py
robertmitchellv: there's another place where this ends up coming up and i think it might be best to do something like this instead?

```python
def _negated_prefix(source: _Element) -> str:
      """
      Return the negation prefix appropriate to an anchor's moodCode.

      Prepended to a negated substanceAdministration's leading cell so the
      row reads as the negative it is rather than as a product. Negating an
      EVN statement says the act did NOT happen ("No Known Medications", a
      refused vaccine); negating a planned one says it is NOT going to be
      done -- a contraindication or a cancelled order, not a missing
      administration.

      Absent `@moodCode` is treated as EVN: it is the CDA default for the
      clinical statements the flat reconstructors anchor on.
      """

      mood = source.get("moodCode") or "EVN"
      return "Not administered: " if mood == "EVN" else "Not planned: "
```

PR #1601, opened 2026-07-29
fix: lazy load environment variables

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR is the first step in addressing a longstanding issue in which we were unable to import files that import `app/core/config.py` somewhere down the chain into `lambda_function.py` or files used by the ops container.

The reason for this is that we were previously using the `ENVIRONMENT` dictionary which attempted to read all of the variables required by the API. However, in the case of Lambda for example, many of these variables are not available in the environment and we shouldn't expect them to be there.

> [!NOTE]
> This only aims to solve imports related to environment variables. This won't, for example, allow you to import a file into `lambda_function.py` that imports `fastapi`.

## 🧪 How to test

Shut down and start your docker compose environment back up.

- App should work exactly as it did before
- All automated tests should be passing

The clearest way to see that this is working as intended is to notice that `lambda_function.py` now imports this line 

```python
from app.core.config import get_env_variable
```

This previously wasn't possible because `ENVIRONMENT` would have attempted to find env variables that didn't exist and the Lambda would die off immediately. That's why we were using the workaround of importing this function like this:

```python
from app.core.utils import get_env_variable # <-- utils didn't import from config.py
```


refiner/app/core/config.py
```diff
@@ -17,16 +18,87 @@
...
+
+class S3Config:
+    """S3 config. Needed by API, Lambda, Ops."""
+
+    def __init__(self) -> None:  # noqa: D107
+        self.S3_BUCKET_CONFIG: str = get_env_variable("S3_BUCKET_CONFIG")
+        self.AWS_REGION: str = get_env_variable("AWS_REGION")
+
+
+@lru_cache
```
jakewheeler, line 75: The intention of this is to only create a single instance and then reuse it upon calling the function again later on, which we do often.

refiner/app/core/config.py
```diff
@@ -17,16 +18,87 @@
...
+    Returns:
+        str: Name of the environment variable
+    """
+    value = os.getenv(name)
+    if value is None:
+        raise OSError(f"Missing environment variable: {name}")
+    return value
+
+
+class AppConfig:
```
jakewheeler, line 40: I'm not fully convinced on how I split this up. For example, we don't use the `AWS_REGION` variable in Lambda but we need it for the app and for #1488  I believe.

Let me know if you have suggestions.

refiner/tests/integration/scenarios/test_all_sections_covid_influenza.py
```diff
@@ -4,8 +4,8 @@
 import pytest
 
 from app.services.pipeline import RefinementResult
+from tests.utils.normalize_xml import normalize_xml
 
-from ...unit.conftest import normalize_xml
```
jakewheeler, line 8: This was causing an issue related to my changes because it was picking up the env variables meant only for the unit tests. Since they are defined at the module level it was overwriting values.

refiner/app/core/config.py
```diff
@@ -17,16 +18,87 @@
...
+        self.AUTH_CLIENT_SECRET: str = get_env_variable("AUTH_CLIENT_SECRET")
+        self.AUTH_ISSUER: str = get_env_variable("AUTH_ISSUER")
+
+
+class S3Config:
+    """S3 config. Needed by API, Lambda, Ops."""
+
+    def __init__(self) -> None:  # noqa: D107
+        self.S3_BUCKET_CONFIG: str = get_env_variable("S3_BUCKET_CONFIG")
+        self.AWS_REGION: str = get_env_variable("AWS_REGION")
```
rogeruiz: `AWS_REGION` isn't S3 specific here. I think if we need it in more places, we should move it to `AppConfig` or a specific `AwsConfig` class in the future.

refiner/app/core/config.py
```diff
@@ -17,16 +18,87 @@
...
+
+    Raises:
+        OSError: raised if environment variable is not present
+
+    Returns:
+        str: Name of the environment variable
+    """
+    value = os.getenv(name)
+    if value is None:
+        raise OSError(f"Missing environment variable: {name}")
```
rogeruiz: small nit; I wonder if we should change this to a `ValueError` or create a new `ConfigError(BaseApplicationException)` to better standardize the type since OSError is semantically imprecise.

refiner/app/core/config.py
rogeruiz: I mentioned it in another comment, but an AwsConfig class or AppConfig might make more sense for `AWS_REGION`.

refiner/app/services/logger.py
```diff
@@ -45,15 +45,15 @@ def filter(self, record: logging.LogRecord) -> bool:
         return True
 
 
-def setup_logger() -> logging.Logger:
+def setup_logger(app_config: AppConfig) -> logging.Logger:
     """
     Called to initially configure the logger.
     """
```
rogeruiz, line 51:
```suggestion
def setup_logger(app_config: appconfig | none = none) -> logging.logger:
    """
    called to initially configure the logger.
 
    args:
        app_config (appconfig | none): optional application config. if not
            provided, resolves via deferred `get_app_config()` call. this
            avoids a hard coupling between logger setup and config resolution,
            making the function usable in contexts where config hasn't been
            explicitly instantiated (cli scripts, migrations, etc.).
    """
    if app_config is none:
        from app.core.config import get_app_config
 
        app_config = get_app_config()
 
    _logger.setlevel(logging.info)
 
    handler = logging.streamhandler()
```

What do you think of something more along these lines? This would make the `app_config` optional so if the logger is set up from a CLI script, migration, or any entry point that isn't `asgi.py`, the caller doesn't need to import and resolve config first and it'll just happen automatically. The existing callers that pass `app_config` explicitly still work unchanged.

rogeruiz (approved):
lgtm 🌈 

Good work on this @jakewheeler I left some comments that we could address here or in a fast follow-up. But they're more suggestions  than anything else for now.

refiner/app/services/logger.py
jakewheeler: My only hesitancy with making this change now is that it makes it look like `setup_logger()` could readily be used outside of the app but even with this change that's not the case due to the required `python-json-logger` dependency and the other config that makes it app specific for now (like setting a `request_id`).

It could make sense to generalize it though if we feel we'd like a single way to configure loggers for our various use cases!

refiner/app/core/config.py
jakewheeler: `AwsConfig` seems like a good move for now 👍 thanks

PR #1604, opened 2026-07-30
fix(dx): run the correct server tests

rogeruiz:
# 🔀 PULL REQUEST

## 💡 Summary

Since #1438, this recipe would fail because it would just call `just server::test`. Since this is a capture-all test recipe, I've updated it to be `just server::test-all` instead.

## 🔗 Related Issue

- n/a

## ✅ Acceptance Criteria

- n/a

## 🧪 How to test

1. Run `just test`
1. All server tests should run sequentially
1. Tests should all pass.

### DIFF
--- a/justfile
+++ b/justfile
@@ -87,7 +87,7 @@ lint:
 [doc('Run tests on all code')]
 [group('test')]
 test:
-    just server::test
+    just server::test-all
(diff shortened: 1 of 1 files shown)


jakewheeler (approved):
🫡

PR #1609, opened 2026-07-30
chore: drop version column from codes

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Changes to the codes table / seed script to support version-agnostic codes. Part of the TES diff work to help with performance of DB queries 

## 🔗 Related Issue
Task within #1430 
- #1430

## ✅ Acceptance Criteria

- [ ] Codes table doesn't have version information
- [ ] All codes <> conditions information remains the same
- [ ] App doesn't have degraded functionality

## 🧪 How to test
- Run a refresh on main with `just db refresh` 
- Switch to this branch and run migrate. Note the codes should be shrinking (150912 -- > 89146 unique codes), but the relationships should stay the same (583,955 total relationships (558 child_rsg, 583,397 non_child_rsg)). Check the child_rsg codes numbers and the relationship join numbers are the same
- Verify app functionality is the same
- Run a down migration to check the db gets reset correctly

### DIFF
--- a/refiner/app/db/codes/db.py
+++ b/refiner/app/db/codes/db.py
@@ -14,14 +14,15 @@ async def get_rsg_codes_by_condition_id_db(
     """
 
     query = """
-        SELECT c.display, c.code, c.version, c.system_id
+        SELECT c.display, c.code, tes.version, c.system_id
         FROM conditions_codes as cc
         LEFT JOIN codes c on c.id = cc.code_id
-        WHERE cc.condition_id = %s AND cc.is_child_rsg;
+        LEFT JOIN conditions cond on cond.id = %(condition_id)s
+        LEFT JOIN tes on cond.tes_id = tes.id
+        WHERE cc.condition_id = %(condition_id)s AND cc.is_child_rsg;
     """
-    params = (condition_id,)
     async with db.get_connection() as conn:
         async with conn.cursor(row_factory=class_row(DbCode)) as cur:
-            await cur.execute(query, params)
+            await cur.execute(query, {"condition_id": condition_id})
 
             return await cur.fetchall()
--- a/refiner/migrations/20260729154745_drop_codes_version_col.sql
+++ b/refiner/migrations/20260729154745_drop_codes_version_col.sql
@@ -1,4 +1,12 @@
 -- migrate:up
+
+-- migrate:up
+CREATE INDEX IF NOT EXISTS idx_conditions_codes_code_id
+    ON conditions_codes (code_id);
+
+CREATE INDEX IF NOT EXISTS idx_conditions_codes_condition_id
+    ON conditions_codes (condition_id);
+
 -- Replace the unique constraint to scope to the existing codes
 ALTER TABLE codes
     DROP CONSTRAINT IF EXISTS codes_system_id_version_value_key;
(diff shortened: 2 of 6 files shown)


refiner/scripts/seeding/load_static_data.py
```diff
@@ -199,15 +198,14 @@ def _build_codes(
...
-                        version=cond_version,
                         display=display,
                         system_id=snomed_db_id,
                     )
                 )
 
-        condition_to_code_relationships[cond_index]["child_rsg_codes"] = (
-            condition_child_rsg_snomed_codes
-        )
+        condition_to_code_relationships[(cond_canonical_url, cond_version)][
```
fzhao99, line 206: we need to keep the relationship indexing version aware so COVID v5 _and_ COVID v6 has a join into code 1234, but can drop version from the code row since that's represented in the join into the condition now

refiner/schema.sql
```diff
@@ -130,6 +130,25 @@ SET default_tablespace = '';
 
 SET default_table_access_method = heap;
 
+--
+-- Name: active_payload_schema_reactivations; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.active_payload_schema_reactivations (
```
fzhao99, line 137: @lina-roth if you wouldn't mind double checking this code-gened table looks right, would be appreciated!

refiner/app/db/codes/db.py
```diff
@@ -14,14 +14,15 @@ async def get_rsg_codes_by_condition_id_db(
...
-        WHERE cc.condition_id = %s AND cc.is_child_rsg;
+        LEFT JOIN conditions cond on cond.id = %(condition_id)s
+        LEFT JOIN tes on cond.tes_id = tes.id
+        WHERE cc.condition_id = %(condition_id)s AND cc.is_child_rsg;
     """
-    params = (condition_id,)
     async with db.get_connection() as conn:
         async with conn.cursor(row_factory=class_row(DbCode)) as cur:
-            await cur.execute(query, params)
+            await cur.execute(query, {"condition_id": condition_id})
```
jakewheeler, line 26: I would love to see this change everywhere. Much nicer than the tuple params 🥲

refiner/migrations/20260729154745_drop_codes_version_col.sql
```diff
@@ -0,0 +1,100 @@
...
+SELECT
+    c.system_id,
+    c.code,
+    t.version,
+    c.display,
+    c.created_at
+FROM conditions_codes cc
+LEFT JOIN codes c ON c.id = cc.code_id
+LEFT JOIN conditions cond ON cc.condition_id = cond.id
+LEFT JOIN tes t ON cond.tes_id = t.id
```
jakewheeler: It's not really going to make a difference in practice but I think using `JOIN` instead of `LEFT JOIN` would technically be more correct for what we're expecting to see here. In practice if we did receive NULL values for some reason (we shouldn't) it'd hit the line below anyway and carry on.

I'm not too bothered by it if we decide to leave it. I was just thinking about it and figured I'd mention it.

refiner/scripts/seeding/load_static_data.py
```diff
@@ -164,18 +165,16 @@ def _build_codes(
     oid_indexed_system_db_ids: dict[SystemOid, SystemDbId],
     condition_to_code_relationships: ConditionToCodeRelationshipIndex,
 ) -> ProcessedCodePayload:
-
     snomed_db_id = oid_indexed_system_db_ids[SNOMED_OID]
-    codes_seen_so_far: set[tuple[str, str, str]] = set()
+    codes_seen_so_far: set[DbCodedConcept] = set()
```
jakewheeler, line 169: 👍

jakewheeler (approved):
Left one very minor feedback point we may or may not want to address but otherwise looks good!

PR #1611, opened 2026-07-31
chore: show which `ops/entrypoint.sh` vars are missing

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

I've made a small update to this script to tell you exactly which variables are missing. This should make debugging a little easier, especially when we see failures in dev.

I also realized that `VERSION` is not required, so I've removed it.

Instead of an error like this if `ENV` and `DB_URL` were missing, for example:
```sh
ERROR: ENV, VERSION, DB_URL, DB_PASSWORD, AWS_REGION, and S3_BUCKET_CONFIG must be set
```

You'll now see something like this:
```sh
ERROR: The following required variables are not set: ENV DB_URL
```

## 🧪 How to test

1. Run `docker compose build ops --no-cache`
2. Delete a variable or two from the `ops` service in `docker-compose.yml`
3. Run `docker compose --profile ops run --rm ops prepare-db`
4. Check that the error message reflects the variables you deleted

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,12 +41,11 @@ services:
       - ./refiner/app/db:/var/task/app/db
       - ./refiner/assets:/var/task/assets
 
+# Run with `docker compose --profile ops run --rm ops prepare-db`
   ops:
     build:
       context: .
       dockerfile: Dockerfile.ops
-      args:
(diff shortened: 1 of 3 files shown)


lina-roth (approved):
this is a good update 🎉

PR #1613, opened 2026-07-31
feat: tes update view

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Adds the TES diff view into the updates page.

https://github.com/user-attachments/assets/14968773-0ea9-4706-b43c-68f1db527a1c

Also added a flag to have the seed script seed all files locally
  
## 🔗 Related Issue
Fixes #1430 
- #1430

## ✅ Acceptance Criteria
- [x] Adds a table for each TES update that includes:
   - [x] All new or updated code sets (condition groupers)
   - [x] Number of added & removed codes per code set
   - [x] "New condition" flag for new conditions (duh)

## 🧪 How to test
- Run full seed by setting  `SEED_ALL_TES_DATA=true` for the seed service in the override docker-compose file and spin up the app
- Navigate to the TES update page
- Notice that the page renders with all of the updated values for TES updates 1-6
- Notice that between versions 1 and 2, there are several values with the "new condition" pill on them if they didn't appear in version one

## ℹ️ Additional Information

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


client/src/api/events/events.ts
```diff
@@ -348,9 +348,3 @@ export function useGetEventsExportApiV1EventsExportGet<TData = Awaited<ReturnTyp
 
   return withQueryKey(query, queryOptions.queryKey);
 }
-
```
fzhao99: I think these are a result of running / not running precommit to trim trailing white spaces on these files, which I wasn't doing before / probably introduced as a result 🙈

client/e2e/tes-updates.spec.ts
```diff
@@ -8,10 +8,21 @@ test.describe('TES updates page', () => {
...
+    page,
   }) => {
     await tesUpdatesPage.goToTesUpdate('5.0.0');
     await expect(makeAxeBuilder).toHaveNoAxeViolations();
 
     await tesUpdatesPage.goToTesUpdate('6.0.0');
+
+    const firstRowVersion6 = page.getByRole('row').first();
+    expect(firstRowVersion6.getByText('Acanthamoeba')).toBeDefined();
+    // since the numbers for the diff are checked in the integration test in
```
fzhao99, line 20: opted to test the validity of the numbers as an integration test and just validate the display values here. Lmk if we feel this is appropriate or if we want to cover the cases in a different way.

client/src/api/events/events.ts
fzhao99: fixed in [this PR](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1631)

refiner/tests/integration/test_tes_diff.py
```diff
@@ -0,0 +1,35 @@
...
+        expected_acanthamoeba = diff_6_and_5[0]
+        assert expected_acanthamoeba.display_name == "Acanthamoeba"
+        assert expected_acanthamoeba.added_code_total == 6
+        assert expected_acanthamoeba.removed_code_total == 0
+
+        expected_rubella = [c for c in diff_6_and_5 if c.display_name == "Rubella"][0]
+        assert expected_rubella
+        assert expected_rubella.added_code_total == 1156
+        assert expected_rubella.removed_code_total == 3
+        assert not expected_rubella.is_new
```
fzhao99: open to suggestions about how to test the `is_new` case here, as well as the maintainability of the test generally. To have a determinate test, think we'd need to have a full seed of all the TES data since otherwise, this test would need to get updated whenever the default / previous fixtures get updated (ie each time we have a TES bump), but obviously seeding the whole TES to run the integration tests is pretty heavyweight // might not be worth the extra setup.

Think it'd be pretty trivial to write the tests otherwise, so it's mainly a q on how much we want to slow down our pipeline by running a full seed in integration tests.

client/src/pages/TesUpdates/TesVersionDetails.tsx
```diff
@@ -0,0 +1,79 @@
...
+          {response.data.map((r) => {
+            const shouldShowNewConditionPill =
+              // don't show pill if oldVersion is undefined (the "baseline" config)
+              // since everything in that version would be a new condition
+              r.is_new && oldVersion != '';
+            return (
+              <tr key={r.canonical_url}>
+                <td className="px-2 py-3">
+                  {r.display_name}{' '}
+                  {shouldShowNewConditionPill ? <NewConditionPill /> : <></>}
```
jakewheeler:
```suggestion
                  {shouldShowNewConditionPill ? <NewConditionPill /> : null}
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -1,21 +1,35 @@
...
+
+export interface TesDiffInformation {
+  selected_update: TesUpdate;
+  prev_update: TesUpdate | null;
+}
 
 export function TesUpdates() {
   const { data: tesUpdates, isPending, isError } = useGetTesUpdates();
-  const [selectedUpdate, setSelectedUpdate] = useState<TesUpdate | null>(null);
+  const [tesDiff, setTesDiff] = useState<TesDiffInformation>();
```
jakewheeler: Very minor but I like this idea of this being explicitly `null` by default instead of implicitly `undefined`. It seems much more obvious to me that this is intentional this way. What do you think? This should not impact the code/functionality otherwise.

```suggestion
  const [tesDiff, setTesDiff] = useState<TesDiffInformation | null>(null);
```

client/src/pages/TesUpdates/index.tsx
```diff
@@ -1,21 +1,35 @@
...
   if (isError) return 'Error occurred!';
 
-  if (!selectedUpdate) {
-    setSelectedUpdate(tesUpdates.data.tes_updates[0]);
+  if (!tesDiff) {
+    const newestUpdate = tesUpdates.data.tes_updates[0];
+    const prevUpdate =
+      tesUpdates.data.tes_updates.length >= 2
+        ? tesUpdates.data.tes_updates[1]
+        : null;
```
jakewheeler, line 27: We could consider calculating these server side. Not the biggest deal currently if we'd like to leave it but figured I'd suggest it in case we don't want to rely on list order or perform the logic client side.

refiner/app/api/v1/tes.py
```diff
@@ -59,6 +62,59 @@ async def get_tes_updates(
     return TesResponse(
         tes_updates=[
             TesUpdate(id=t.id, version=t.version, created_at=t.created_at)
-            for t in updates
+            for t in sorted(updates, key=lambda x: x.version, reverse=True)
```
jakewheeler: I might suggest creating a function in `services/tes.py` that performs the list sorting by version using `parse` from `package.version`. An incorrect sort would break the UI at the moment. Not sure how likely that would be in reality but probably worth protecting against it since it doesn't take much effort.

`get_latest_tes_version()` in that file is a good example to work from.

refiner/app/db/tes/db.py
```diff
@@ -22,3 +30,158 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[DbTes]
             await cur.execute(query)
             rows = await cur.fetchall()
             return rows
+
+
+async def get_tes_by_version_number_db(
```
jakewheeler: Can we make this "private"?
```suggestion
async def _get_tes_by_version_number_db(
```

refiner/app/db/tes/db.py
```diff
@@ -22,3 +30,158 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[DbTes]
...
+    """
+
+    async with db.get_connection() as conn:
+        async with conn.cursor(row_factory=class_row(DbTesConditionUpdate)) as cur:
+            await cur.execute(query, {"tes_id": tes_id})
+            result = await cur.fetchall()
+            return result
+
+
+async def get_tes_update_diff_db(
```
jakewheeler: Same thing here.
```suggestion
async def _get_tes_update_diff_db(
```

refiner/app/db/tes/db.py
```diff
@@ -22,3 +30,158 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[DbTes]
...
+    async with db.get_connection() as conn:
+        async with conn.cursor(row_factory=class_row(DbTesConditionUpdate)) as cur:
+            await cur.execute(
+                query, {"cur_tes_id": cur_tes_id, "prev_tes_id": prev_tes_id}
+            )
+            rows = await cur.fetchall()
+            return rows
+
+
+async def get_tes_version_diff(
```
jakewheeler:
```suggestion
async def get_tes_version_diff_db(
```

refiner/app/db/tes/db.py
```diff
@@ -22,3 +30,158 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[DbTes]
...
+    Returns an array off all loaded TES version records.
+    """
+    cur_tes_record = await get_tes_by_version_number_db(db=db, version=cur_tes_version)
+
+    if prev_tes_version:
+        prev_tes_record = await get_tes_by_version_number_db(
+            db=db, version=prev_tes_version
+        )
+    else:
+        prev_tes_record = cur_tes_record
```
jakewheeler: Using a ternary here is pretty clean
```suggestion
    prev_tes_record = (
        await get_tes_by_version_number_db(db=db, version=prev_tes_version)
        if prev_tes_version
        else cur_tes_record
    )
```

refiner/app/db/tes/db.py
```diff
@@ -22,3 +30,158 @@ async def get_loaded_tes_versions_db(db: AsyncDatabaseConnection) -> list[DbTes]
...
+            return rows
+
+
+async def get_tes_version_diff(
+    db: AsyncDatabaseConnection, cur_tes_version: str, prev_tes_version: str | None
+) -> list[DbTesConditionUpdate]:
+    """
+    Returns an array off all loaded TES version records.
+    """
+    cur_tes_record = await get_tes_by_version_number_db(db=db, version=cur_tes_version)
```
jakewheeler: We don't handle the potential `ValueError` that could come from `get_tes_by_version_number_db()`. Should we? FastAPI will respond with a 500 if it doesn't work.

refiner/app/db/tes/model.py
```diff
@@ -13,3 +13,16 @@ class DbTes:
     version: str
     created_at: datetime
     updated_at: datetime
+
+
+@dataclass
+class DbTesConditionUpdate:
```
jakewheeler, line 19: I keep debating myself on how to handle these cases where we return a shape from the database but it doesn't map directly to a type in the database, which was the original intention.

Nothing to do! Just something I've been thinking about. I think the meaning of the `Db` prefix has shifted over time but that's not necessarily bad or wrong.

refiner/scripts/seeding/lib.py
```diff
@@ -589,12 +589,12 @@ def is_condition_grouper(vs: dict) -> bool:
     return any("conditiongroupervalueset" in str(prof).lower() for prof in profiles)
 
 
-def collect_files_to_parse(is_local: bool, versions_to_keep=2) -> list[Path]:
+def collect_files_to_parse(seed_all_tes_data: bool, versions_to_keep=2) -> list[Path]:
```
jakewheeler, line 592: Good change 👍 For some reason I was thinking we made this change in the initial PR.

refiner/tests/integration/test_tes_diff.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.mark.asyncio
+class TestTesDiff:
+    async def test_tes_diff_between_conditions(
+        self, default_tes_version, previous_tes_version, db_pool
+    ):
+        diff_6_and_5 = await get_tes_diff_details(
+            cur_version=default_tes_version,
+            prev_version=previous_tes_version,
+            db=db_pool,
+        )
```
jakewheeler: Probably worth hitting the API endpoint instead of the function directly. We hit the endpoint in the e2e tests but we usually do for integration tests as well since that's the only way the function can be invoked.

refiner/tests/integration/test_tes_diff.py
jakewheeler: The actual impact of an incorrect diff on a user is very low IMO since it's only really informational. I don't feel overly concerned about this testing being extremely thorough. 

Perhaps we just stick with checking version 6 vs. version 5 for now? Meaning we hard-code these values. If this test ever starts failing we will know something is actually broken.

It might be easier to mock out the data if we want to check `is_new`.

fzhao99: yeah I agree, think this ultimately will be ok. Don't think the value of mocking out the data insert would be worth the extra value of the test

refiner/app/db/tes/db.py
fzhao99: Added in a 404 in the event that one of the find operations from the DB raise. Let me know if there's a better error to return

client/src/pages/TesUpdates/index.tsx
fzhao99: Think since it's just the initial load state, I'll leave it on the frontend in the event we do/don't want to auto-click the first TES in the future.

refiner/app/api/v1/tes.py
```diff
@@ -50,15 +42,64 @@ async def get_tes_updates(
...
+                is_new=c.is_new,
+            )
+            for c in sorted(conditions_changed, key=lambda x: x.display_name)
         ]
-    )
+    except ValueError:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Specified TES version(s) {cur_version} or {prev_version} not found.",
+        )
```
jakewheeler, line 105: 💯

jakewheeler (approved):
Great job! 👏

PR #1616, opened 2026-07-31
refactor: custom codes functionality clean up

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR cleans up some of the custom codes related application code, noted in #1528.

- Adapt UI code to handle modified backend types / responses
- Moves custom codes API code into `api/v1/configurations/custom_codes`
- Moves custom codes DB code into `db/configurations/custom_codes`
- Removes `IndexedCodeSystem`
  - We can access custom codes by ID now that they all have a unique ID 
- Removes hand-written API body shape validation in favor of letting FastAPI handle it automatically
- Adds a `find_code_system_by_display_name()` function to `services/code_systems.py`
- Updated tests to account for API / DB function changes

## 🔗 Related Issue

Fixes #1591 

- #1591 

## 🧪 How to test

- There are no functionality changes introduced to the app. Please check that individual custom code actions work as expected and that the CSV upload functionality is also working correctly
  - API responses have changed in some instances but it won't impact the application at all 
- All automated tests should continue to pass


refiner/app/db/configurations/custom_codes/db.py
```diff
@@ -0,0 +1,290 @@
...
+        updated_at,
+        configuration_id
+    FROM custom_codes
+    WHERE configuration_id = %s
+    """
+    params = (configuration_id,)
+
+    async with db.get_connection() as conn:
+        async with conn.cursor(row_factory=class_row(DbCustomCode)) as cur:
+            await cur.execute(query, params)
```
fzhao99, line 36: do we want to take the chance here to swap these out with the named param pattern that we were talking about in the other PR?

client/src/pages/Configurations/ManageCodes/CustomCodes/CsvImport/Modals.tsx
```diff
@@ -93,7 +92,7 @@ interface PreviewEditModalProps {
   closePreviewEditModal: () => void;
   setError: React.Dispatch<React.SetStateAction<UploadError | null>>;
   error: UploadError | null;
-  codeSystems: IndexedCodeSystem;
+  codeSystems: DbCodeSystem[];
```
fzhao99, line 95: since this is an array now, think we may be able to simplify the `Object.values` iteration [further down this file](https://github.com/CDCgov/dibbs-ecr-refiner/blob/fd84321f7bfeeec186b33a54570b8dfdc00ced68/client/src/pages/Configurations/ManageCodes/CustomCodes/CsvImport/Modals.tsx#L192) into a flat `.map` and `.find`

refiner/app/db/code_systems/db.py
```diff
@@ -14,14 +13,11 @@ class DbCodeSystem:
     """
 
     id: UUID
-    key: CodeSystemKey
+    key: str
     display_name: str
     oid: str
 
 
-type IndexedCodeSystem = dict[CodeSystemKey, DbCodeSystem]
```
fzhao99, line 22: 🧹

fzhao99 (commented):
a few minor cleanup comments but otherwise lgtm!

client/src/pages/Configurations/ManageCodes/CustomCodes/CsvImport/Modals.tsx
jakewheeler: Yes, will do. Thanks!

fzhao99 (approved):
🎉

PR #1617, opened 2026-08-01
docs(decision): add the ADR for code-exclusion and overrides

### LINKED ISSUE #1560: [SPIKE] Investigate how we'll do overrides
## 🚧 What needs to be done

We need to figure out how we can (or want to) do overrides in terms of the refinement algorithm

## 🤔 Why it needs to be done

We can't implement the overrides feature without knowing how it'll work on the backend

## ⏰ Timebox

2-3 days

## 📝 To-do list

  - [ ] ...
- [ ] Research + Analysis
  - [ ] ...
  - [ ] ...
  - [ ] ...
- [ ] Review and Finalize
  - [ ] ...

## 📜 Additional context

The goal of overrides is to remove specific coded information from the refined output. How we go about doing that is up to us to define. For example:
- Do we just remove the entire entry if it contains an overrides code?
- Are we able to more clinically excise or obfuscate the specific information while preserving the entry?

robertmitchellv:
# 🔀 PULL REQUEST

  

  

  ## 💡 Summary

  Spike write-up for letting a jurisdiction describe what it **does not** want in a
  refined eICR. The core reframing: this is two distinct needs on two different
  pipeline seams, not one mechanism — and after review, two ADRs rather than one.

  - **[ADR 0014 — code-exclusion](docs/decisions/0014_2026-07-29_code-exclusion.md).**
    Turn off a code a TES condition/code set contributes so it stops pulling
    entries into the output. Resolved as an **anti-join at projection time** inside
    `convert_config_to_storage_payload`'s condition loop, before `active.json` is
    built — nothing downstream changes, no backfill, and origin-scoping is free.
  - **[ADR 0015 — overrides](docs/decisions/0015_2026-08-05_overrides.md).**
    Guarantee a code never appears, _no matter what_, including in sections
    refinement never touches. Recommended as an `apply_overrides` **pre-pass**
    before `refine_eicr` that reuses the IG `EntryMatchRule`s in subtractive
    (inverted keep/drop) mode, falling back to entry-level removal where no rules
    exist.

  **Why two documents:** exclusion touches no XML and its open questions are
  resolved; overrides needs a refactor, a new pass, and new snapshots. They're on
  different tracks with different risk, and reviewing them as one doc was making
  both harder. 0015 refers back to 0014 for shared framing rather than restating it.

  Terminology now matches the application throughout — "exclusion" and "overrides."
  "Suppression" and "de-selection" are gone except for one note explaining the
  rename.

  Each ADR also captures resolved decisions (scope, granularity, provenance),
  prototype interfaces, test-scenario checklists, and the open questions the team
  should settle before implementation. 0015 keeps its YAGNI-first sequencing
  (extract the shared finder first, build overrides second, abstract the pruner
  only once a second consumer proves the shape).

  ## 🔗 Related Issue

  Fixes #1560

  - #1560
  - Exclusion implementation + tests: #1633

  ## ✅ Acceptance Criteria

  - [ ] Team agrees with the split and with the two recommended options (A1 / B1).

  ## 🧪 How to test

  Docs-only change — no application or test code is modified on this branch.

  1. Preview both ADRs rendered as GitHub-flavored markdown (the callouts and tables
     are the main things to eye).
  1. Spot-check that the `refiner/...` code references resolve to real files/lines.
     These were re-verified against `main` in the latest push; a few had drifted by
     4–8 lines and are corrected.
  1. 0014 describes shipped behavior. To check its claims against real code, see
     `robert/de-selection`, which has the migration, `get_code_exclusions_db`, the
     six `TestCodeExclusions` unit cases, and the integration tests the ADR names.

  ## ℹ️  Additional Information

  **One finding that didn't come from review and may need a product decision:**
  overrides as designed only guarantee removal of _coded_ data. Narrative `<text>`
  is not covered — `narrative="retain"` leaves `<text>` verbatim even when every
  entry is pruned, and the narrative-only sections (Chief Complaint, Reason for
  Visit, HPI, Review of Systems) have no entries to remove at all, which are exactly
  the sections most likely to carry stigmatizing free text.

  This is scoped out deliberately — removing coded data is tractable and testable,
  masking free text is a different problem with a much worse failure mode — but it
  means **we shouldn't describe overrides to jurisdictions as "this code will never
  appear in the output," only "will not appear as coded data."** Written up as a
  `CAUTION` in 0015 with test cases that pin the boundary, plus an open question
  sketching follow-up options.

  Also recorded: **jurisdiction-wide overrides are deferred, not dismissed.** We
  expect real value there, but starting at the configuration level tests the
  mechanism on the surface the runtime already has; widening later is a union at
  projection time and doesn't change the pass, the pruning strategy, or the
  provenance.


docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+to be retained.
+
+We want to add the ability to describe what a jurisdiction _does not_ want. From
+talking it through, there are two distinct needs, and they are **not** the same
+mechanism:
+
+1. **De-selection (the "overrides" case).** A jurisdiction is looking at the
+   codes a TES condition (or a code set) contributes and decides it does not
+   want one of them to participate in refinement. This is a statement about the
+   _want-set_: "this code should stop pulling entries into the output." It is
```
jakewheeler: I understand what you mean but it might be beneficial for folks if we avoid the term `overrides` for this case since the application feature we're calling "overrides" maps more closely to "suppression" in this list.

As you've seen already I've been referring to this as "exclusion" in my PR which I feel is distinct from other terms we've been using. Deselection is good too.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+- The DB remains the expressive source of truth: the UI can render "de-selected"
+  state, users can re-select, and the choice is auditable. `active.json` stays a
+  clean projection.
+
+**Cons**
+
+- The projection loses the distinction ("this code came from TES but was turned
+  off")--but that distinction lives in the DB where it belongs, so this is
+  acceptable.
+- Requires a new table + read path (mirrors the `custom_codes` work in #1528).
```
jakewheeler: One nice thing about this is that no data backfilling is required since the condition code sets in full are the default.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+
+#### A2. Subtract at `ProcessedConfiguration.from_dict` runtime
+
+Keep the full `code_system_sets` in `active.json` plus a separate `de_selected`
+set, and subtract on read in `terminology.py`.
+
+**Pros**: preserves the "was in TES, turned off" intent inside the payload.
```
jakewheeler: I'd argue this isn't even a pro since there is no gain from it 🙈 unless I'm not considering something?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+  present). Suppression beats inclusion with no reconciliation logic.
+- Reuses the IG rule metadata and pruning strategies instead of duplicating them.
+
+**Cons**
+
+- The container-prune decision is _inverted_ relative to positive refinement
+  (drop matched containers, keep the rest), so it is the same rule metadata
+  driving new keep/drop logic--a bounded amount of new code, not a free reuse
+  of `_prune_at_container_level`.
+- Two traversals of refinable sections (suppress, then refine). Cheap, and the
```
jakewheeler: Can you further explain why this is cheap? I'm not sure I'm fully understanding. I do think the logical split between suppress and then refine makes things clear.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+> [!IMPORTANT]
+> Direction not yet finalized--this document captures the recommended shape
+> from the spike so the team can react before implementation. The two
+> reframings below are the load-bearing conclusions.
+
+Recommended direction:
+
+- **Feature A (De-selection): option A1.** De-selected codes are stored in the
+  DB as first-class config state and anti-joined out during projection in
+  `convert_config_to_storage_payload`. Nothing downstream changes.
```
jakewheeler: 👍 I think you and I coming to the same conclusion independently is probably a good sign 😂 This is already working well for me in practice in #1633.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+- **Feature B (Suppression): option B1.** A `suppress_eicr` pre-pass runs before
+  `refine_eicr` across all sections, reusing IG `EntryMatchRule`s in subtractive
+  mode where they exist and falling back to entry-level removal where they do
+  not.
```
jakewheeler: I think this is good as well. My only real question is performance but it sounds like you're thinking this shouldn't be too much of an issue.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+   matching engines.** This is the structural reason B1 beats B3, and it should
+   be stated as an invariant wherever the pass is documented.
+
+> [!NOTE]
+> **Scope: eICR only; the RR is unaffected.** De-selection edits
+> `code_system_sets`, not the `included_condition_rsg_codes` the RR pass
+> (`refine_rr`) filters on--so a de-selected clinical code cannot change which
+> conditions the RR reports. Suppression is an eICR-only pre-pass. Neither
+> feature touches RR refinement; if a future need arises to suppress content from
+> the RR, it is a separate design.
```
jakewheeler: 👍 good note.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+
+Recording the reasoning path, not just the endpoints, so a later reader can see
+_why_ the design looks the way it does--and why some tempting alternatives were
+passed over.
+
+1. **"A way to say what we don't want" turned out to be two needs, not one.**
+   The starting ask was singular. The first useful move was noticing that
+   "de-select a code a condition contributes" and "never emit this code, ever"
+   behave differently: one edits the _want-set_, the other overrides the
+   _output_. Forcing them down one path would have compromised both.
```
jakewheeler: 💯

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+#### Storage (mirror `custom_codes`)
+
+`custom_codes` (schema.sql:263) is the pattern to copy: a per-configuration table
+keyed by `(configuration_id, system_id, code)`.
+
+```sql
+-- deselected_codes: codes the jurisdiction has turned OFF for a configuration.
+-- a row means "this (system_id, code) should NOT enter the want-set for this
+-- configuration, even though a linked condition/code set contributes it."
+CREATE TABLE public.deselected_codes (
```
jakewheeler: This differs slightly from my [WIP table](https://github.com/CDCgov/dibbs-ecr-refiner/pull/1633/changes#diff-2f20363f6b4034c5ea6649a577b849aa69ee1f9e258d182c3a361989c6db909bR3). Does what I've done work for what you think will be needed?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+codes (`configurations.py:222`) and condition codes (`configurations.py:248`),
+then calls `CodeSystemSets.from_dict`. The anti-join sits right before that call:
+
+```python
+# build the set of (system_key, code) pairs the jurisdiction turned off.
+# keyed by system so we never remove the same digits under a different system.
+deselected: set[tuple[CodeSystemKey, str]] = {
+    (code_systems[dc.system_id].key, dc.code)
+    for dc in configuration.deselected_codes
+}
```
jakewheeler: This is the right idea.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+  to.
+- **Leanness** where speed matters--the flat set is generated when needed, and
+  the persisted/served payload does not grow (consistent with the origin-free,
+  "only what refinement needs" principle above).
+
+> [!NOTE]
+> **Open for the team--please react.** Three things to settle:
+>
+> 1. **Where the system-aware model lives** and which object the flat set is
+>    projected from, in `convert_config_to_storage_payload`'s flow.
```
jakewheeler: Codes are really only going to be viewable/manageable in the app from the "manage codes" page which my PR is currently handling. All config usages (aka refining) should be going through `convert_config_to_storage_payload()`.

It should be straightforward to get all of a configuration's excluded/deselected codes by configuration ID, so I wouldn't be too concerned about how this happens. I think it's not unlikely that within `convert_config_to_storage_payload()` we have something like this:

```python
excluded_codes = await get_excluded_codes_db(configuration_id=config.id, db=db)
# do something with these
```
They could be attached to an object like `DbConfiguration` but there's no real need for it. Does this help? Let me know if I didn't quite hit on what you were looking for here.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+> 2. **Whether `validate_custom_code` should move to `(system, code)` granularity**
+>    to match the anti-join. Today it is deliberately broader (bare string), which
+>    _over_-blocks safely but is inconsistent with the chosen granularity--a
+>    jurisdiction cannot add "code X in a local system" if "code X" exists in
+>    SNOMED, even though those are different codes.
```
jakewheeler: I'm not a big fan of the rules around custom codes in general. I think it'd make sense to let folks add whatever they want and we'll de-dupe on the backend. I get that the idea is to not waste user's time adding codes that are a part of a code set but if users are uploading codes via CSV anyway I don't think it's much of a risk. Increased flexibility seems like it would outweigh potential time waste.

Something to consider perhaps but I don't know that we should do anything different with custom codes as a part of this work. Do changes to custom codes need to be included?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+> 3. **The scope key for de-selection.** De-selection is conceptually **per
+>    configuration, per condition, per jurisdiction** (the same scope as
+>    suppression; jurisdiction is implied by the configuration). The current
+>    `deselected_codes` sketch is keyed only by `(configuration_id, system_id,
+code)`--if de-selection is truly condition-scoped, that table needs a
+>    `condition_id` the way `suppressed_codes` does, and the anti-join must load
+>    the de-selected set **per condition** at the `refine_for_condition` seam
+>    rather than once per configuration.
```
jakewheeler: > if de-selection is truly condition-scoped, that table needs a `condition_id`

This is how I've gone about it. My table does include `condition_id`.

> the way `suppressed_codes` does

Does it? In my mind `suppressed_codes` are conceptually closer to custom codes than TES codes since they are associated directly with a configuration rather than with a condition.

Here's what the overrides UI mockup looks like if that helps at all. This group of override codes is not associated with a condition in any way.

Image

Let me know if I misunderstood.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+
+### Prototyping--Feature B (Suppression)
+
+#### Storage (per-config, condition-scoped)
+
+Scope is configuration + condition. Mirror `custom_codes` but add a condition
+association so the deny-set can be narrowed to the condition being refined.
+
+> [!IMPORTANT]
+> **Per-condition scoping is not free at the current seam--decide this before
```
jakewheeler: I think I need more details on the current condition requirement before I can give you more feedback here.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+> `ProcessedConfiguration`--effectively **option 1 collapsed to config-wide**.
+> That is correct as a "never emit, no matter what" deny-set, but it does not yet
+> distinguish conditions. The identical tension applies to Feature A's projection
+> anti-join (same flattening seam)--see _Feature A design question_ and open
+> question 3.
+
+```sql
+-- suppressed_codes: codes that must NEVER appear in the output for a given
+-- (configuration, condition), regardless of section instructions.
+CREATE TABLE public.suppressed_codes (
```
jakewheeler: How will grouping overridden codes work?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+# app/services/pipeline.py, inside refine_for_condition (~line 298)
+
+eicr_plan = create_eicr_refinement_plan(...)
+
+# NEW: suppression runs FIRST, across all sections, ignoring section
+# instructions. removing suppressed codes before refinement means the
+# refinement pass--which only RETAINS what is present--can never
+# re-introduce them. precedence (suppress > include) falls out of ordering.
+# reuse the spec create_eicr_refinement_plan already loaded (no re-detect).
+suppression_plan = create_suppression_plan(
```
jakewheeler: Fairly minor and could just be a me thing but I think it'd be easier to follow refinement if we tried to use naming that closely resembles the name of the feature. Not a huge deal overall but if someone wants to make changes to "overrides" related code and they see "suppression" instead there is some cognitive load required to translate it in your mind.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+"byte-identical" is the same property applied to the refactor).
+
+### Open questions to resolve before building
+
+1. **Feature A de-selection vs. custom codes** (full detail in the _Feature A
+   note_ above). Because the payload is origin-free and the duplicate check is
+   advisory, a custom/TES duplicate can reach the payload, where a de-selection
+   would strip it as collateral. Decide: (a) enforce the duplicate check on the
+   write paths (`add_custom_code` + CSV confirm) or (b) scope the anti-join to
+   condition-origin codes. Recommended: (a).
```
jakewheeler: What will happen if there is a duplicate code in the payload?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+2. **De-selection vs. suppression overlap.** A code could in principle be both
+   de-selected (removed from want-set) and suppressed (removed from output).
+   These are independent and compose cleanly--de-selection stops retention,
+   suppression removes anything present--but the UI should probably not offer
+   both toggles on the same code in a way that reads as contradictory.
```
jakewheeler: You could be right but I don't know that we need to worry about it as a part of this work. This could be a helpful enhancement later though.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+3. **De-selection granularity & scope** (see the _Feature A design question_
```
jakewheeler: Mentioned in another comment but you can let me know if you think my "exclusion" model will meet what you need.

jakewheeler (commented):
Great work Robert! I left some questions for you about a handful of things.

docs/decisions/0014_2026-07-29_overrides-and-removal.md
fzhao99: I wonder if introducing pre and post language would also be helpful here, something like "pre-pruning" to refer to the curtailing of the want set and "post-pruning" to refer to the curtailing of the output. Though they're talking about the action at different levels of the pipeline process, we're using the  sequentiality of each in the label

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+**Cons**: this is exactly the reinvention we want to avoid--de-selection has no
+reason to touch XML at all, since a code absent from the want-set already
+retains nothing. Rejected.
+
+### Feature B--Suppression
+
+#### B1. Cross-cutting pre-pass that reuses IG rules in subtractive mode (RECOMMENDED)
+
+Add a suppression pass, `suppress_eicr`, that runs in `refine_for_condition`
+(`refiner/app/services/pipeline.py:219`) **before** `refine_eicr`--at the
```
fzhao99: Think I was operating under the assumption that the suppression pass would be a post-processing one. Don't know if I'm wedded to this ordering in particular, but just to have an answer when it inevitably comes up, any particular pros and cons for doing it in the reverse order? Might be easier to say "since omission comes last, it wins in cases where retention refinement and omission refinement conflict"

docs/decisions/0014_2026-07-29_overrides-and-removal.md
fzhao99: Guessing maybe a "not" got dropped here? 😅

docs/decisions/0014_2026-07-29_overrides-and-removal.md
```diff
@@ -0,0 +1,1382 @@
...
+  `convert_config_to_storage_payload`. Nothing downstream changes.
+- **Feature B (Suppression): option B1.** A `suppress_eicr` pre-pass runs before
+  `refine_eicr` across all sections, reusing IG `EntryMatchRule`s in subtractive
+  mode where they exist and falling back to entry-level removal where they do
+  not.
+
+Resolved design decisions (from spike discussion):
+
+- **Suppression scope: configuration + condition, per jurisdiction.** Not
+  jurisdiction-wide. A suppression entry belongs to a `(configuration,
```
fzhao99: Noting that in a hypothetical world we **would** eventually implement this globally like it was before the designs were simplified, but probably not something we should plan against given the circumstances

docs/decisions/0014_2026-07-29_overrides-and-removal.md
fzhao99: Read further down and realized both are happening prior to refinement, so maybe this isn't that useful 😅 

pruning might be a good term though?

docs/decisions/0014_2026-07-29_overrides-and-removal.md
fzhao99: +1

fzhao99 (approved):

PR #1633, opened 2026-08-04
feat: Establish codes management foundation

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR is the first of more to come for the "Codes management" feature. The goal in this PR was to establish a base to work from. I'll highlight below what is complete and what work remains.

### Completed

- [x] Infinite scroll works

### Needs to be completed

- [ ] Individual codes can be included/excluded
- [ ] Code counts are updated when actions are performed on codes
- [ ] Custom codes can be managed from this view
  - [ ] Individual add/edit/deletes
  - [ ] Bulk upload
- [ ] Filters (including search) are implemented
- [ ] Bulk code modification (this impacts all codes in the set, not just what is in view on the page)

## 🧪 How to test

1. Create a configuration
2. Add a few custom codes
3. Add additional code sets
4. Navigate to the temporary page for the new view, `http://localhost:8081/configurations/<config-id>/manage-codes/view`
5. You'll see the new manage codes page. You can add condition code sets and custom codes (individually) from this view as well. You'll need to refresh currently to see the updates
6. (assuming enough codes are attached) Scroll down the page and you should see codes continue to load in as you scroll. A "page" is currently 100 codes, but this is just a temporary value for now


refiner/app/db/configurations/codes/db.py
```diff
@@ -0,0 +1,227 @@
...
+                NULL::uuid AS condition_id,
+                'Custom Code' AS source,
+                c.code,
+                c.display AS description,
+                c.system_id,
+                s.display_name AS system_name,
+                'included' AS status
+            FROM custom_codes c
+            JOIN systems s ON s.id = c.system_id
+            WHERE c.configuration_id = %s
```
fzhao99: Would it be possible to build a dict gradually to use the named parameter pattern instead of the positional one? Think it'd be a pretty significant increase to the legibility of this function

refiner/app/db/configurations/codes/db.py
```diff
@@ -0,0 +1,227 @@
+import base64
```
jakewheeler, line 1: Feel free to leave me feedback on file groupings, by the way. Happy to move things around if a different structure would make more sense.

refiner/app/api/v1/configurations/codes/codes.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+
+    id: UUID
+    condition_id: UUID | None
+    source: str
+    code: str
+    description: str
+    system_id: UUID
+    system_name: str
+    status: Literal["Included", "Excluded"]
```
jakewheeler, line 32: I considered making this a boolean which would also work just fine. I mainly left it this way so we could just directly show `status` text in the UI.

 Let me know if you have any strong feelings one way or the other.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -0,0 +1,331 @@
...
+        configurationId={id}
+        reportable_condition_display_name={display_name}
+        disabled={disabled}
+      />
+    </>
+  );
+}
+
+function CodeInformationBar({ id }: { id: string }) {
+  const { data: codeCounts, isPending, isError } = useGetCodeCounts(id);
```
fzhao99, line 234: don't know if this is meant to happen yet since there's still more to be done with hooking up inclusion / etc, but the codes bar seems to need a re-render in order to update counts when condition codes sets are added

https://github.com/user-attachments/assets/195b98ba-507e-4b1f-b61d-9e7660541384

refiner/app/api/v1/configurations/codes/codes.py
fzhao99: think this is fine! not outside the realm of possibility we change the language / add some other status, and I imagine we may need to do something for the activity log, in which case keeping this on the backend would help

fzhao99 (commented):
woops, realized I started leaving comments in review mode, so submitting a couple

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
jakewheeler: Sorry, that's what I was trying to explain in my testing step 5.

> 5. You'll see the new manage codes page. You can add condition code sets and custom codes (individually) from this view as well. **You'll need to refresh currently to see the updates**

The data fetching is still very much a WIP at this point.

refiner/app/db/configurations/codes/db.py
jakewheeler: Done 🫡  this is better for sure.

fzhao99 (approved):
🎉

PR #1635, opened 2026-08-05
fix: less confusing narrative reconstruction fallback text

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary
Splits out the "narrative reconstruction unavailable" case into distinct messages for "attempted, but failed" as a result of no matches or no configured ruleset. This is to address confusion from WI about the reconstruction happening for no-match sections.

Screenshot 2026-08-06 at 9 54 10 AM

Also applied some extra nesting to our fixtures folder to include the files I needed for testing in dev and localhost

## ✅ Acceptance Criteria

Fixes #1638 

- [ ] Reconstruction fallback uses less confusing retention text

## 🧪 How to test

- Spin up dev and make a COVID configuration with the Problems section set to "Reconstruct"
     - Can also set up a section that has reconstruction (ie Medications Administered) to "Reconstruct" to see the different outputs
- Use the `all_sections_SDDH_nonsense_problems.zip` file to the config, which has all the problem codes swapped to not match
- Observe in the output that the Problems output footnote uses the retention text 

## ℹ️ Additional Information
Do we want this particular case's output footnote to mention reconstruction at all?


refiner/app/services/ecr/narrative/constants.py
```diff
@@ -80,8 +80,11 @@
     SectionOutcome.REFINED_WITH_MATCHES: "Refined; matches found",
     SectionOutcome.REFINED_NARRATIVE_REMOVED: "Refined; narrative removed",
     SectionOutcome.REFINED_NARRATIVE_RECONSTRUCTED: "Refined; narrative reconstructed",
-    SectionOutcome.REFINED_RECONSTRUCT_FALLBACK_RETAINED: (
-        "Refined; reconstruction unavailable, original narrative retained"
+    SectionOutcome.REFINED_RECONSTRUCT_NO_MATCHES_FALLBACK_RETAINED: (
+        "Refined; reconstruction attempted but no matching entries were found, original narrative retained"
+    ),
+    SectionOutcome.REFINED_RECONSTRUCT_UNAVAILABLE_FALLBACK_RETAINED: (
+        "Refined; reconstruction currently unavailable for this section, original narrative retained"
```
jakewheeler, line 87: 💯

jakewheeler (approved):
Seems to be working as expected in my testing. Good job on this!

robertmitchellv (approved):
thanks for making that other ticket bob!

PR #1656, opened 2026-08-10
feat: manage code - single code actions

jakewheeler:
# 🔀 PULL REQUEST

### Completed in this PR

### Previously completed (#1633)

- [x] Infinite scroll works

### Needs to be completed

- [ ] Filters (including search) are implemented
- [ ] Switch colors updated (UI)

## 📺 Demo

Kapture 2026-08-11 at 11 04 55

## 🧪 How to test

1. Run `just migrate local`
2. Create a configuration
3. Navigate to the manage codes view
4. Add `/view` to the end of the URL and navigate to the page (it'll look like this: `http://localhost:8081/configurations/eba668f4-5fc9-4e23-ab5b-f39feeeef133/manage-codes/view`)
5. Try adding and deleting code sets. You'll notice that both the codes list and code counts bar update automatically as you make changes
6. Try adding including/excluding **individual codes** (using the toggle). You will see excluded codes appear in the `configurations_conditions_code_exclusions` table for this configuration
7. Try adding, editing, and deleting custom codes 

> [!IMPORTANT]
> If you see a code in the `configurations_conditions_code_exclusions` table for your configuration but do not see it in the codes list on the page, it is likely that you excluded the code and then removed the code set it was a part of.

### Not including condition ID in exclusions table

I realized that `condition_id` as a part of the exclusions table didn't make much sense because we want to exclude that code for all conditions, not just per condition code set.

What this means is that when we go to generate a `ProcessedConfiguration` using `convert_config_to_storage_payload()` we will need to check if any of the code sets connected to the configuration include that code. If they do, we exclude it from the final configuration output.

If that code was excluded but the configuration doesn't have a connected code set using that code, we don't make any modifications to the final configuration output.

### Helper SQL script

If you want to do a quick check to see which conditions are associated with a code you can use this query

```SQL
SELECT c.*
FROM codes co
JOIN conditions_codes cc ON cc.code_id = co.id
JOIN conditions c ON c.id = cc.condition_id
JOIN tes t ON t.id = c.tes_id
WHERE co.code = '1351801008' -- <-- put your code here
AND t.version = '6.0.0'
```


client/e2e/manage-codes-dev.spec.ts
```diff
@@ -0,0 +1,334 @@
+/**
```
jakewheeler, line 1: There is more work to be done on the UI but it's in a solid enough spot to get some testing started. We can at least check if the UX is working as expected now.

No integration tests yet since the backend is still changing at this point and I imagine will a couple more times 🙈

client/e2e/manage-codes-dev.spec.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    await test.step('Check stats bar', async () => {
+      await expect(page.getByTestId('codes-included-display')).toHaveText(
+        '1 of 2 codes included'
+      );
+      await expect(page.getByText('1 excluded')).toBeVisible();
+    });
+  });
+
+  test('Individual custom codes can be added, edited, and deleted', async ({
```
jakewheeler, line 179: Some of these tests are a bit verbose since our helpers are designed for the current workflow. These should be a bit nicer later once the helpers can be updated for this new workflow.

client/src/pages/Configurations/ManageCodes/CodeSets/ConditionCodeSetListItem.tsx
```diff
@@ -53,6 +55,12 @@ export function ConditionCodeSetListItem({
           await queryClient.invalidateQueries({
             queryKey: getGetConfigurationQueryKey(configurationId),
           });
+          await queryClient.invalidateQueries({
+            queryKey: getGetCodesInfiniteQueryKey(configurationId),
+          });
+          await queryClient.invalidateQueries({
+            queryKey: getGetCodeCountsQueryKey(configurationId),
+          });
```
jakewheeler, line 63: FYI that these queries aren't requested during the current workflow so no additional requests will be made in that scenario.

Image

client/src/pages/Configurations/ManageCodes/CustomCodes/CustomCodeModal.tsx
```diff
@@ -148,6 +150,12 @@ function CustomCodeForm({
             await queryClient.invalidateQueries({
               queryKey: getGetConfigurationQueryKey(configurationId),
             });
+            await queryClient.invalidateQueries({
+              queryKey: getGetCodesInfiniteQueryKey(configurationId),
+            });
+            await queryClient.invalidateQueries({
+              queryKey: getGetCodeCountsQueryKey(configurationId),
+            });
```
jakewheeler, line 158: Same thing here - network requests won't be made during the current workflow.

client/src/pages/Configurations/ManageCodes/CustomCodes/DeleteCustomCodeButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useQueryClient } from '@tanstack/react-query';
+import { useToast } from '../../../../hooks/useToast';
+
+interface DeleteCustomCodeButtonProps {
+  configurationId: string;
+  id: string;
+  code: string;
+}
+
+export function DeleteCustomCodeButton({
```
jakewheeler, line 17: I'm trying not to mess with the current workflow as I implement the new workflow. That's why these separate buttons are being created here.

client/src/pages/Configurations/ManageCodes/CustomCodes/index.tsx
```diff
@@ -88,6 +90,14 @@ export function CustomCodesDetail({
...
                                   getGetConfigurationQueryKey(configurationId),
                               });
+                              await queryClient.invalidateQueries({
+                                queryKey:
+                                  getGetCodesInfiniteQueryKey(configurationId),
+                              });
+                              await queryClient.invalidateQueries({
+                                queryKey:
+                                  getGetCodeCountsQueryKey(configurationId),
+                              });
```
jakewheeler, line 100: Same thing here 😄

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -217,7 +304,19 @@ function AddCodeSetsButton({
   const [isDrawerOpen, setIsDrawerOpen] = useState(false);
   return (
     <>
-      <Button onClick={() => setIsDrawerOpen(true)}>Condition code sets</Button>
+      <Button
+        variant="unstyled"
+        className="border-blue-cool-50! h-8 rounded-md border-2! bg-white px-3 text-sm! whitespace-nowrap hover:cursor-pointer hover:bg-[#eef5f8]!"
```
jakewheeler: `eef5f8` does not appear to be a USWDS color and is not used elsewhere in the app currently. I thought about pulling it out into a variable but am not sure how to label it.

Let me know if you have a suggestion or if I should leave it as is.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -230,53 +329,67 @@ function AddCodeSetsButton({
...
+  const total = codeCounts.data.total_code_count;
+  const barFillPercentage = total > 0 ? (includedCount / total) * 100 : 0;
 
   return (
     <div className="bg-blue-cool-5 border-blue-cool-20! flex min-h-20 w-full flex-col gap-2 border px-10 py-4">
-      {isPending ? (
-        <div className="absolute inset-0 top-10 flex items-center justify-center">
-          <Spinner />
+      <div className="flex flex-row items-center justify-between">
+        <div data-testid="codes-included-display">
```
jakewheeler, line 377: Splitting up the text by spans was causing a lot of trouble checking this in testing 🙃 let me know if a comment here would be helpful.

refiner/app/api/v1/configurations/codes/codes.py
```diff
@@ -89,7 +91,7 @@ async def get_codes(
 
     if not config:
         raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
+            status_code=http_status.HTTP_404_NOT_FOUND,
```
jakewheeler, line 94: I ran into a naming conflict due to using `status` (included/excluded) in the handler.

refiner/app/api/v1/configurations/custom_codes/custom_codes.py
```diff
@@ -48,6 +48,65 @@
 router = APIRouter(prefix="/{configuration_id}/custom-codes")
 
 
+@router.get(
+    "/{id}",
+    response_model=CustomCodeResponse,
+    tags=["configurations"],
+    operation_id="getCustomCode",
+)
+async def get_custom_code(
```
jakewheeler, line 57: This may or may not change later. It's basically being used to drive the custom code "edit" experience for the time being. We request the custom code given its ID and then use that to fill in the fields the user can edit.

refiner/app/db/configurations/codes/db.py
```diff
@@ -152,7 +152,6 @@ async def get_codes_db(
         JOIN systems s ON s.id = c.system_id
         LEFT JOIN configurations_conditions_code_exclusions e
             ON e.configuration_id = cfgc.configuration_id
-            AND e.condition_id = cfgc.condition_id
```
jakewheeler, line 155: See my note about `condition_id` in the PR summary.

refiner/app/db/configurations/codes/db.py
```diff
@@ -182,6 +181,56 @@ async def get_codes_db(
     return rows, next_cursor
 
 
+async def set_codes_status_db(
+    configuration_id: UUID,
+    code_ids: list[UUID],
+    status: Literal["included", "excluded"],
```
jakewheeler, line 187: Bulk including/excluding doesn't work in the UI yet but I designed this from the start to be able to operate on a list of code IDs.

refiner/migrations/20260810165940_remove_condition_id_from_exclusions.sql
```diff
@@ -0,0 +1,30 @@
+-- migrate:up
+
+DROP TABLE configurations_conditions_code_exclusions;
```
jakewheeler, line 3: As mentioned in my summary, I got rid of `condition_id` as a requirement. Since we aren't actually collecting exclusions yet I am simply dropping and recreating the table. This was by far the easiest way to make the change 😂

client/src/pages/Configurations/ManageCodes/CustomCodes/EditCustomCodeButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+        aria-label="Edit custom code"
+        disabled={disabled}
+        onClick={async () => {
+          await refetch();
+          setIsOpen(true);
+        }}
+      >
+        Edit
+      </Button>
+      <CustomCodeModal
```
jakewheeler, line 37: Some of our components are tailored to the current config building experience and I'm doing my best to not go too far with refactoring them right now.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
fzhao99: I think we should just use `blue-cool-5` since its the color in the design system that gets the closest. If I were to hazard a guess, think this shade probably got generated by Claude

refiner/app/db/configurations/codes/db.py
```diff
@@ -182,6 +181,56 @@ async def get_codes_db(
...
+            DELETE FROM configurations_conditions_code_exclusions
+            WHERE configuration_id = %(configuration_id)s
+              AND code_id = ANY(%(code_ids)s)
+            RETURNING code_id
+        """
+    else:
+        query = """
+            INSERT INTO configurations_conditions_code_exclusions (configuration_id, code_id)
+            SELECT %(configuration_id)s, UNNEST(%(code_ids)s::uuid[])
+            ON CONFLICT DO NOTHING
```
fzhao99, line 217: nice!

fzhao99 (approved):
🎉

PR #1657, opened 2026-08-11
chore: add source information to TES seeding

### LINKED ISSUE #1655: [TASK] Update TES seeding to pull in condition source
As part of code management, information about condition source needs to be added to the seeding flow

- [ ] Source information is added to the TES seeding script in a format that supports code management
- [ ] No regressions are observed in the application seeding flow / operations

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Adds source information into the `conditions_codes` table as part of the seeding run in order to support code management work. 

## 🔗 Related Issue
Fixes #1655 
- #1655

## ✅ Acceptance Criteria
- [ ] Source information is added to the TES seeding script in a format that supports code management
- [ ] No regressions are observed in the application seeding flow / operations

## 🧪 How to test
- Start on main with a fresh `just db refresh` migrated db. Take note of the child_rsg and non_child_rsg seeding numbers
- Switch to this branch and run a local migrate. Ensure migration works
- Run the new seed script and verify no errors occur. Note for the output steps
     - The inserted code numbers are the same (89146 new rows, but no new inserts)  
     - The child RSG numbers match (1641 for a full seed, 558 for a local-only seed across the two most recent TES versions) 
     - The non_child_rsg relationships have increased to account for the relaxed source uniqueness constraint (583,397 --> 644,762 for the two most recent seeds)
- Run a down migration to make sure that works too


refiner/app/db/conditions/db.py
```diff
@@ -179,14 +179,12 @@ async def get_condition_by_id_db(
                 c.coverage_level_date
             FROM conditions c
             JOIN tes t ON t.id = c.tes_id
-            WHERE c.id = %s
+            WHERE c.id = %(id)s
```
fzhao99, line 182: opted to fix these guys to the new pattern since I was doing a sweep of these files anyways

refiner/app/db/pool.py
```diff
@@ -40,6 +42,7 @@ def __init__(
             self.connection_url,
             min_size=min_size,
             max_size=max_size,
+            timeout=timeout,
```
fzhao99, line 45: I was getting integration run failures locally because of seeding timeouts, so bumping this up.

refiner/migrations/20260810163437_add_tes_source_to_seeding.sql
```diff
@@ -0,0 +1,31 @@
+-- migrate:up
+DROP INDEX IF EXISTS codes_upsert_constraint_idx;
+
+ALTER TABLE conditions_codes ADD COLUMN source_url TEXT DEFAULT '' NOT NULL;
+ALTER TABLE conditions_codes ADD COLUMN source_name TEXT DEFAULT '';
+
+CREATE UNIQUE INDEX IF NOT EXISTS code_system_source_constraint_idx
+    ON conditions_codes (condition_id, code_id, source_url);
```
fzhao99, line 11: opted to have both `source_url` and `source_name` to the table, with the former being the column factored into the primary key, since we needed an ID-like string to dedupe the relationships. From what I understand, url is a good candidate for it since it uniquely identifies the valueset that the code in question came from, but if folks have concerns / other thoughts, happy to look into a replacement

refiner/scripts/seeding/load_static_data.py
```diff
@@ -159,100 +168,158 @@ def _upsert_tes_data(
     return version_to_tes_id
 
 
+@dataclass
+class BuildCodeContext:
```
fzhao99, line 176: the parent `_build_code` function was getting a little unwieldy, so I opted to refactor things into child / parent steps with a shared context.

refiner/migrations/20260810163313_rename_conditions_codes_pkey.sql
fzhao99, line 1: realized this pkey was misnamed after the recent work with this table, so added this in to fix things

refiner/app/db/tes/db.py
```diff
@@ -178,62 +178,63 @@ async def get_tes_update_condition_diff_db(
...
+            cond.canonical_url,
+            cond.display_name AS condition_name,
+            cond.tes_id,
+            c.id AS code_id,
+            c.code,
+            c.display AS code_name,
+            s.display_name AS system_name
+        FROM conditions cond
+        JOIN (
+            SELECT DISTINCT condition_id, code_id
```
fzhao99, line 192: since there's a possibility of the same condition ID <> code ID showing up in this, we only want to grab the distinct ones in cases where we don't care about the source information. Refactored these into a central CTE to make this a little easier to reason about, but lmk if anyone is concerned since these queries are getting pretty gnarly

refiner/app/db/pool.py
jakewheeler: Weird. I can't seem to get this to happen on my machine when I go back to using the default value. Do you think it could just be due to your local machine running too slowly?

Is [this](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/tests/integration/conftest.py#L561) the final message you'd see when running `pytest tests/integration -s` before things start failing?

refiner/migrations/20260810163437_add_tes_source_to_seeding.sql
jakewheeler: If the same source is a part of two different ValueSets (different `source_url` values, same `source_name` values), what is the best way to determine that they are the same source?

For example, these codes have these `source_urls`:

- `G83.13` -> `https://tes.tools.aimsplatform.org/api/fhir/ValueSet/rs-grouper-47000000`
- `G83.23` -> `https://tes.tools.aimsplatform.org/api/fhir/ValueSet/rs-grouper-698293005`

both point to `Acute Flaccid Myelitis Reporting Specification Grouper` in TES 6.0.0.

This is the query I'm testing things out with if it's helpful.
```sql
select c.code, cc.* from conditions_codes cc
join codes c on cc.code_id = c.id
join conditions cond on cc.condition_id = cond.id
join tes t on cond.tes_id = t.id
where cc.source_name = 'Acute Flaccid Myelitis Reporting Specification Grouper'
and t.version = '6.0.0'
```

refiner/app/db/conditions/db.py
```diff
@@ -291,7 +289,7 @@ async def get_conditions_by_child_rsg_snomed_codes_db(
     db: AsyncDatabaseConnection, codes: list[str]
 ) -> list[DbCondition]:
     """
-    Given a list of RC SNOMED codes, find their assocaited CG rows.
+    Given a list of RC SNOMED codes, find their associated CG rows.
```
jakewheeler, line 292: Had to check if this typo was my doing and it actually wasn't for once 😭

refiner/app/db/conditions/db.py
jakewheeler: Thanks!

refiner/app/db/pool.py
```diff
@@ -31,6 +32,7 @@ def __init__(
             db_password (str): The PostgreSQL password.
             min_size (int, optional): Minimum number of connections to maintain in the pool. Defaults to 1.
             max_size (int, optional): Maximum number of connections allowed in the pool. Defaults to 10.
+            timeout (int, optional): Amount of time to wait for DB to be ready, defaults to a minute to allow for db migration/seeding.
```
jakewheeler, line 35: Should we keep the default value for non-testing situations (appears to be 30 seconds according to the [docs](https://www.psycopg.org/psycopg3/docs/api/pool.html#psycopg_pool.ConnectionPool)) and set it to 60 in `integration/conftest.py`'s [`db_pool` fixture?](https://github.com/CDCgov/dibbs-ecr-refiner/blob/main/refiner/tests/integration/conftest.py#L402)

refiner/app/db/pool.py
fzhao99: I'm like 90% sure it's a local resources problem: getting messages about the db connection pool being starved for timeouts at the 30 second mark. Will change the default back to the 30 / pass in the longer version for the integration conftest like you've suggested though.

fzhao99:
closing in favor of #1661 after decisions to make changes to the architecture

PR #1658, opened 2026-08-12
feat: Manage codes - UI filter options

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

> [!IMPORTANT]
> This PR only collects the data for the filters and adds the components to the page. Filtering the data from the API will come in the next PR!

This PR adds the `Code system`, `Source`, and `Status` filters to the manage codes page. The user is able to selected options from each of these filters. Every filter has its own "Clear selection" button which will deselect all options in that filter.

These filters are all stored in client-side state when selections are made. The next PR will be able to use this client-side state and pass it to the API in order to receive back filtered, paginated data to display on the page.

See the demo below.

### Completed in this PR

- [x] Code system filter component exists on page
- [x] Source filter component exists on page
- [x] Status filter component exists on page
- [x] Filter state is collected client side
- [x] e2e tests for filter UX

### Previously completed

#1633 
- [x] Infinite scroll works

#1656 

### Needs to be completed

- [ ] Paginated data can be filtered by filter selections
- [ ] Paginated data can be filtered by search bar
- [ ] Switch colors updated (UI)
- [ ] Checkbox colors updated (UI)
- [ ] Performance optimizations

## 📺 Demo

This gif provides a brief demo of what using the filter functionality looks like. You'll notice that when I remove a condition being used for filtering that it automatically deselects that filter.

Kapture 2026-08-14 at 12 11 16

## 🧪 How to test

1. Create a configuration
2. Navigate to the manage codes page
3. In your browser, add `/view` to the end of the page URL (example: `http://localhost:8081/configurations/cc8eff9e-9628-49bb-9aec-7300f5aa1095/manage-codes/view`)
4. Try using the filters. Ensure they behave as you expect and that the data appears to be correct


client/src/pages/Configurations/ManageCodes/CodeSets/ConditionCodeSetListItem.tsx
```diff
@@ -61,6 +62,9 @@ export function ConditionCodeSetListItem({
           await queryClient.invalidateQueries({
             queryKey: getGetCodeCountsQueryKey(configurationId),
           });
+          await queryClient.invalidateQueries({
+            queryKey: getGetCodeFiltersQueryKey(configurationId),
+          });
```
jakewheeler, line 67: This is needed because the associated code sets are what build up the "source" filter. So if one is removed or added, the filter options need to be updated.

client/src/pages/Configurations/ManageCodes/CustomCodes/CustomCodeModal.tsx
```diff
@@ -156,6 +157,9 @@ function CustomCodeForm({
             await queryClient.invalidateQueries({
               queryKey: getGetCodeCountsQueryKey(configurationId),
             });
+            await queryClient.invalidateQueries({
```
jakewheeler, line 160: Same idea as my other comment.

client/src/pages/Configurations/ManageCodes/Filters.tsx
```diff
@@ -0,0 +1,212 @@
...
+  testId,
+  onChange,
+}: FilterComboboxProps<T>) {
+  const handleChange = (values: (T | typeof CLEAR_OPTION)[]) => {
+    if (values.some((v) => v.id === CLEAR_OPTION.id)) {
+      onChange([]);
+    } else {
+      onChange(values as T[]);
+    }
+  };
```
jakewheeler, line 41: I initially didn't want `Clear selection` to be an option in the listbox. However, after a lot of fighting with it, I finally settled on this being a fair compromise. Getting the styling we're after without introducing a11y issues was a struggle.

client/src/pages/Configurations/ManageCodes/Filters.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  return (
+    <Combobox multiple value={selected} onChange={handleChange}>
```
jakewheeler, line 44: Mostly just an FYI: this combobox component is nothing like our existing `@components/Combobox` which is why I needed to go 100% custom 🥲

client/src/pages/Configurations/ManageCodes/Filters.tsx
```diff
@@ -0,0 +1,212 @@
...
+      <div className="has-[input:focus-within]:outline-blue-40v relative has-[input:focus-within]:outline-5 has-[input:focus-within]:-outline-offset-2">
+        <ComboboxInput aria-label={label} className="sr-only" />
+        <ComboboxButton
+          data-testid={`${testId}-button`}
```
jakewheeler, line 48: Unfortunately more test IDs being introduced. There simply was not a good way to reliably select all of these different components in my e2e tests. Using `getByTestId('custom-system-button')` for example solved a lot of my headaches.

client/src/pages/Configurations/ManageCodes/useFilterState.ts
```diff
@@ -0,0 +1,79 @@
...
+      pruned.sources.length !== filters.sources.length ||
+      pruned.statuses.length !== filters.statuses.length;
+
+    if (changed) {
+      // eslint-disable-next-line react-hooks/set-state-in-effect
+      setFilters(pruned);
+    }
+  }, [availableFilters]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  return { filters: activeFilters, setFilters };
```
jakewheeler, line 55: This is essentially how we keep all of the filters in sync as the user makes changes to the data. For example, if a code set is associated with the config and is being used to filter but is then disassociated, this will ensure that filter is automatically removed. Same for if a custom code is deleted.

There are likely better ways to do this. I experimented with using a `FilterContext` that allows us update the filters when things change but it ended up being a lot of extra work to manage. Having this "filter watcher" works without any additional effort.

I'm open to suggestions. I may revisit this in the future if time permits.

jakewheeler: FYI I incorporated a `search` filter into this in #1679 and it works well. I don't mind this pattern too much after working with it some more but still feel free to let me know your thoughts.

fzhao99: this looks pretty clean with me! good job for covering the disassociation case too

fzhao99 (approved):
🎉

PR #1661, opened 2026-08-14
chore: extend context grouper table to include all valuesets

### LINKED ISSUE #1655: [TASK] Update TES seeding to pull in condition source

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Extends the `condition_context_groupers` table to store complete valueset information as a first-class citizen. This is to support related operations in code management

## 🔗 Related Issue

Fixes #1655 
- #1655 

## ✅ Acceptance Criteria

- [ ] Source information is added to the TES seeding script in a format that supports code management
- [ ] No regressions are observed in the application seeding flow / operations

## 🧪 How to test
- Run a refresh on main, switch to this branch, and run a migrate
- Run a db seed
- Notice that codes total information hasn't changed from previous seed steps
- Verify every condition has at least one corresponding valueset 

```sql
SELECT *
FROM conditions c
WHERE NOT EXISTS (
    SELECT 1 
    FROM valuesets v 
    WHERE v.condition_id = c.id
);
```
- Spot check that the valueset code totals from both the `valueset` table and the `conditions_codes_temp` table match by
     - find a row in the valueset table to check, noting the valueset ID, condition ID, and code total 
     - run the below query with the valueset / condition ID's you noted and ensure the returned query rows matches the code total you referenced

```sql
SELECT * FROM conditions_codes_temp 
WHERE valueset_id = '<YOUR VALUESET ID>' and condition_id='<YOUR CONDITION ID>';
```


refiner/app/db/conditions/db.py
```diff
@@ -511,14 +511,14 @@ async def get_context_groupers_by_condition_id_db(
         SELECT
             id,
             condition_id,
-            name,
+            display_name,
```
fzhao99, line 527: we seem to be using `name`, `display` and `display_name` in various places, so going to try to unify them in a followup. Think `display_name` is the most unambiguous but please weigh in if folks think another value is better

refiner/migrations/20260813142528_rename_conditions_codes_pkey.sql
fzhao99, line 1: after we changed the name of this table, we never went back and renamed this primary key, so adding this in here

refiner/app/db/tes/db.py
```diff
@@ -108,7 +108,7 @@ async def _get_baseline_tes_update_condition_diff_db(
     cond_url: str,
 ) -> ConditionDiffExportData:
     query = """
-        SELECT
+        SELECT DISTINCT
```
fzhao99, line 111: we need these now since there's a possibility the same code <> system but with different valuesets get returned when querying the outputs for the TES updates. Since we don't expose valuesets, we need to select distinct in order to not have duplicate codes

refiner/scripts/seeding/load_static_data.py
```diff
@@ -159,100 +181,215 @@ def _upsert_tes_data(
...
-        condition_child_rsg_snomed_codes: set[DbCodedConcept] = set()
-        condition_non_child_rsg_snomed_codes: set[DbCodedConcept] = set()
+@dataclass
+class BuildCodeContext:
+    """
+    Context for code and code relationships that .
+    """
+
+    db_ids: SystemOidToDbIdMap
+    unique_codes: dict[tuple[UUID, str], CodeRow] = field(default_factory=dict)
```
fzhao99, line 195: hopefully this new object makes the code building flow a bit more readable. Open to any suggestions on naming, style, etc.

refiner/tests/integration/conftest.py
```diff
@@ -616,6 +607,15 @@ def setup(request):
...
     print("🏃‍♀️ Database is ready!")
 
+    print("⚙️ Restarting refiner server")
+    _restart_service(service="server", compose_path=path)
+
+    print("⚙️ Restarting refiner Lambda")
+    _restart_service(service="lambda", compose_path=path)
+
+    refiner_service.wait_for("http://0.0.0.0:8080/api/healthcheck")
+    print("✨ Message refiner services ready to test!")
```
fzhao99: as part of debugging the integration tests, I had a (wrong) hypothesis that moving the lambda / app containers after the db seeding might help. It ended up being something else, but think it makes a bit more sense (and may be a bit more performant?) to only start the app up once the DB is ready.

If I'm missing something happy to undo

refiner/app/db/codes/db.py
```diff
@@ -15,7 +15,7 @@ async def get_rsg_codes_by_condition_id_db(
 
     query = """
         SELECT c.display, c.code, tes.version, c.system_id
-        FROM conditions_codes as cc
+        FROM conditions_codes_temp as cc
```
fzhao99, line 18: we made a temp table to allow for us to extend the `code_id` <> `condition_id` constraint to include `valueset_id`, which have broken Postgres uniqueness constraints otherwise due to the null rows within the new column we needed to add to the junction table. Think eventually once this deploy makes its way to prod, we can add another migration to rename the table // do away with these reads

fzhao99:
will need to touch up some of the codes management query code to get the tests to pass since the relevant portions are expecting a different DB schema shape on the edited table. Will put in draft until those get in

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -201,10 +201,16 @@ function CodesTable({ id, disabled, filters }: CodesTableProps) {
...
+              <th scope="col" className="w-[10%]">
+                Code no.
+              </th>
+              <th scope="col" className="w-[5%]">
+                System
+              </th>
+              <th scope="col" className="w-[40%]">
+                Description
+              </th>
+              <th scope="col" className="w-[35%]">
```
fzhao99, line 213: thought I'd tweak the width of the columns to better distribute the width across the relative length of the columns. Lmk if anyone foresees issues with this

before:
Image

after:
Image

jakewheeler: Raw percentages won't scale well as we shrink the window down unfortunately. The table looks really crowded on my typical display size.
image

I plan to revisit some of these design details later on so totally up to you if you'd prefer to revert it for now or try to dial it in some more.

client/src/pages/Configurations/ManageCodes/Filters.tsx
```diff
@@ -97,7 +98,7 @@ function FilterCombobox<T extends FilterOption>({
                   <span className="text-md flex-1 text-gray-800">
                     {option.label}
                   </span>
-                  {option.count !== undefined && (
+                  {option.count != null && (
```
jakewheeler: `FilterOption`'s `count` is typed as `count?: number;` so the prior code was correct. Looks like we also lost a `=` here.
```suggestion
                  {option.count !== undefined && (
```

client/src/api/schemas/codeResponse.ts
```diff
@@ -6,7 +6,7 @@ import type { CodeResponseStatus } from './codeResponseStatus';
 export interface CodeResponse {
   id: string;
   condition_id: string | null;
-  source: string;
+  source: string[];
```
jakewheeler, line 9: Nice 👀

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -277,7 +285,7 @@ interface SourceCellProps {
 }
 
 function SourceCell({ configurationId, code }: SourceCellProps) {
-  if (!code.is_custom) return code.source;
+  if (!code.is_custom) return code.source.join(', ');
```
jakewheeler, line 288: 🫡

refiner/app/db/conditions/db.py
jakewheeler: Agreed with `display_name` being the best option 👍

refiner/app/db/configurations/codes/db.py
```diff
@@ -340,56 +350,75 @@ async def get_all_filter_options_db(
...
-            con.display_name || ' CG' AS source,
-            CASE WHEN e.code_id IS NULL THEN 'included' ELSE 'excluded' END AS status
+            v.id AS source_id,
+            v.display_name AS source_name,
+            CASE WHEN e.code_id IS NOT NULL THEN 'excluded' ELSE 'included' END AS status
         FROM configurations_conditions cfgc
         JOIN conditions con ON con.id = cfgc.condition_id
-        JOIN conditions_codes cc ON cc.condition_id = cfgc.condition_id
+        JOIN conditions_codes_temp cc ON cc.condition_id = con.id
+        INNER JOIN valuesets v ON v.id = cc.valueset_id AND v.condition_id = con.id
```
jakewheeler, line 364: 🧠

refiner/app/db/configurations/codes/db.py
```diff
@@ -445,16 +474,17 @@ async def get_code_count_metadata_db(
 
     query = """
     SELECT
-        COUNT(*) AS total_code_count,
-        COUNT(*) FILTER (WHERE e.code_id IS NOT NULL) AS excluded_code_count,
+        COUNT(DISTINCT c.id) AS total_code_count,
+        COUNT(DISTINCT e.code_id) AS excluded_code_count,
```
jakewheeler, line 478: Nice!

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
fzhao99: updated this to give a bit more space to the smaller columns

Screenshot 2026-08-20 at 3 46 22 PM
Screenshot 2026-08-20 at 3 46 16 PM

refiner/migrations/20260813142548_generalize_context_grouper_table.sql
```diff
@@ -0,0 +1,41 @@
...
+    CONSTRAINT fk_code_id_fkey
+        FOREIGN KEY(code_id)
+        REFERENCES codes(id)
+        ON DELETE CASCADE
+);
+
+
+-- migrate:down
+DROP TABLE conditions_codes_temp;
+
```
jakewheeler: I think we need to restore the primary key constraint.
```suggestion
ALTER TABLE conditions_codes ADD CONSTRAINT conditions_codes_pkey
    PRIMARY KEY (condition_id, code_id);
```
I was running into an issue trying to roll back `20260813142528_rename_conditions_codes_pkey.sql`. This seems to fix it.

fzhao99:
@jakewheeler I think I'm seeing a roughly comparable seed script completion time to what's on main. Fftm tomorrow if it looks good to you

jakewheeler (approved):
This all seems to be working well. Performance is much better! Thanks 🫡

PR #1677, opened 2026-08-17
chore: fix integration test config

fzhao99:
# 🔀 PULL REQUEST

## 💡 Summary

Fix to the integration test setup that seems to have broken when we refactored the env variables. It seems that calling the DB config via the cache breaks the creation of the DB pool connection. Moving things to the top of the module seems to work.

## 🔗 Related Issue

Fixes local integration test runs

## ✅ Acceptance Criteria

- Local runs of the integration test suite pass 

## 🧪 How to test
- Run the integration tests locally via pytest or just
- Notice all tests pass

## ℹ️ Additional Information
Any concerns with the way we're maintaining the env import?


jakewheeler (approved):
👍

PR #1679, opened 2026-08-17
feat: Manage codes - search

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR makes the search component functional on the codes management page. Typing into the search bar will fire off a request to the server to return newly filtered results. The search bar will work in tandem with the filter options to allow users to really narrow down their results.

Here's an example where I'm searching while also using a code system filter (`SNOMED`) and a status filter (`Included`).
image

### Completed in this PR

- [x] Paginated data can be filtered by search bar

### Previously completed

#1662 

- [x] Paginated data can be filtered by filter selections

#1658 

- [x] Code system filter component exists on page
- [x] Source filter component exists on page
- [x] Status filter component exists on page
- [x] Filter state is collected client side
- [x] e2e tests for filter UX

#1633 
- [x] Temporary codes management page established at `http://localhost:8081/configurations/<config-id>/manage-codes/view`
- [x] Adds migration script for the new `configurations_conditions_code_exclusions`, which is used to manage codes that have been excluded from a condition code set for a configuration
- [x] Added a stats bar with code count information
- [x] Added a codes table with a mix of custom codes and condition code set codes
  - [x] Custom codes come first in the list
- [x] Infinite scroll works
  - [x] Paginated querying implemented to support this
  - [x] Infinite scroll component used
- [x] Custom codes can be added from this view (temporary, legacy styling is used)
- [x] Condition code sets can be managed from this view

#1656 
- [x] Exclusions table / migration script adjusted to not factor in condition ID (see note on this below)
- [x] Individual codes can be included/excluded
- [x] Code list refreshes when actions are performed on codes/sets
- [x] Code counts refresh when actions are performed on codes/sets
  - [x] Bar fill adjusts when codes are included/excluded
- [x] Individual custom codes can be added/edited/deleted
  - [x] New route added to fetch a custom code by ID
  - [x] New button/menu styling is used
  - [x] Custom codes `Source` column uses new styling
- [x] e2e tests for the new view have been started

### Needs to be completed

- [ ] Table header is sticky
- [ ] Excluded codes are included in the configuration cloning process
- [ ] Excluded codes are included in the refinement process (`convert_config_to_storage_payload()`, take a look at #1641)
- [ ] This view replaces the existing legacy code management view
- [ ] "Source" column reflects the RSG or ACG
  - [ ] This data is pulled into the DB via the seeding script
- [ ] Condition codes in table are de-duped
  - [ ] Codes with the same number show multiple sources
[...]

PR #1662, opened 2026-08-14
feat: Manage codes - server-side codes filtering

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR adds the server-side component for filtering data on the manage codes page. Modifying filters on the page will result in the table of codes being updated as changes occur.

See the demo below.

### Completed in this PR

- [x] Paginated data can be filtered by filter selections

### Previously completed

#1658 

- [x] Code system filter component exists on page
- [x] Source filter component exists on page
- [x] Status filter component exists on page
- [x] Filter state is collected client side
- [x] e2e tests for filter UX

#1633 
- [x] Temporary codes management page established at `http://localhost:8081/configurations/<config-id>/manage-codes/view`
- [x] Adds migration script for the new `configurations_conditions_code_exclusions`, which is used to manage codes that have been excluded from a condition code set for a configuration
- [x] Added a stats bar with code count information
- [x] Added a codes table with a mix of custom codes and condition code set codes
  - [x] Custom codes come first in the list
- [x] Infinite scroll works
  - [x] Paginated querying implemented to support this
  - [x] Infinite scroll component used
- [x] Custom codes can be added from this view (temporary, legacy styling is used)
- [x] Condition code sets can be managed from this view

#1656 
- [x] Exclusions table / migration script adjusted to not factor in condition ID (see note on this below)
- [x] Individual codes can be included/excluded
- [x] Code list refreshes when actions are performed on codes/sets
- [x] Code counts refresh when actions are performed on codes/sets
  - [x] Bar fill adjusts when codes are included/excluded
- [x] Individual custom codes can be added/edited/deleted
  - [x] New route added to fetch a custom code by ID
  - [x] New button/menu styling is used
  - [x] Custom codes `Source` column uses new styling
- [x] e2e tests for the new view have been started

### Needs to be completed

- [ ] Table header is sticky
- [ ] Excluded codes are included in the configuration cloning process
- [ ] Excluded codes are included in the refinement process (`convert_config_to_storage_payload()`, take a look at #1641)
- [ ] This view replaces the existing legacy code management view
- [ ] "Source" column reflects the RSG or ACG
  - [ ] This data is pulled into the DB via the seeding script
- [ ] Condition codes in table are de-duped
  - [ ] Codes with the same number show multiple sources
- [ ] Codes can be included/excluded in bulk
- [ ] Custom code bulk uploads can be managed from this view
- [ ] Paginated data can be filtered by search bar
- [ ] Switch colors updated (UI)
[...]

PR #1680, opened 2026-08-18
feat: Manage codes - bulk updates

jakewheeler:
# 🔀 PULL REQUEST

> [!IMPORTANT]
> This does not include the "select every single code in my filter whether it's loaded on the page or not" workflow. That will come later.

## 💡 Summary

This PR updates the code management page to make bulk update actions available to users. Users are able to include codes, exclude codes, and delete custom codes using the "control panel" that pops up at the bottom of the screen when selecting codes in the UI.

Please see the demo section below to understand how this all works.

This PR specifically adds the following:

- [x] Codes can be included/excluded in bulk (by code IDs available in the UI)
- [x] Custom codes can be deleted in bulk
- [x] Remove the switch to toggle included/excluded status individually (only control panel is available)

### Previously completed

#1679 

- [x] Paginated data can be filtered by search bar

#1662 

- [x] Paginated data can be filtered by filter selections

#1658 

- [x] Code system filter component exists on page
- [x] Source filter component exists on page
- [x] Status filter component exists on page
- [x] Filter state is collected client side
- [x] e2e tests for filter UX

#1633 
- [x] Infinite scroll works

#1656 

### Needs to be completed

- [ ] "select all" allows users to perform actions (including, excluding) on all filtered codes
- [ ] Table header is sticky (#1684)
- [ ] "Source" column reflects the RSG or ACG (#1661)
- [ ] Checkbox colors updated (UI)
- [ ] Performance optimizations

## 📺 Demo

Kapture 2026-08-20 at 13 39 57

## 🧪 How to test

1. Create a configuration
2. Navigate to the manage codes page
3. In your browser, add `/view` to the end of the page URL (example: `http://localhost:8081/configurations/cc8eff9e-9628-49bb-9aec-7300f5aa1095/manage-codes/view`)
4. Try various combinations of including/excluding/deleting using codes and custom codes by using the control panel at the bottom of the screen when codes are selected

Some things to notice from my demo GIF above:
- You will only see include and exclude buttons when 0 custom codes are selected
- You will see all 3 options when at least 1 custom code is selected
- Statuses can be bulk updated to the same status without issue (include -> include, for example)
- Hitting "Exclude" with no custom codes selected will exclude the codes immediately
  - If a custom code is included with code set codes you will see a warning modal
  - If only a custom code is included, the warning modal will appear slightly differently


client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -164,19 +157,37 @@ function CodesTable({ id, disabled, filters }: CodesTableProps) {
   const [selectedIds, setSelectedIds] = useState<Set<string>>(new Set());
   const [isSourceModalOpen, setIsSourceModalOpen] = useState(false);
 
+  useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
+    setSelectedIds(new Set());
+  }, [filters]);
+
```
jakewheeler: Selections will be automatically reset when filters change.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -255,13 +266,7 @@ function CodesTable({ id, disabled, filters }: CodesTableProps) {
...
                   <SourceCell configurationId={id} code={code} />
                 </td>
-                <td>
-                  <IncludeSwitch
-                    configurationId={id}
-                    code={code}
-                    disabled={disabled}
-                  />
-                </td>
+                <td>{code.status}</td>
```
jakewheeler, line 277: This was a slight design tweak I included here.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
```diff
@@ -296,60 +299,6 @@ function SourceCell({ configurationId, code }: SourceCellProps) {
   );
 }
 
-interface IncludeSwitchProps {
-  configurationId: string;
-  code: CodeResponse;
-  disabled: boolean;
-}
-function IncludeSwitch({
```
jakewheeler, line 312: There is no more individual toggle. All actions will now go through the control panel.

refiner/app/api/v1/configurations/custom_codes/custom_codes.py
```diff
@@ -561,16 +561,18 @@ async def delete_custom_code(
             detail=f"Failed to find custom code to delete with ID: {id}",
         )
 
-    deleted_code = await delete_custom_code_db(
-        config=config, id=custom_code.id, user_id=user.id, db=db
+    deleted_codes = await delete_custom_codes_db(
```
jakewheeler, line 564: This function was updated to take a list of custom code IDs to delete.

refiner/app/api/v1/configurations/custom_codes/custom_codes.py
```diff
@@ -584,6 +586,91 @@ async def delete_custom_code(
...
 
+class BulkDeleteCustomCodesInput(BaseModel):
+    """
+    Input model for a bulk custom codes deletion request.
+    """
+
+    ids: list[UUID]
+
+
+@router.post(
```
jakewheeler, line 597: I intentionally opted to leave the more conventional single ID `DELETE `alone. It seems to be generally recommended to handle bulk deletes like this in a REST API.

client/src/pages/Configurations/ManageCodes/ControlPanel.tsx
```diff
@@ -0,0 +1,345 @@
...
+} from '@components/Modal';
+import { useState } from 'react';
+
+interface ControlPanelProps {
+  configurationId: string;
+  selectedCodeIds: Set<string>;
+  selectedCustomCodes: CodeResponse[];
+  clearSelections: () => void;
+}
+export function ControlPanel({
```
jakewheeler, line 29: It would be good to make the toast component not as long in another PR so it doesn't overlap this component. You'll notice in my demo it's constantly getting in the way when I'm trying to make updates.

client/e2e/manage-codes-dev.spec.ts
```diff
@@ -26,6 +26,445 @@ async function goToManageCodesDevPage(
...
+        page.getByRole('heading', { name: 'Configurations', level: 1 })
+      ).toBeVisible();
+      await page.getByRole('link', { name: 'Anotia' }).click();
+      await expect(
+        page.getByRole('heading', { name: 'Customize eICR sections' })
+      ).toBeVisible();
+      await goToManageCodesDevPage(page, configurationPage);
+      await expect(
+        page.getByRole('heading', { name: 'Manage codes', level: 2 })
+      ).toBeVisible();
```
fzhao99, line 99: not something we have to do here, but would be good to maybe factor these into POM operations at some point

client/src/pages/Configurations/ManageCodes/ControlPanel.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const hasCustomCodesSelected = selectedCustomCodes.length > 0;
+  return (
+    <>
+      <ExclusionWarningModal
+        isOpen={isOpen}
+        onClose={() => setIsOpen(false)}
+        customCodeCount={customCodeIds.size}
+        totalCodeCount={selectedCodeIds.size}
+        updateCodesToExcluded={() => updateSelectedCodesStatus('Excluded')}
+      />
```
fzhao99, line 94: nice!

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
fzhao99:
```suggestion
const [prevFilters, setPrevFilters] = useState(filters);
  if (filters !== prevFilters) {
    setPrevFilters(filters);
    setSelectedIds(new Set());
  }
```

thoughts on refactoring this per [the suggestions in the React docs](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)?

fzhao99: alternatively, [if there's a key variable that we could pass into the control panel](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) that's derived from the filters prop, that would be another (simpler) way of getting the prop to re-render on change. 

It's not immediately obvious what that might be, but perhaps you have some ideas. Since it seems in your other PR we're using configuration ID in the hook for filters, so maybe that'd work?

refiner/app/api/v1/configurations/custom_codes/custom_codes.py
fzhao99: nice!

fzhao99 (commented):
one suggestion about state updates interactions in the control panel, but otherwise lgtm! exciting to see this working now :)

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
jakewheeler: At one point I was doing this:
```ts
<CodesTable
  key={JSON.stringify(filters)}
  id={id}
  disabled={disabled}
  filters={filters}
/>
```
Which also seemed to do the trick. I wasn't sure which way was worse 🙈 

I could stringify within `useFilterState` and expose it via `filterKey` or something like that. Do you think that would be better?

client/e2e/manage-codes-dev.spec.ts
jakewheeler: Agreed. I'm still doing a lot of copy/pasting at this point but I think the UI is getting to a stable enough place that we could get a POM together. I won't do it here but do plan on doing it soonish.

client/src/pages/Configurations/ManageCodes/ManageCodesDev.tsx
fzhao99: Exposing a key via `useFilterState` would work!

I think having the key as a prop is an improvement over what we currently have since doing it this way pushes the re-rendering logic into React // takes care of any re-renders for future nested components (if any). It's a little ugly if the key is just a stringified lists of filters, but think it's fine otherwise.

fzhao99 (approved):

PR #1681, opened 2026-08-19
chore: update e2e job to make use of Playwright cache

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

I noticed that the e2e job was downloading Playwright dependencies every time during a job run even though we have a cache step. This PR updates the following steps after the cache step to:

1. Download dependencies on no hit
2. Use cached dependencies on a hit

I've also limited the browsers that get installed to only `chromium` since that is all the project currently makes use of.

## 🧪 How to test

- Take a look at the job and see that the cache is working

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,6 +125,7 @@ jobs:
 
       - name: Cache Playwright Browsers
         uses: actions/cache@v5
+        id: playwright-cache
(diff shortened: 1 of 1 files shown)


lina-roth (approved):

rogeruiz (approved):
lgtm 🌈

PR #1682, opened 2026-08-19
test: fix manage codes e2e test reliability

jakewheeler:
# 🔀 PULL REQUEST

## 💡 Summary

This PR updates the manage codes e2e tests to improve their reliability in the pipeline. We were noticing some issues related to timing and network requests.

- Use `toHaveCount` instead of `count()` and `toBe`
- Fix the way we wait for network requests
- Improve test by not checking specific row order

## 🧪 How to test

- Tests should be passing locally and in the pipeline


rogeruiz (approved):
lgtm 🌈

